PR #12, opened 2016-04-13
Replace eoslhcb addresses with full urls

olantwin:
Dear Thomas,

Outside of the CERN network eos seems to be only accessible if we use the full address using the cern.ch domain (see the email thread with Jacques and Francois on the ship-software list).

This patch is a simple search-and-replace which will hopefully fix most of the related issues. It should be automatically merge-able. 

For machines within the CERN network the change should not change anything.

All the best,

Oliver


ThomasRuf:
ok

olantwin:
Great, thanks!

PR #13, opened 2016-05-13
Fix of possible bug in ShipReco.hit2wire

olantwin:
I stumbled across some duplicated dictionary keys that are later accessed. In my analysis I can't tell whether it affects any results negatively, but that may not be generally true.

If this is indeed intentional, maybe we could add a comment and remove the earlier, overwritten assignment?

I also removed some variables only referenced in a commented out piece of code.

-- Oliver

### DIFF
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -219,15 +219,11 @@ class ShipReco:
      top = ROOT.TVector3()
      bot = ROOT.TVector3()
      modules["Strawtubes"].StrawEndPoints(detID,bot,top)
-     ex = ahit.GetX()
-     ey = ahit.GetY()
-     ez = ahit.GetZ()
    #distance to wire, and smear it.
      dw  = ahit.dist2Wire()
      smear = dw
      if not no_amb: smear = abs(self.random.Gaus(dw,ShipGeo.straw.resol))
      smearedHit = {'mcHit':ahit,'xtop':top.x(),'ytop':top.y(),'z':top.z(),'xbot':bot.x(),'ybot':bot.y(),'z':bot.z(),'dist':smear}
-     # print 'smeared hit:',top.x(),top.y(),top.z(),bot.x(),bot.y(),bot.z(),"dist",smear,ex,ey,ez,ox,oy,oz
      if abs(top.y())==abs(bot.y()): h['disty'].Fill(dw)
      if abs(top.y())>abs(bot.y()): h['distu'].Fill(dw)
      if abs(top.y())<abs(bot.y()): h['distv'].Fill(dw)
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -223,7 +223,8 @@ class ShipReco:
      dw  = ahit.dist2Wire()
      smear = dw
      if not no_amb: smear = abs(self.random.Gaus(dw,ShipGeo.straw.resol))
-     smearedHit = {'mcHit':ahit,'xtop':top.x(),'ytop':top.y(),'z':top.z(),'xbot':bot.x(),'ybot':bot.y(),'z':bot.z(),'dist':smear}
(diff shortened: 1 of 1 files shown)


ThomasRuf:
Hallo Oliver,
Don’t see how your change could work. The dictionary sm is made with self.hit2wire:

smearedHit = {'mcHit':ahit,'xtop':top.x(),'ytop':top.y(),'z':top.z(),'xbot':bot.x(),'ybot':bot.y(),'z':bot.z(),'dist':smear}

the key of the dictionary is 'z', not ‘zbot’, or ‘ztop’. Note, the straws are perpendicular to the z-axis, top and bottom z are identical, as long as we don’t have any misalignments. Maybe this is what caused the confusion:

'z':top.z(),'xbot':bot.x(),'ybot':bot.y(),'z':bot.z(), could also be 'z':top.z(),'xbot':bot.x(),'ybot':bot.y(),'z':top.z(),

cheers,
thomas
Von: Oliver Lantwin [mailto:notifications@github.com]
Gesendet: Freitag, 13. Mai 2016 15:35
Betreff: [ShipSoft/FairShip] Fix of possible bug in ShipReco.hit2wire (#13)


I stumbled across some duplicated dictionary keys that are later accessed. In my analysis I can't tell whether it affects any results negatively, but that may not be generally true.

If this is indeed intentional, maybe we could add a comment and remove the earlier, overwritten assignment?

I also removed some variables only referenced in a commented out piece of code.

-- Oliver

You can view, comment on, or merge this pull request online at:

  https://github.com/ShipSoft/FairShip/pull/13

Commit Summary

  *   Fix duplicate key
  *   Remove unused variables and commented statement

File Changes

  *   M macro/ShipReco.py<https://github.com/ShipSoft/FairShip/pull/13/files#diff-0> (10)

Patch Links:

  *   https://github.com/ShipSoft/FairShip/pull/13.patch
  *   https://github.com/ShipSoft/FairShip/pull/13.diff

olantwin:
Ah, I see. Thank you for clearing up my confusion.

So given that the ideal SHiP straws are perpendicular to the z-axis this should never affect the results.

I think we should add a comment along the lines of `# Note: bot.z()==top.z() unless misaligned` to make the assumptions of the code clear. While I for example thought that may be the case I wasn't sure.

Currently this assumption is hard-coded by just having a `z` key. If we want to keep it like that we should remove one of the two assignments as the first is overwritten anyway.

I'm not sure what you mean by the change working; I am modifying the code that creates the dictionary and all its uses (in ShipReco). But in case other code depends on it is a good argument for not changing the key, and for just removing the duplicate assignment.

(*If you're OK with adding a comment + removing the first assignment I will modify the changes accordingly.*)

ThomasRuf:
The git link only showed me changes to the usage of the dictionary, not a change regarding its creation. Yes, you can go ahead to add this comment and only use one the z values.
Cheers,
thomas

Von: Oliver Lantwin [mailto:notifications@github.com]
Gesendet: Dienstag, 17. Mai 2016 11:31
Betreff: Re: [ShipSoft/FairShip] Fix of possible bug in ShipReco.hit2wire (#13)


Ah, I see. Thank you for clearing up my confusion.

So given that the ideal SHiP straws are perpendicular to the z-axis this should never affect the results.

I think we should add a comment along the lines of # Note: bot.z()==top.z() unless misaligned to make the assumptions of the code clear. While I for example thought that may be the case I wasn't sure.

Currently this assumption is hard-coded by just having a z key. If we want to keep it like that we should remove one of the two assignments as the first is overwritten anyway.

I'm not sure what you mean by the change working; I am modifying the code that creates the dictionary and all its uses (in ShipReco). But in case other code depends on it is a good argument for not changing the key, and for just removing the duplicate assignment.

(If you're OK with adding a comment + removing the first assignment I will modify the changes accordingly.)

olantwin:
I've modified the commits to reflect our discussion. (add comment, remove duplicate assignment, cleanup)

It should be OK to merge now, unless you would like me to make further changes.

Cheers, 
Oliver

PR #17, opened 2016-12-08
Correct magnet length

olantwin:
Hi Thomas,

Here the fixed magnet lengths. I will split magnet `dZ3` in a separate pull request, probably by using `dZ8` in the muonshield config 7. So far I've only corrected config 7, but I can also fix the gaps for configuration 6 if required.

Also some minor cosmetic changes, deleted duplicate assignments and superfluous whitespace. I've only touched the muon shield section though.

Let me know if you'd like me to make any changes. I can also rebase to combine all changes into a single commit if you prefer that.

Cheers,
    Oliver

### Background for those who don't know about the gaps:

The muon shield `C++` code automatically shortens magnets to add gaps, i.e. the lengths given in `geometry_config.py` should be the lengths including gaps. But the current lengths did not account for that and as a result every magnet is 10cm shorter than it should be. By now accounting for the gaps automatically in `geometry_config.py` this is fixed.

### DIFF
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -186,9 +186,9 @@ with ConfigRegistry.register_config("basic") as c:
     c.muShield       =  AttrDict(z=0*u.cm)
     c.muShieldDesign = muShieldDesign
     # design 4,5,6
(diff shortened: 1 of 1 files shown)


ThomasRuf:
config 7 is enough. We should forget about 6. Only 5 and 7 are really supported and maintained.

olantwin:
Should we maybe remove config 6 then (or at least print a scary warning if it is used)? If people need it they can always go back to older commits.

PR #18, opened 2016-12-08
Add supports and elevated floor for the muon shield

olantwin:
Hi Thomas,

Attached my changes to the ShipMuonShield.cxx for the supports and floor.

The logic for the supports is fully contained in commit  660e70d .

The other commits are small cosmetic changes (mostly in an attempt to make the code easier to understand). There's also a bugfix for an incorrect type (`Int_t testGap = 0.1;` results in `testGap == 1`), which is not the intention, but C++ automatically converts this.

As usual, if you'd like me to change anything, remove commits, or merge into a single commit, let me know.

Cheers,
    Oliver

PS: Do we have a convention whether to use `double/int` or `Double_t/Int_t`?

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -140,7 +140,7 @@ void ShipMuonShield::CreateMagnet(const char* magnetName,TGeoMedium* medium,TGeo
 				  Double_t gap,Double_t gap2, Double_t Z, Bool_t NotMagnet)
   {
     Double_t Clgap,Clgap2;
-    int color[4];
+    int color[4] = {30,31,38,45};
 
     if (NotMagnet) {
       Clgap = gap;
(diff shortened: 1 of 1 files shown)


ThomasRuf:
Since we are fully based on ROOT, it would be based to use everywhere the root types, Double_t, Int_t. This ensures same results across operating systems, I think.

olantwin:
A second change that is included, that should not change the geometry is a first step to automatise the substraction of the magnets from the absorber in 19b681c . It doesn't yet automatically detect the list of magnet components, but from the list of magnets and components it substracts them from the absorber geometry to create space.

olantwin:
OK. In future I'll use ROOT types.

ThomasRuf:
one more comment. I think you should use stainless steel and not iron for the support, something which is not magnetic, I guess.

PR #27, opened 2017-02-15
Remove obsolete code

olantwin:
Hi Thomas,

As discussed with you the other day a quick pull request to delete some obsolete code. We should probably deal with commented out code in a similar way, but that would be more effort than it is worth currently.

As old code can always be recovered from the git history there is no need to keep unmaintained code around.

I'm very happy to help if anyone needs instructions on how to selectively recover old versions etc.. From e.g. the Github website it is very simple, but it is also simple to do from the command line.

Cheers,
    Oliver


ThomasRuf:
Hallo Oliver,
I only have one objection: if we remove the code, how will people learn to know that it existed in older versions without digging deep into the history ?

olantwin:
Hi Thomas,

Good point. But are people really aware of the alternate versions currently? I spend a lot of time with FairShip and only discovered them by complete accident. Maybe adding a link to the wiki or note to the README could help, regardless of whether we end up removing the obsolete code.

Cheers,
Oliver

ThomasRuf:
prefer to keep the obsolete code. in any case, folder is called obsolete, so people should be aware that it is just some garbage.

olantwin:
Since we're doing a lot of modifications and cleaning up, maybe we can reconsider removing this old folder?

PR #30, opened 2017-03-07
Fix typos in definition of emulsion pillar dimensions

olantwin:
Hi Thomas,

I came across these typos which break the TP geometry, as `shipDet_conf.py` expects the version with the period. This is consistent to how the new emulsion design (for shield design 7) names the attributes.

Cheers,
Oliver


ThomasRuf:
what does it mean break the TP geometry? It crashes, produces something different?
cheers,
Thomas

olantwin:
A python exception during `shipDet_conf.py` because an expected attribute of `ship_geo` is not found.
    
      File "/[...]/FairShip/python/shipDet_conf.py", line 188, in configure
    EmuMagnet.SetPillarDimensions(ship_geo.EmuMagnet.PillarX, ship_geo.EmuMagnet.PillarY, ship_geo.EmuMagnet.PillarZ)
    AttributeError: 'AttrDict' object has no attribute 'PillarY'

Oliver

olantwin:
So the geometry setup just fails before the simulation runs.

PR #35, opened 2017-04-24
New track pattern recognition

hushchyn-mikhail:
This is the first commit of the new track pattern recognition methods. It contains only one method based on Hough Transform. RunPR.py script runs the pattern recognition. The scripts are under development.

### DIFF
--- a/python/track_pattern_recognition/RunPR.py
+++ b/python/track_pattern_recognition/RunPR.py
@@ -1,0 +1,196 @@
+__author__ = 'Mikhail Hushchyn'
+
+import ROOT
+import numpy
+
+import getopt
+import sys
+
+# For ShipGeo
+from ShipGeoConfig import ConfigRegistry
+from rootpyPickler import Unpickler
+
+# For modules
+import shipDet_conf
+
+# For track pattern recognition
+from geo_init import initialize
+from mctruth import getReconstructibleTracks, get_fitted_trackids, get_track_ids
+from smear_hits import smearHits
+from execute import execute
+
+# For track pattern recognition quality measure
+from quality import init_book_hist, quality_metrics, save_hists
+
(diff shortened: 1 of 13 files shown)


ThomasRuf:
Hi Mikhail,
some comments, starting with the main program, python/track_pattern_recognition/RunPR.py. 

1) in order not to confuse with the existing PR code, could it be placed in a directory maybe called track_pattern_recognition_development? Or mabye we open a new directory, tracking, and inside   pattern_recognition_development? only need to add the directory with the python code to the python path.
2)not obvious to me why this argument  threeprong is needed. You only need to know what is your signal MC particle, then searching the MCTrack container for all particles with Mother == signal type, you have all particles it decayed to. Like this, you can easily expand it to n prong events. Also, you only check for if PDG.GetParticle(atrack.GetPdgCode()).GetName()[:5]=="nu_mu":  but there are also nu_e, nu_tau.

3) initialize(fGeo, ShipGeo): instead of emulating the positions of the tracking stations, it would be better to retrieve their position from the geometry file directory. This will make you independent of any changes in the way the stations are positioned. 

4) would prefer to have one place for smearing hits and not smeared hits code be in any PR code itself. 
Ideally, import shipDigiReco.py and use smearHits method. 

cheers,
thomas

hushchyn-mikhail:
Dear Thomas,

Thanks for your comments.

1. Agree.

(1.2 and 2.) There are several functions that were taken from FairShip/python/shipPatRec.py script and from FairShip/python/shipDigiReco.py. So I keep their arguments. I just slightly modified them to avoid global variables inside the functions. They will be revised in future commits.

3. Totally agree, but I modified your function to avoid global variables (sTree, modules, h). 

Best regards,
Mikhail

PR #36, opened 2017-05-02
Make ShipReco work with files on EOS

olantwin:
Hi Thomas,

This makes the necessary changes to reconstruct files if the geofile or the input file is on EOS.

Not sure whether this covers all cases (or if there is a more elegant way to achieve the same), but it works for my use-case.

Cheers,
Oliver

### DIFF
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -84,14 +84,15 @@ else:
 # outfile should be in local directory
   tmp = outFile.split('/')
   outFile = tmp[len(tmp)-1]
-  if saveDisk: os.system('mv '+inputFile+' '+outFile)
+  if inputFile[:7]=="root://" : os.system('xrdcp '+inputFile+' '+outFile)
+  elif saveDisk: os.system('mv '+inputFile+' '+outFile)
   else :       os.system('cp '+inputFile+' '+outFile)
 
 if not geoFile:
  tmp = inputFile.replace('ship.','geofile_full.')
  geoFile = tmp.replace('_rec','')
 # try to figure out which ecal geo to load
-fgeo = ROOT.TFile(geoFile)
+fgeo = ROOT.TFile.Open(geoFile)
 sGeo = fgeo.FAIRGeom
 
 from ShipGeoConfig import ConfigRegistry
--- a/python/shipDigiReco.py
+++ b/python/shipDigiReco.py
@@ -9,7 +9,7 @@ start = ROOT.TVector3()
 class ShipDigiReco:
  " convert FairSHiP MC hits / digitized hits to measurements"
  def __init__(self,fout,fgeo):
-  self.fn = ROOT.TFile(fout,'update')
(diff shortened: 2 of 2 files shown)


ThomasRuf:
For the geofile, the change is ok. 
For the reco case, your change implies that starting with a file on EOS, the output would also be on EOS, and in the same directory. It is fine if the person running the reconstruction is also the person running the simulation. But a more general use case would be, somebody runs large stats of simulation to be used by others. Reconstruction is improving. Different people will work on it. Not everybody should then write the output to the same directory. 
For ShipReco, I would prefer not to couple directly the input and output. For your use case, nothing prevents you to run a top level macro, which calls ShipReco and when ShipReco finished copies the local output file to the desired destination.

thomas

olantwin:
Thomas,

I agree. 

But the outputfile is in the current directory anyway (this does not change). Essentially this automatically makes a local copy of the inputfile (wherever it is) and uses that as the outputfile.

So regarding the inputfile the code just learns how to copy files from EOS.

Oliver

ThomasRuf:
you are right. I overlooked that xrdcp makes the local copy. There is no automatic upload copy to EOS.  Then the following would also include the use case of not keeping the simulation file:
  if inputFile[:7]=="root://" : os.system('xrdcp '+inputFile+' '+outFile) 
  if saveDisk: os.system('mv '+inputFile+' '+outFile) 
  else :       os.system('cp '+inputFile+' '+outFile)

olantwin:
Are you suggesting that I replace `elif` by `if`?

In the case of EOS the `saveDisk` option is silently ignores because of the `elif` (as we probably never want to delete the EOS copy). If we replaced the `elif` with an `if` we reach the `os.system('mv'...)` or `os.system('cp'....)` statements, which will fail because `inputfile` is remote.

olantwin:
Ah, I see what you mean. If we copied to some intermediate file first we could then either move or copy the local copy of the EOS file. I could add a statement that assigns the name of the local copy to `inputFile`, so that the following lines work.

ThomasRuf:
i think all what is needed is
NEW: +  if inputFile[:7]=="root://" : os.system('xrdcp '+inputFile+' '+outFile) 
if saveDisk: os.system('mv '+inputFile+' '+outFile) 
else :       os.system('cp '+inputFile+' '+outFile)

For my education, can I change your commit before merging?

thomas

olantwin:
You should have write access to my `reco_eos` branch. So you should be able to do `git remote add olantwin git@github.com:olantwin/FairShip.git`, `git fetch olantwin/reco_eos`, `git co reco_eos`, make your changes and then `git push -u olantwin reco_eos`.

I'm not convinced you're change would work like that. We would need an intermediate file to copy to for this to work (e.g. outFile with `_rec` removed). Otherwise `mv` or `cp` would fail because of inputFile being remote.

ThomasRuf:
you are right. Your code should work as it should. Sorry for the confusion.
thomas

PR #37, opened 2017-05-15
Track Pattern Recognition

hushchyn-mikhail:
News:
* More quality metrics and plots
* Hough Transform is tuned and uses dist2wire
* All functions are described and commented
* Short general description is added
* python requirements are added


ThomasRuf:
Hi,
Is this pull request still up-to-date? 
Thomas

hushchyn-mikhail:
Dear Thomas,

Updates:
- Rewritten base PR and new PR methods -> python/shipPatRec.py
- Run new PR to measure quality metrics -> python/shipStrawTracking.py
- Old version of PR -> python/shipPatRec_old.py
- Old version of shipStrawTracking.py -> python/shipStrawTracking_old.py
- New PR methods are integrated into python/shipDigiReco.py

Results of working PR methods are saved in output .root file of marco/ShipReco.py script. Results are in PatRecPoint branch.

Development directory is removed.

Currently the base PR (rewritten) will be runed if set --realPR option in macro/ShipRec.py.

Mikhail

ThomasRuf:
Looking quickly at the proposed changes, I see
+  self.patRecArray  = ROOT.TClonesArray("strawtubesPoint") 
+  self.patRecPoint   = self.sTree.Branch("PatRecPoint",  self.patRecArray, 32000, -1) 
this is done independent if the realPR is called or not. 

Independent of this, if I understand correctly, you make a complete copy of strawtubesPoint contributing to the found tracks persistent. Did you checked how much this adds to the data size? This should be avoided also because you are using the MC truth, momentum of particle, pdg code, etc.. How would you apply the code do data where no MC information is present? You should use an array with the keys to smearedHits,  a list of integers / track. Eventually, much better would be to use the digitized hits.  
I think you should really make a sharp distinction between simulated real data, like smearedHits or Digi_StrawtubesHit. MC info should only be used for debugging/monitoring.
 If you base it on Digi_StrawtubesHit, and make the list of keys which contribute to the track persistent, you get the access to the MC truth for free via digiSBT2MC which return for each Digi_StrawtubesHit the key of the underlying strawtubesPoint, which then has a link to the underlying MCTrack. 
This avoids to duplicate any existing information on the output file.

hushchyn-mikhail:
PatRecPoint does not contain MC truth. It has the following fields:

- DetectorID - strawtube identifiers
- TrackID - identifier of recognized tracks. -999 means, that a hit is not recognized.
- X, Y, Z - reconstructed hit coordinates
- Px, Py, Px - reconstructed particle momentum
- dist2Wire - distance between a track and a wire inside a strawtube. This is smeared value.
- EventID - event id.
- PdgCode - 11 for negative charge of reconstructed particle , -11 for positive charge. -999 - charge is not reconstructed.
- Time  - -999 (values are unknown)
- Length  - -999 (values are unknown)

PatRecPoint branch has the same size as strawtubesPoint brach. PatRecPoint contains reconstructed information about hits stored in strawtubesPoint. 

For the real data, PatRecPoint will contain the same information. It is MC independent. However, it saves all information about PR procedure. This information can be used to easily  measure the PR efficiency.  PatRecPoint can be very useful for future updates of the experiment. That is the main point why PatRecPoint was created :)

This can be made as an option of ShipReco, of course.

ThomasRuf:
Sorry, don't understand.
1) PatRecPoint branch has the same size as strawtubesPoint branch. How can this be? There will be many strawtubesPoints which will not be used for pattern recognition. And vice versa, tracks which share strawtubesPoints.
2) What is the relation to digiStraw?
3) The straw tracker is a 1-dim detector. How can you derive 3d points for each PatRecPoint?
4) What is the meaning of px,py,pz for each point?
5) For measuring the PR efficiency, I only need to know what is a reconstructible track using some basic criteria (number of hits, fiducial volume), and the tracks found by PR. In any case, don't understand why to make all these information persistent for eventually measuring the efficiency. Most people will not use it, and it is easy to recreate it if necessary for determining the efficiency.   
5) Keep also in mind that genfit keeps information about each measurement used in the trackfit. Persistency switched off by default, but can be made persistent for tracking studies.

ThomasRuf:
May I make another remark. I think it would be useful to have a setup that we can change easily from one pattern recognition algorithm to another to compare their performances. I don't know if you meant by changing the previous version to XXX_old, that his version is still functioning, or you only renamed these files to XXX_old for historic reasons. If it is the first case, a better name than old might be needed.

hushchyn-mikhail:
Dear Thomas,

I just try to find the most convenient way to store information after PR to get simple way to work with PR for other people. But I can remove PatRecPoint, if you sure that it is not necessary. Probably, can we discuss this at a meeting?

> PatRecPoint branch has the same size as strawtubesPoint branch. How can this be? There will be many strawtubesPoints which will not be used for pattern recognition. And vice versa, tracks which share strawtubesPoints.

If a hit does not belong to any recognized track, its track_id will be -999.
Share hits will have only one recognized track_id. 

> What is the relation to digiStraw?

Dist2wire is digitized. If you mean that.

> The straw tracker is a 1-dim detector. How can you derive 3d points for each PatRecPoint?
> What is the meaning of px,py,pz for each point?

The same way as it done in strawtubesPoint branch. I am afraid, that we misunderstand each other. StrawtubesPoint branch contains information about hits. PatRecPoint branch contains information about the same hits after PR in the same form as in strawtubesPoint. Simply speaking, you can take strawtubesPoint and PatRecPoint branches and go through them hit by hit, comparing information in the branches. So, px, py, pz for a hit in PatRecPoint are momentum projections of a recognized track which has this hit. The same for MC truth in strawtubesPoint branch.

> For measuring the PR efficiency, I only need to know what is a reconstructible track using some basic criteria (number of hits, fiducial volume), and the tracks found by PR. In any case, don't understand why to make all these information persistent for eventually measuring the efficiency. Most people will not use it, and it is easy to recreate it if necessary for determining the efficiency.

Thera are a lot of metrics can be measured for PR: efficiencies for Y, Stereo views in stations 1&2, 3&4; track purity for different views and stations, resolution of left-right ambiguity, humber of hits per recognized track for different views and stations, different momentum dependencies ... 
And PatRecPoint provides (for my opinion) the simplest way to compute these metrics and store all needed information for them.

> Keep also in mind that genfit keeps information about each measurement used in the trackfit. Persistency switched off by default, but can be made persistent for tracking studies.

It is true. But I didn't go this way because PR and trackfit are different steps. And I am not sure, this is the most convenient to store all needed info.

> May I make another remark. I think it would be useful to have a setup that we can change easily from one pattern recognition algorithm to another to compare their performances. I don't know if you meant by changing the previous version to XXX_old, that his version is still functioning, or you only renamed these files to XXX_old for historic reasons. If it is the first case, a better name than old might be needed.

Totally agree :)

ThomasRuf:
Just to clarify one thing: strawtubesPoint is pure MC information, contains the x,y,z coordinates of particle traversing the corresponding straw (gas volume to be precise). This should not be used directly to start the pattern recognition. This information is not available for real data.
My view:  The start of the pattern recognition should just be a list of straw ids, eventually together with the drift time (smeared, including resolution). With the straw id, one can ask the geometry manager for the position of the straw in space, start and end point of the wire. In this step, one can in future account also for misalignments.  With this information, the pattern recognition will come up with an array of lists of straw indices, track candidate == list of straw indices. Track candidates can share straw ids, no requirement to be unique.
For this purpose, I created some time ago the class strawtubesHit. Objects are created from strawtubesPoint (using the method digitizeStrawTubes ) but can also be created from real data in future. It just contains 
    *@param detID    Detector ID
     *@param digi      digitized/measured TDC 
     *@param flag      True/False, false if there is another hit with smaller tdc 
     **/
and methods to access
    void StrawEndPoints(TVector3 &vbot, TVector3 &vtop); 
   and detID/TDC.

there is a one to one relation between Digi_StrawtubesHits and strawtubesPoint. With the index of a strawtubesHit in the Digi_StrawtubesHits container, one can also access the strawtubesPoint in the strawtubesPoint container to access the MC truth information.  

The output of the pattern recognition, track candidate == list of strawtubesHit indices,  is given to the trackfit. A normal user will be only interested in the result of the trackfit, momentum at a given z-position. Eventually, information about tracklets, tracks which have too less hits to be useful for a trackfit might be interesting for vetoing background events. 

The monitoring, determining efficiencies using MC truth, should be done outside the standard pattern recognition algorithm.    

thomas

hushchyn-mikhail:
Dear Thomas,

The PR code updates are done:
- New PR methods are added to shipDigiReco.py and ShipReco.py. 
- It is also possible to run previous PR version.
- Results of the new PR methods are stored in the Tracklets branch in the output *_rec.root file.
- The new PR recognizes tracks in all stations and also tracklets in stations 1&2.

How to run:
$ python $FAIRSHIP/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root --realPR=<something>

- --realPR='' - cheated PR
- --realPR=Prev - previous version of the PR
- --realPR=FH - Hough Transform (the new PR)
- --realPR=AR - Artificial Retina (the new PR)
- --realPR=Baseline - rewritten previous version of the PR (was taken as baseline solution)

Metrics:
$ python $FAIRSHIP/python/shipStrawTracking.py -f ship.conical.Pythia8-TGeant4_rec.root -g geofile_full.conical.Pythia8-TGeant4.root -o hists.root

Metrics are calculated using track indexes stored in the output *_rec.root file.


Can you describe the next steps should be done to merge the new PR methods into the master, please?

Best regards,
Mikhail

hushchyn-mikhail:
Dear Thomas,

thanks for the idea. It is done.

Best regards,
Mikhail

ThomasRuf:
could you please remove the .gitignore from this pull request? It does not fit to the thematic.

ThomasRuf:
I started to have a look at the reco. Simulated 1000 events. Then run with default option, --realPR=Prev and  --realPR=Baseline. I get:

default, fake pattern: 328 fitted tracks
prev pattern:  249 fitted tracks (looks very low to me, eff onl 76%!)
baseline: 0 fitted tracks, tracklets seem to exist, but never promoted to full fitted tracks. Is this planned?

thomas

olantwin:
This shouldn't delete the `.gitignore` though, right?

hushchyn-mikhail:
What is the right way then?

olantwin:
Maybe there's a better way, but I think the easiest would be to:

1. Get rid of that last commit (best google this part. I never remember the exact command. Let me know if you need help)
2. Identify which commit changed the `.gitignore`:
    * If it's a commit that changed nothing else, it's easy to remove
    * It it's a commit that changed several things it will require a little bit more effort.
3. Use `git rebase -i` to fix it (I will tell you exactly what you need to do, if you tell me what you figured out in step 2.).

olantwin:
Now we have even more commits.

@hushchyn-mikhail, let's discuss this on telegram. This might be a long back and forth otherwise...

olantwin:
@ThomasRuf:

The last two commits we managed to remove. The problem is that the simple solution I had in mind can't deal with all the merges/syncs between the two repositories, so we'll need to find a different solution (well, there's a brute force solution as well, which I'd like to avoid).

I'll look into this further after the JPD meeting tomorrow.

hushchyn-mikhail:
Dear Thomas,

About your tests. Code for the fake and previous solution have not been changed. The differences between the number of fitted tracks can be explained by restrictions applied to fitted tracks in the previous solution.

However, 0 fitted tracks for the baseline is strange. Could you describe your steps please?

ThomasRuf:
I run
  python /afs/cern.ch/user/t/truf/FairShip/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root 
  python /afs/cern.ch/user/t/truf/FairShip/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root --realPR=Prev 
  python /afs/cern.ch/user/t/truf/FairShip/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root --realPR=Baseline 
  
and analyzed the output with ShipAna.

thomas

hushchyn-mikhail:
Dear Thomas,

The PR is fixed. The problem was in fittedtrackids. They must be the MC true track ids of the fitted tracks (as in the fake PR and previous PR).

Best regards,
Mikhail

ThomasRuf:
still no success with
python $FAIRSHIP/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root --realPR=baseline

zero tracks. Could you check please that you can produce a rec file, which can be analyzed by ShipAna?

best regards,
Thomas

ThomasRuf:
I have one more issue. There are now several options how to run the pattern recognition, fake, prev, FH, AR, baseline. If one looks at a reconstructed file, how can one find out which PR had been used? Does anybody have a suggestion?

ThomasRuf:
Concerning my zero tracks, I figured out that this is due to giving the wrong option, --realPR=baseline, instead of --realPR=Baseline. There is no check done that the argument fits any known option.

olantwin:
@ThomasRuf maybe we should use argparse? It automatically checks for known/unknown options, defaults etc. One can easily define lists of allowed values as well.
With getopt we'd have to do it manually for every script.

ThomasRuf:
Found another issue, clones:
using the FH PR option, I found several events with more than one reconstructed HNL candidate. The reason for this, reconstructed tracks which are clones of each other. Here one example:
>>> sTree.FitTracks[1]
("FitTrack") chi2/dof:0.8  P: 1.51GeV/c pdg:-13
>>> sTree.FitTracks[2]
("FitTrack") chi2/dof:0.8  P: 1.51GeV/c pdg:-13
>>> sTree.FitTracks[3]
("FitTrack") chi2/dof:0.8  P: 1.51GeV/c pdg:-13

pointing to the same MC particle:
>>> sTree.fitTrack2MC[1]
15
>>> sTree.fitTrack2MC[2]
15
>>> sTree.fitTrack2MC[3]
15

which then results in three HNL candidates:
>>> sTree.Particles[0]
("ShipParticle") N2 M: 0.62GeV/c2 P:20.79GeV/c VxZ:-3.82m
>>> sTree.Particles[1]
("ShipParticle") N2 M: 0.61GeV/c2 P:20.78GeV/c VxZ:-3.25m
>>> sTree.Particles[2]
("ShipParticle") N2 M: 0.61GeV/c2 P:20.78GeV/c VxZ:-3.23m

if there is no special precaution taken this will result in an overstimate of the signal efficiency. Maybe we need to think about an algorithm which removes such track clones and only select one.

thomas

PR #38, opened 2017-05-28
Actually set the mass + fix misleading indentation in `run_simScript.py`.

Element-126:
Hi,

I noticed that specifying a custom mass using the `-m`/`--mass` command-line parameter of `run_simScript.py` currently does not make Pythia use the requested value (for HNLs and RPVSUSY) due to a trivial error. Instead it keeps using the default value of 1 GeV.

This PR fixes this behaviour. The following notebooks allow you to compare the results [before](https://github.com/Element-126/My-notebooks/blob/master/HNL_mass_before.ipynb) / [after](https://github.com/Element-126/My-notebooks/blob/master/HNL_mass_after.ipynb) when setting `--mass '1.5'` (look at the first histogram).

Note that the recorded mass is still wrong (it remains fixed at 1 GeV), but this is a minor error and does not affect the physics.

In addition, I replaced a few raw tabs by spaces, since they were messing up the indentation, resulting in hard to find errors.

This is the squash of 2 commits:
1. Fix misleading indentation due to rogue tabs.
2. Actually set the mass in the HNL/RPVSUSY case, when the `-m`/`--mass` command-line parameter is used.

Feel free to edit this PR or request the unsquashed version if you prefer.

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -63,7 +63,7 @@ except getopt.GetoptError:
         print '       --SusyBench to specify which of the preset benchmarks to generate (default 2)'
         print '       --mass or -m to set HNL or New Particle mass'
         print '       --couplings \'U2e,U2mu,U2tau\' or -c \'U2e,U2mu,U2tau\' to set list of HNL couplings'
-	print '       --epsilon value or -e value to set mixing parameter epsilon' 
+        print '       --epsilon value or -e value to set mixing parameter epsilon' 
         print '                   Note that for RPVSUSY the third entry of the couplings is the stop mass'
         sys.exit()
 for o, a in opts:
@@ -121,7 +121,7 @@ for o, a in opts:
             dv = int(a)
         if o in ("--muShieldDesign",):
             ds = int(a)
-	if o in ("--nuTauTargetDesign",):
+        if o in ("--nuTauTargetDesign",):
             nud = int(a)
         if o in ("--charm",):
             charm = int(a)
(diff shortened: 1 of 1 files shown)


ThomasRuf:
Hi,
Thanks for finding this mistake. It seems, it had been introduced with this commit: https://github.com/ShipSoft/FairShip/commit/b136c180d43f06dc697609cff28717156945b8a5 

the previously used variable -theHNLmass = 1.0*u.GeV  , had been replaced by +theMass = 1.0*u.GeV . 

I would prefer to move back using theHNLmass, rather than applying your patch  if HNL or RPVSUSY: theMass = float(a) . Could you make the change, then I will merge the new pull request including the changes of tabs.

cheers,
Thomas

Element-126:
Hi,

It looks like your fix `c5e86bc1` should do the job. Since I am not familiar with RPVSUSY, I was not aware that the relevant mass was that of the dark photon.

I will run my test case on the latest `master` and close this issue if everything is ok.

Element-126:
I just checked the results and they now seem correct. Thanks for the quick resolution !

I'm closing the issue.

ThomasRuf:
Hi,
No, my fault, dark photon mass should not be used for RpvSusy. However, there is no dedicated default value for RpvSusy mass and none for the couplings. So, I will change back using the same values for HNL and RpvSusy and only have theDPmass and theDPepsilon extra.

Thanks for reporting,
Thomas

Von: Element-126 [mailto:notifications@github.com]
Gesendet: Dienstag, 6. Juni 2017 23:19
Betreff: Re: [ShipSoft/FairShip] Actually set the mass + fix misleading indentation in `run_simScript.py`. (#38)


Hi,

It looks like your fix c5e86bc1 should do the job. Since I am not familiar with RPVSUSY, I was not aware that the relevant mass was that of the dark photon.

I will run my test case on the latest master and close this issue if everything is ok.

Element-126:
Ok. Since I'm not yet familiar with RpvSusy/DP (only νMSM), I'd rather trust your solution than attempt to fix it myself.

Good luck !

PR #43, opened 2017-06-14
Add SetSupports method to disable supports; don't create floor if it's height is 0

olantwin:
Hi Thomas,

here the requested change.

I will quickly need to fix some merge issues, as Fons seems to have changed all the line endings...

Cheers,
Oliver

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -45,7 +45,7 @@ ShipMuonShield::ShipMuonShield(const char* name, const Int_t Design, const char*
  fDesign = Design;
  fField  = field;
  if (fDesign==1){
-     fMuonShieldLength = L1;
+     fMuonShieldLength = L1;   
     }
  if (fDesign==2 || fDesign==3 || fDesign==4 ){
      Fatal("ShipMuonShield","Design %i not anymore supported",fDesign);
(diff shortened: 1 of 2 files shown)


olantwin:
OK, the changes now are mergeable. We should probably try to correct the line endings on a file-by-file level at some point, as we seem to have some files with DOS and some with Linux line endings...

ThomasRuf:
Isn’t there some git command for this?
thomas

Von: Oliver Lantwin [mailto:notifications@github.com]
Gesendet: Mittwoch, 14. Juni 2017 14:39
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [ShipSoft/FairShip] Add SetSupports method to disable supports; don't create floor if it's height is 0 (#43)


OK, the changes now are mergeable. We should probably try to correct the line endings on a file-by-file level at some point, as we seem to have some files with DOS and some with Linux line endings...

olantwin:
I think there are some git commands that work if the entire code-base has the same endings and only the local endings differ. Not sure there is a way to handle file-by-file endings.

Converting an entire file is easy using the `dos2unix` or `unix2dos` tools, which strip whitespace at the end of the line, which is responsible for the white-space changes in the updated commit.

PR #45, opened 2017-06-20
vertex fit implementation

ghkorol:


ThomasRuf:
Hi Ievgen,
I would prefer to decouple the reconstruction step with the monitoring step, meaning not filling histograms during the reconstruction step itself, if possible. But rather adding another method, monitoring or checking.
thomas

ghkorol:
Hi Thomas, 
Can we just comment declaration and filling of the histograms by now?
Later we will put it to monitoring script.
Cheers, Ievgen

ThomasRuf:
OK

ThomasRuf:
Looking in more detail at the code:

what is the difference between 
HNLtest =  ROOT.ShipParticle(13,0,-1,-1,12,13,P,vxFit) , why hardcoded indices of tracks?
and 
particle = ROOT.ShipParticle(9900015,0,-1,-1,t1,t2,HNL,vx)  

shouldn't there be only one? which one should be made persistent?

Fill(1.-HNL.M()) : you assume that all HNLs have mass = 1, which is only the default. Try to get the mass from the ROOT.TDatabasePDG.Instance()

PR #46, opened 2017-06-22
Fix overlaps in absorber and make sure absorber correct length for magnets

olantwin:
Hi Thomas,

This should now properly fix both the overlaps and the previous issue of the absorber not being the correct length.

It turns out I was slightly to hasty previously and had only fixed the obvious symptoms of the length mis-match, but didn't notice the more subtle underlying problem.

Cheers,
	Oliver

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -434,13 +434,15 @@ void ShipMuonShield::ConstructGeometry()
 		   gapIn[nM],gapOut[nM],Z[nM],1);
       }
 
-      TGeoTranslation* mag1 = new TGeoTranslation("mag1",0,0, -dZ1);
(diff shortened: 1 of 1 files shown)


ThomasRuf:
All checks have failed
1 failing check 
continuous-integration/travis-ci/pr — The Travis CI build failed  
see https://github.com/ShipSoft/FairShip/pull/46

what does it mean?

olantwin:
If you check the [details](https://travis-ci.org/ShipSoft/FairShip/builds/245668948?utm_source=github_status&utm_medium=notification) you'll see that the macOS build fails, but the linux build succeeds.

Currently master [also fails](https://travis-ci.org/ShipSoft/FairShip/builds/245797875?utm_source=github_status&utm_medium=notification) to pass the checks on macOS, so I think this is related to the most recents commits by @FonsRademakers , and unrelated to my code.

PR #47, opened 2017-06-29
Implement the new beam smearing scheme

olantwin:
At the muon flux meeting and the last collaboration meeting [Marco Calviani presented a new beam smearing scheme](https://indico.cern.ch/event/638036/contributions/2593440/attachments/1471589/2277517/mc__BDF_TTC_SHiP_CM_June2017_v1.3.pdf#page=12):

* 8mm beam size (1σ)
* 50mm radius of ring

This pull-request implements this scheme, but there are still some issues I would like feedback on:

* Currently this just replaces the old scheme. We probably want to keep the old scheme as an option
* Instead of just one parameter of the scheme (the radius), we now have two (radius and width of the ring). How do we handle this?
* Needs some more testing.

PS: I suggest [WiP] (work in progress) as a tag for pull-requests that are not ready for merging yet.

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -387,7 +387,7 @@ if simEngine == "MuonBack":
  primGen.SetTarget(ship_geo.target.z0+50*u.m,0.)
  MuonBackgen = ROOT.MuonBackGenerator()
  MuonBackgen.Init(inputFile,firstEvent,phiRandom)
- MuonBackgen.SetSmearBeam(3*u.cm) # beam size mimicking spiral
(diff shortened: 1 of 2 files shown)


olantwin:
OK, I updated this and I think I also figured out why it didn't work previously.

Tests for the old muon background:

smearing

@ThomasRuf: If you could test with the new muon background (or could give me a file I could test with) that would be great.

olantwin:
I think the main question is whether we want to have this new scheme as an option in parallel to the old or whether it should replace the old (which it currently does in this implementation).

PR #48, opened 2017-07-08
Pull request for Vertex fit

ghkorol:


ThomasRuf:
Hi Ievgen,
we should avoid using "from amodule import *", one could override other variables/functions without noticing. 
we should also add doca as a proper data member to the ShipParticle class. 
will test the code now. 

thomas

PR #49, opened 2017-07-17
B fields using the VMC interface

jback08:
This set of code changes & additions allows B fields to be specified via the VMC Geant4/ROOT framework, allowing global/local fields as well as (ROOT) field maps. See the README file in the field directory for info.


ThomasRuf:
Dear John,
Thanks for your contribution. It is a long list of important changes, and difficult to see the overall impact. 

Would you be able to present the underlying idea in a ShipSoft meeting, together with a demonstration that the muon background rate does not change, and the momentum/mass resolution for HNL decays stays constant?

Looking quickly at the files, some immediate comments:
field/BFieldSetup.txt, this looks like a static text file and I am wondering how it is synchronized with passive/ShipMuonShield.cxx and the different configurations inside, changes of names etc. Who is the master? Couldn't the txt file be created dynamically from the geometry created by passive/ShipMuonShield.cxx? 
If I want to change the field value from 1.8 Tesla to 1.6 Tesla, I would have to change the txt file? 

cheers, thomas

olantwin:
Hi John,

I can't look at the changes in detail right now (I will later), but I am a bit confused by the `REAMDE.md`:
Do fields that are already defined as `TVirtualMagField`s need to be listed in `field/BFieldSetup.txt`?

I am trying to wrap my head around what the implications would be for the muon shield implementation, especially  as we are maintaining several configurations.

Cheers,
Oliver

jback08:
Hi Thomas, Oliver,

I can give an update in a meeting, although I have not yet calculated the background rate and mass resolutions; are there any scripts I can run to get these numbers, and what is the best procedure (using CERN lxplus) to generate enough statistics for meaningful comparisons?

From my testing so far, if you already have a `TVirtualMagField` defined for a geometry object in the C++ code you do not need to add it again to the `field/BFieldSetup.txt` file, since the VMC already knows about it (and each field will remain local to its previously defined volumes) via the following call in `gconfig/g4config.in`:

```
/mcDet/setIsLocalMagField true
```

For this reason the `setMagnetField()` function originally in `python/geomGeant4.py` is removed, since the VMC knows about the fields and tells Geant4 what they are.

The `field/BFieldSetup.txt` file is static, but I can imagine this could be produced dynamically with the input geometry. This input file is read by the `ShipFieldMaker::makeFields()` function that simply calls other `ShipFieldMaker` methods to create new, or override existing, fields. Instead of this input file we can call the `ShipFieldMaker` field creation methods directly in the python script(s), although I have not yet tried this.

If you want to change the muon shield field from 1.8 to 1.6 T then we can either a) change the values in `field/BFieldSetup.txt`, or b) comment out the call 

```
fieldMaker = geomGeant4.addVMCFields('field/BFieldSetup.txt', False)
```

in `macro/run_simScript.py` (or provide an empty text file name here), and update the variable `c.muShield.Field` from 1.8 to 1.6 in `geometry/geometry_config.py`. 

Values in `field/BFieldSetup.txt` override all other definitions, since the field maker object is created near the end of the `macro/run_simScript.py` script. If the `BFieldSetup.txt` file is empty, or not present, then no new fields are created.

The main aim of this code is to add the option of using field maps, as well as allowing the configuration of a general global field, and defining which volumes feel either the local and/or global fields.

ThomasRuf:
Hi John,
For HNL signal:
execute run_simScript.py say for 1000 events:
python $FAIRSHIP/macro/run_simScript.py -n 1000
will produce ship.conical.Pythia8-TGeant4.root and geofile_full.conical.Pythia8-TGeant4.root

run reco:
python $FAIRSHIP/macro/ShipReco.py -f ship.conical.Pythia8-TGeant4.root -g geofile_full.conical.Pythia8-TGeant4.root
will produce
ship.conical.Pythia8-TGeant4_rec.root

run analysis:

python -i $FAIRSHIP/macro/ShipAna.py -f ship.conical.Pythia8-TGeant4_rec.root -g geofile_full.conical.Pythia8-TGeant4.root

will produce plots with momentum, mass, doca, IP resolutions.

For muon background, this is a bit more time consuming, but done with run_simScript.py and option --muonBack  --FollowMuon -f /eos/ship/data/Mbias/pythia8_Geant4-withCharm_onlyMuons_4magTarget.root

Counting the number of reconstructed tracks in the spectrometer would be one way of counting background. Running over the full input file, one would expect about ~30 tracks if the shield works.

Cheers,
thomas


Von: John Back [mailto:notifications@github.com]
Gesendet: Montag, 17. Juli 2017 17:45
Betreff: Re: [ShipSoft/FairShip] B fields using the VMC interface (#49)

Hi Thomas, Oliver,
I can give an update in a meeting, although I have not yet calculated the background rate and mass resolutions; are there any scripts I can run to get these numbers, and what is the best procedure (using CERN lxplus) to generate enough statistics for meaningful comparisons?
From my testing so far, if you already have a TVirtualMagField defined for a geometry object in the C++ code you do not need to add it again to the field/BFieldSetup.txt file, since the VMC already knows about it (and each field will remain local to its previously defined volumes) via the following call in gconfig/g4config.in:
/mcDet/setIsLocalMagField true
For this reason the setMagnetField() function originally in python/geomGeant4.py is removed, since the VMC knows about the fields and tells Geant4 what they are.
The field/BFieldSetup.txt file is static, but I can imagine this could be produced dynamically with the input geometry. This input file is read by the ShipFieldMaker::makeFields() function that simply calls other ShipFieldMaker methods to create new, or override existing, fields. Instead of this input file we can call the ShipFieldMaker field creation methods directly in the python script(s), although I have not yet tried this.
If you want to change the muon shield field from 1.8 to 1.6 T then we can either a) change the values in field/BFieldSetup.txt, or b) comment out the call
fieldMaker = geomGeant4.addVMCFields('field/BFieldSetup.txt', False)
in macro/run_simScript.py (or provide an empty text file name here), and update the variable c.muShield.Field from 1.8 to 1.6 in geometry/geometry_config.py.
Values in field/BFieldSetup.txt override all other definitions, since the field maker object is created near the end of the macro/run_simScript.py script. If the BFieldSetup.txt file is empty, or not present, then no new fields are created.
The main aim of this code is to add the option of using field maps, as well as allowing the configuration of a general global field, and defining which volumes feel either the local and/or global fields.

olantwin:
Hi John,

If `TVirtualMagFields` are discovered correctly, could you please remove the muon shield magnets from `field/BFieldSetup.txt`?

Otherwise no objections from me (assuming the backgrounds are unchanged). Especially the new ways to add field maps will be very useful!

Cheers,
Oliver

jback08:
Hi Thomas, Oliver,

Thanks for the additional info and feedback. Apologies for the delay getting back to you since I did not get your recent messages from github via email.

I will run the MC/reco tests. The `field/BFieldSetup.txt` file is only there as an example that "replicates" what the C++ geometry currently does, so I will rename this file to `field/ExampleBFieldSetup.txt`.  I think we only really need to define the global field for the spectrometer (need to double check this via the tests), so I will leave out the muon shield magnets in a more streamlined `field/BFieldSetup.txt` file.

Cheers, 
    John.

jback08:
There is a bug in the original implementation of `setMagnetField` in `python/geomGeant4.py` which assigns magnetic fields from Root/TGeo to Geant4. No fields are set for the following (tertiary) sub-volumes for the tau spectrometer:

a) volUpYoke, volLowYoke, volArm1MS and volArm2MS (within volMagneticSpectrometer), b) ColumnVol and c) volTarget

This is fixed by adding a third loop over the geometry volumes within the second loop.

After this correction, I get exact agreement for 1000 HNL reconstructed events (using the same random seed) between the original (fixed) `setMagnetField` method and the VMC interface that just uses

`/mcDet/setIsLocalMagField true`

with either an empty `field/BFieldSetup.txt` (i.e. the fields are already defined by the C++ geometry) or using the example setup file that "replicates" the field definitions.

What I need to do now is check the background rates, although running over the full ~18 M muon events will take over a day per test. Is there a way I can speed this up (was thinking of just running ~1M events)? Also, what tuple variable from the reco output stores the number of reconstructed tracks in the spectrometer?

olantwin:
For the muon background you can split it into several jobs which makes everything quite quick. I think for this purpose looking at the background for muons only (via the `--FollowMuon`) option should be sufficient.

Now that I'm back from my vacation I could even run the muon background for you, as I have all the scripts etc.. already set up, and only need to substitute your branch for master.

jback08:
Hi Oliver,

Yes, it will be very good if you have time to run over the muon sample with this BField branch. I have just updated this to not read in the example set-up text file by default, as well as illustrating the changes required for the original `setMagnetField` function in `python/geomGeant4.py`.

I would like to know how to run such tests myself, so please forward to me the job instructions/scripts you use (I have not used the docker functionality yet).

Thanks, John.

olantwin:
OK, I will start the simulation for the current commit of  your BField branch.

I will send you the scripts and some instructions in a separate email tomorrow.

olantwin:
@jback08 Any reason you removed the executable permissions from `macro/run_simScript.py`?

jback08:
@olantwin Not sure what happened here. I've reinstated the executable permissions for the run script.

olantwin:
I also noticed that a lot of links in `fields/README.md` point to your fork of FairShip. Not sure if there's a way to use internal links within the `github.com` repo, but if not maybe they should point to the ShipSoft repository instead?

jback08:
The links in the `field/README.md` file are actually defined to be local within the given repository, so when this code is added to the official branch it will pick up that version and not the fork. For example, links are defined as

[filename.extension] (../directory/actual_file_name)

e.g. 

"\[geomGeant4.py\]\(../python/geomGeant4.py\)"

It only appears that a specific git fork is being used for the file links when viewed in the web browser.

olantwin:
Ah, OK. Thanks for clarifying!

olantwin:
So, I ran the muon background (several times actually), and while the performance seems to be consistent with the previous implementation, I have not been able to reproduce the same numbers. My hypothesis is, that at some point the muons see a different amount of field which essentially changes their random numbers relative to the configuration on `master`.

Should the fields as seen by Geant be exactly the same?

jback08:
The fields should be exactly the same, although note that there was a bug in some tertiary fields not being set in the original version of `setMagnetField` in `python/geomGeant4.py`.

I am running the muon background sim (after doing clean FairShip builds) between the two versions and will try to see where the differences are occurring.

jback08:
Running over the full 18M event muon background sample and counting the number of (unique) muon tracks in the strawtube detector (points) I find 38 muons in the updated interface and 31 for the original. 

Looking at the some of the event outputs in detail, there are cases where (after thousands of OK events) in one of the constant field regions (e.g. a muon shield sub volume) there is a different electron or photon interaction, for example, which then creates a divergence with respect to the rest of the random number sequence. The fields appear to be exactly the same, so it is a mystery why this is happening.

We need a way to look at the detailed G4 stepping info for these specific events. Although enabling G4 tracking verbosity generates too large output files (since we need to run over 1000s of OK events to preserve the random number sequence ordering).

olantwin:
To make the muon shield sensitive you can use the `--FollowMuon` option to `run_simScript.py`. Note that this also changes the way non-muons are tracked, so if the problem disappears we might need to separate those two changes.

A much smaller (about 37 times) muon sample that was constructed to be biased towards high-$p$ and high-$p_T$ muons is available here: root://eoslhcb.cern.ch//eos/ship/user/olantwin/resampled.root . Let me know if you have trouble accessing it.

PR #50, opened 2017-08-23
Add Support for aliBuild

PMunkes:
This Pull Request is mainly to keep track of the work necessary to merge FairShip-Support into FairShip.


CMakeLists.txt
```diff
@@ -141,13 +136,19 @@ find_package(HEPMC)
...
-    set(GSL_DIR ${SIMPATH})
-  Else(${ROOT_LIBRARY_DIR} MATCHES /lib/root)
-    set(BOOST_ROOT ${SIMPATH}/basics/boost)
-    set(GSL_DIR ${SIMPATH}/basics/gsl)
-  EndIf(${ROOT_LIBRARY_DIR} MATCHES /lib/root)
+  IF(DEFINED ENV{SIMPATH})
+    If(${ROOT_LIBRARY_DIR} MATCHES /lib/root)
+      set(BOOST_ROOT ${SIMPATH})
+      set(GSL_DIR ${SIMPATH})
+    Else(${ROOT_LIBRARY_DIR} MATCHES /lib/root)
```
olantwin: Can this case be reached? It seems not.
In that case we should probably remove it.

PMunkes:
TODO: 
- [ ] Transfer all repositories to ShipSoft.
- [ ] Merge all commits from PMunkes/shipdist into aliBuild.
- [x] Upstream changes into FairRoot

CMakeLists.txt
```diff
@@ -304,19 +307,21 @@ MESSAGE(STATUS "Looking for GENIE ...")
 # genie in general
 FIND_PATH(GENIE  NAMES genie PATHS
   ${SIMPATH}/../FairSoft/generators
+  ${GENIE_ROOT}
   NO_DEFAULT_PATH
-)
+  )
```
olantwin: I don't think we need to change the indentation here.

field/ShipFieldCreator.cxx
```diff
@@ -50,7 +50,7 @@ FairField* ShipFieldCreator::createFairField()
 	// Initialise field
 	if ( fMagneticField ) {
 		fMagneticField->Init();
-		fMagneticField->Print();
+		fMagneticField->Print("");
```
olantwin: Why does this need to change?

CMakeLists.txt
PMunkes: I am not sure, but I am keeping it in there, to provide feature parity on FairSoft based installations.

field/ShipFieldCreator.cxx
PMunkes: Upstream FairRoot no longer allows calling the Print-function with an empty bracket.

olantwin: OK, good to know.

CMakeLists.txt
olantwin: I appreciate that you're trying to change as little as possible, but in this case it seems like a clear bug, unless I completely misunderstand this conditional, which we probably should not carry over.

CMakeLists.txt
```diff
@@ -309,7 +309,7 @@ FIND_PATH(GENIE  NAMES genie PATHS
   ${SIMPATH}/../FairSoft/generators
   ${GENIE_ROOT}
   NO_DEFAULT_PATH
-  )
+)
```
olantwin: You could probably just squash this into the previous commit.

PMunkes: Yes, but this can easily be done during the merge.

CMakeLists.txt
PMunkes: I accidentally deleted my comment to this section. Here it is reproduced:
I am not sure, why this is here, but I kept it to provide feature parity on FairSoft based installations.

olantwin:
I think all the changes in here are either already implemented by now or out of date, so I suggest we close this pull request?

ThomasRuf:
not anymore uptodate.

PR #51, opened 2017-09-25
Small changes to support lxplus7

olantwin:
This fixes both the issue with `lcqenv.sh` not being found and the `$PYTHONPATH` being wrong.
(Note I've changed both `sed`s to use `!` to avoid trouble with pathnames and to make the escaping unnecessary.)

There are many variables that are expanded that may contain spaces.
If I may I'll make a second commit to quote those properly and to make some small cosmetic changes.

### DIFF
--- a/localConfig.sh
+++ b/localConfig.sh
@@ -18,8 +18,8 @@ fi
 
 $SHIPBUILD/alibuild/alienv -w $SHIPBUILD/sw printenv FairShip/latest > config.sh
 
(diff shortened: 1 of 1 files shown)


olantwin:
Didn't see your commit. I'll redo the changes so that that they work on your updated code.

olantwin:
Github forced me to create a new pull-request, as reopening one does not seem to be possible if you forcepush...

So here we go: https://github.com/ShipSoft/FairShip/pull/52

PR #54, opened 2017-11-09
Use alibuild from system if installed

olantwin:


ThomasRuf:
If I understand correctly, this introduces a new workflow. The current workflow assumes user starts in a directory SHiPBuild, where FairShip is cloned, which then installs shipdist and alibuild in the same directory, if not yet exist.
The new proposed workflow takes alibuild out of this loop, and forces the user to put alibuild in the PATH environment. I don't see the advantage of this approach. If a user wants to have alibuild at a different place for whatever reason, I would propose that the user creates a symlink in SHiPBuild to alibuild.

olantwin:
Sorry for not explaining the intention originally (and I now noticed and correcte a mistake with the `$PATH`).

Anyway, the intended workflows are the following:

1. Basic usage: The user has now idea what `aliBuild` is and does not have it installed -> `which alibuild` fails and `alibuild` is cloned and added to the path (change from previous workflow: add to patch instead of using relative locations).
2. Advances usage: The user already has `alibuild` installed *and* in their `$PATH` -> assume the user knows what they are doing and warn them that `alibuild` is used from path, don't install another copy.

So in either case no action from the user is required.

ThomasRuf:
yes, but what happens when alibuild is called again to recompile FairShip after a change? If the user has not put alibuild on the path, it will again clone alibuild.

olantwin:
Ah, I see. I guess it then should check whether `alibuild` exists in the destination. I did not think about that case. Thank you for bringing it to my attention.

I will commit a correction.

olantwin:
The case where `alibuild` was previously cloned in now also considered (and I fixed another mistake I had made originally when setting the path.

So now on rebuilds it should also require no user interaction in either case.

ThomasRuf:
I am still hesitating to introduce this use case. What is the added value of having alibuild at an arbitrary location? Debugging user problems will only become more complex adding this possibility.

olantwin:
Since I'm not aware of any other SHiP users currently, here why I find this quite useful (and maybe others would as well):

* Installing `alibuild` in the `$PATH` via the standard install procedure allows using `alienv` directly to set up the SHiP environment without polluting the environment with any environment variables at all (or optionally with only `ALICE_WORK_DIR` for some more convenience). This is particularly useful if needed for startup scripts, or I imagine if users work on several experiments with conflicting environment variables etc..
* Until our `alibuild` procedure is fully stable and tested (see e.g. https://github.com/ShipSoft/shipdist/pull/1) I find it useful to have `alibuild` not too tighly coupled to the current `aliBuild.sh`
* I quite frequently have several `FairShip` installations on the same machine, e.g. if I quickly want to investigate something without disturbing my main installation. Installing `alibuild` several times seems unnecessary.

Again, this change would not change anything for the default use case. If somebody installs `alibuild` separately, some basic troubleshooting skills can be assumed, in my opinion. The `echo` statements also should make it trivial to determine which `alibuild` is used, if a user does encounter an issue.
If you like, we could add a sternly phrased warning as well, for the case were `alibuild` is discovered in `PATH`.

olantwin:
I guess an alternative would be that advanced users just forgo `aliBuild.sh` entirely and use `aliBuild ... build ...` directly, in which case I rest my case.

ThomasRuf:
If you have several FairShip versions, I would then propose to use one SHiPBuild, and reuse this SHiPBuild for all of your FairShip versions via localconfig.sh . Otherwise, do you have for each FairShip version a set of the whole ROOT/GEANT4/PYTHIA/... libraries?

olantwin:
Yes, that's something I didn't consider, as I only just moved to alibuild and so only used multiple FairShip installations with FairSoft.

Anyway, I think I will probably just use `ali{Build,Doctor,env}` directly for my desktop and laptop. From my side this pull request can be closed.

PR #59, opened 2017-12-31
Fix issue reported by Ievgen

olantwin:
Colon missing after if condition, second if without condition redundant/syntactycally wrong.


olantwin:
@ghkorol

olantwin:
Closed because 2b48f54 implements the same change.

PR #60, opened 2018-01-15
Random seed support

sashabaranov:


ThomasRuf:
thinking a bit more about the seed from time, I come to the conclusion that this is not quite safe. There is some probability that a job is started with an already existing random number. For the background mass production, do you plan to feed the seed as external parameter?
if you plan to use seed from time, we need to think a bit more before using it. 
 
thomas

ThomasRuf:
assuming we want a unique number for any job submitted within a month,  ~3000000 sec, and assuming jobs do not start within 1/100 of a second, than this might be better:  int((time.time()*100)%3000000).

what do you think?

sashabaranov:
> For the background mass production, do you plan to feed the seed as external parameter? 

Yes, we do. The easiest way would be just to use job id, which is unique ever increasing number in system. It can be easily [passed](https://github.com/skygrid/docker-worker/blob/master/dockerworker/worker/util.py#L44) to command line.

Maybe it would be easier to just use system entropy source? Seems like ROOT's SetSeed default does this: https://root.cern.ch/doc/v606/classTRandom.html#a3234e73bce2b3e0751cdb5bb79f5408d

ThomasRuf:
this looks interesting. I checked, we are using <ROOT.TRandom* object ("Random3") 
ROOT.gRandom.SetSeed(0)
ROOT.gRandom.GetSeed()
3104493518L
ROOT.gRandom.SetSeed(0)
ROOT.gRandom.GetSeed()
2415155258L

so, we just remove then seedFromTime, replace it with a default value = 0. Should I change it after your pull request, or you change it before?

Thomas

sashabaranov:
Seed is also used here:

```python
P8gen = ROOT.FixedTargetGenerator()
...
P8gen.SetSeed(args.seed)
```

I dont know what SetSeed does there

ThomasRuf:
yes, it is the same, see http://home.thep.lu.se/~torbjorn/pythia81html/RandomNumberSeed.html 

so. what you propose looks correct.
Thomas

PR #61, opened 2018-01-22
Add script to generate hit maps to FairShip.

olantwin:
Now that it is fairly stable I think it might be time to add the script to create hit maps to `FairShip`. Previously I was developing it in a separate repository, which is why the commit history is not imported.

Currently there are two open issues:

1. Since the SBT volumes have a complicated naming scheme all `vetoPoint` hits that can't be identified as one of the other veto detectors is identified as an SBT hit. Adding geometric cuts based on the geometry information could fix this (or another robust way to identify SBT hits).
2. I save the kinematics for all muon hits, and, for all muons that create at least one hit, the kinematics at the origin. Unfortunately this means the number of muons doesn't agree between the two sets of histograms.

Please let me know if you have any suggestions for changes etc.

### DIFF
--- a/macro/flux_map.py
+++ b/macro/flux_map.py
@@ -1,0 +1,244 @@
+#!/usr/bin/env python2
+import argparse
+import numpy as np
+import ROOT as r
+# Fix https://root-forum.cern.ch/t/pyroot-hijacks-help/15207 :
+r.PyConfig.IgnoreCommandLineOptions = True
+import shipunit as u
+import rootUtils as ut
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Script to create flux maps.')
+    parser.add_argument(
+            'inputfile',
+            help='''Simulation results to use as input. '''
+            '''Supports retrieving files from EOS via the XRootD protocol.''')
+    parser.add_argument(
+            'geofile',
+            help='''Geometry file to use. '''
+            '''Supports retrieving files from EOS via the XRootD protocol.''')
+    parser.add_argument(
+            '-o',
+            '--outputfile',
+            default='flux_map.root',
+            help='''File to write the flux maps to. '''
+            '''Will be recreated if it already exists.''')
+    args = parser.parse_args()
+    g = r.TFile.Open(args.geofile, 'read')
+    sGeo = g.FAIRGeom
+    f = r.TFile.Open(args.outputfile, 'recreate')
+    f.cd()
+    maxpt = 10. * u.GeV
+    maxp = 360. * u.GeV
+    h = {}
+
+    # Define histograms
+    for suffix, title in [('mu', '#mu#pm hits'), ('all', 'All hits')]:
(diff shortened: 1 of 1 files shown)


ThomasRuf:
1. Not sure I understand the use case. A vetoPoint has the full x,y,z information, with gGeoManager->FindNode(x,y,z); and either GetName() or GetVolume()->GetName(), the detector is identified. 
2. the kinematic of the muon at the origin is stored in MCTrack. Why should this be duplicated?

olantwin:
1. The detector names one gets this way for the SBT are not really meaningful to humans and too granular to e.g. automatically create a histogram per detector. I have tried various rules based on the name to improve this, but so far this has always led to problems with backwards compatibility. I am open to suggestions, but it's not as simple as just asking for the detector name. Maybe @Plamenna or @ghkorol have a suggestion?
One of the reasons I would like to add this script to FairShip is, that sooner or later the people implementing the detectors can ensure that the detector identification and the histogram dimensions/binning are appropriate for their detector, instead of me having to make educated guesses.

2. None of the information in these histograms is new. They all duplicate the information in the actual simulation results. The point here is to create histograms of the hits per detector and some additional summary statistics to give context. I expect for advanced users who are able to interrogate the simulation results directly, this script will at most serve as a guide on how to do certain things (e.g. how to get the weights of muons for the MuonBack production).

Finally, I upstream this so that other people can run this at their convenience for e.g. checking changes to the geometry of their subsystems etc.. But additionally, as the output of this script is going to be used for technical decisions, it would probably be good to have this script available for review to ensure there are no bugs etc..

ThomasRuf:
ok, it is only a macro filling histograms, no modification of event data. Leave it to the potential users for other comments.

olantwin:
Thanks!

Are you opposed to using GitHub issues? If not, we could continue the discussion about detector naming there. This is a problem where e.g. FairShip could provide a reusable function, that for a given FAIRGeom object and the detector ID figures out which detector it is in a human readable way.

ThomasRuf:
For digitized hits, this is solved by using a unique channel identifier, which encodes the high level detector ID. 
For the point classes, this is also solved by using a different class for each subdetector. 

I can only see maybe some issue with that vetoPoint is sometimes misused for other things, like sensitive active muon shield to be able to plot the trajectory of a muon.

olantwin:
Good point. Indeed, this is only a problem for vetoPoints, and should disappear there once UVT/Timing use their own points.

Maybe I should implement a new point type for the muon shield (essentially a clone of vetoPoint), to prevent confusion?

ThomasRuf:
Yes, maybe a generic ship point class would not be bad. it is not only muon shield, it is also used for rock, absorber, and maybe other things in future.

olantwin:
Shall we maybe open an issue for that? It's nothing urgent, but it would be good to not forget about it.

PR #62, opened 2018-01-23
Update readme and some cosmetic changes

olantwin:
This just updates some things in the readme to avoid confusion. The run instructions still need to be adjusted.

I also fixed some of the links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,13 @@
 
 ## Introduction
 
-FairShip is the software framework for the SHiP experiment which is based on FairRoot. To use this software you need to install three packages: FairSoft, FairRoot and FairShip. The first two
-pacakges are quite stable and you don't have to modify them. They will be updated
-infrequently only when the FairRoot team releases a new version. In such a case you will be
-warned and you have to rebuild them. All packages are managed in Git and GitHub. Please
-read [the Git tutorial for SHiP](https://github.com/ShipSoft/FairShip/wiki/Git-Tutorial-for-SHiP) first, even if you already know Git, as it explains how development is done on GitHub.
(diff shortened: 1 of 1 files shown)


olantwin:
Also, do we really need to keep the old install instructions? If someone needs to compile an old version of FairShip, the matching README is included.

ThomasRuf:
two comments:
1) removing old instructions, ok with me.
2) since you introduce the link to the git tutorial, there seems to be something wrong with its instructions, Alexander Korzenev reported a problem this morning:
git pull
>>> shipgen/GenieGenerator.h: needs update
...
seems to be due to using the  .gitconfig: In the section "Setting up Git on your AFS or local account" of the "Git tutorial for SHiP" it is written to copy

$ cp /afs/cern.ch/user/r/rdm/public/ShipGit/.gitconfig ~

If I remove .gitconfig from my home directory I do not see anymore the message about "refusing to pull with rebase: ..."

olantwin:
Hi Thomas,

1. OK, I'll push another commit. Please wait before you merge.
2. The link was already there. I just reformatted the paragraph in the plaintext. However, I agree, we should update the git tutorial. Having such nonstandard things in the `.gitconfig` is dangerous. Recently also ran into a related problem. I will update the wiki directly/separately.

olantwin:
I suggest we explicitly tell people how to set their name and email.

The git ignore we should just merge with the one we have already.

For the remaining settings defined I'll look into which ones are needed and which ones aren't and figure out how we best set them. I think they also can be set for the repository.

Maybe @FonsRademakers can comment on why

```
[branch]
	autosetuprebase = always
```

is needed. It seems that setting is the culprit to Alexander's problem.

PR #63, opened 2018-01-23
Fix code blocks

olantwin:
This fixes the code blocks in the README. Just a cosmetic change.


ThomasRuf:
i am now very confused. These are modified instructions for the old procedure? gcc62 does not exist anymore, was some temporary folder for testing gcc62.

PR #66, opened 2018-01-25
Remove superfluous UVTs

olantwin:
If I am not completely mistaken, in vessel configuration we currently still
create `VetoTimeDet` for vessel 6 (the one with the new iSHiP) even though we
should not.

Additionally, the conditional for `fDesign < 6` is superfluous, as it can only
be reached if `fDesign == 4`.


ThomasRuf:
mmh, why should this not work?
      //Add veto-timing sensitive plane before vacuum tube, not for new design 6, January 2018
      if (fDesign<6){

       TGeoVolume *VetoTimeDet = gGeoManager->MakeBox("VetoTimeDet",Sens,aO1+wallo/2.,6.*m,10.*mm);
...

olantwin:
If you look at the scope, that piece of code is only reached if `fDesign==4`. In all other cases the `VetoTimeDet` is defined elsewhere and the condition `fDesign<6` has no effect.

ThomasRuf:
OK.

PR #67, opened 2018-01-25
Separate liquid and plastic SBT

olantwin:
Use detector ID's to separate liquid and plastic SBT as explained to me by
Plamenna.

Next step would be to separate the two layers of plastic SBT.

There O(10) unidentified VetoPoints which don't seem to follow the detector ID
convention. Their source is being investigated.


ThomasRuf:
You might check if this unknown points originate from a counting station in the second hall, after muon detectors.

olantwin:
After some further discussion with Plamenna, I think it might be due to floating point errors when checking the detector name by using it's coordinates.

To be checked.

PR #68, opened 2018-01-26
Remove out-dates setup script

olantwin:
I think this file in FairShip also either needs updating or should be removed.
It still exports the locations of FairSoft/FairRoot.

But even if we do update, it essentially duplicates the installation
instructions in the readme, so I suggest we remove it, as implemented in this
pull-request.

I also remind you of the related SIMPATH issue, we will have to deal with.

### DIFF
--- a/shipsetup.csh
+++ b/shipsetup.csh
@@ -1,8 +1,0 @@
-# for setup at CERN lxplus
-setenv mySHIPSOFT /cvmfs/ship.cern.ch/ShipSoft  # should point to your own version 
-setenv SHIPSOFT /cvmfs/ship.cern.ch/ShipSoft
-setenv SIMPATH  ${SHIPSOFT}/FairSoftInst
-setenv FAIRROOTPATH ${SHIPSOFT}/FairRootInst
-setenv FAIRSHIP ${SHIPSOFT}/FairShip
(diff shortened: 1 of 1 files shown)


ThomasRuf:
did the following cross check, SIMPATH used in:
FairShip/configure.sh, script replaced by localBuild.sh or aliBuild.sh depending on use case. ok
FairShip/python/saveBasicParameters.py, with new build and configure scripts, "FAIRSHIP_HASH" should be in os.environ, therefore lines with SIMPATH not used anymore. ok
FairShip/CMakeLists.txt, if SIMPATH not defined switch to alibuild recipe. ok
FairShip/shipgen/CMakeLists.txt, adds ${SIMPATH}/include and ${SIMPATH}/lib to path, not needed, can be removed.
FairShip/shipgen/FixedTargetGenerator.cxx, if SIMPATH not defined take particle file from getenv("EVTGENDATA"), ok.

olantwin:
Hm, maybe something with my `lxplus` set-up is misconfigured. I installed via localBuild, but used the old FairSoft/FairRoot method before, so maybe there are some conflicts/remnants of the old method.

olantwin:
Indeed, it was some conflict with left-overs from the previous installation. Cleaned everything up and now localBuild works for me. Sorry for the noise.

PR #69, opened 2018-01-30
Timing Detector class is introduced

korzenev:


ThomasRuf:
This method will be needed later when the segmentation exists, to make the mapping between detectorID and path to the node in the detector tree.
thomas

ThomasRuf:
This will replace the implementation of the timedet in veto.cxx. No option to go back to the old version. Should be ok, since timedet wasn't really used heavily.

PR #70, opened 2018-02-08
Use named temporary file for event display file

olantwin:
This solves a(n addmittedly rare) problem that occurs when the eventDisplay is
run somwhere where one cannot create files. As to my knowledge this is a file
that is only used by FairRoot and contains nothing useful otherwise I suggest
we use a temporary file (as implemented here) until a use for that file is
found.

### DIFF
--- a/macro/eventDisplay.py
+++ b/macro/eventDisplay.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python -i
 import ROOT,sys,getopt,os,Tkinter,atexit
+import tempfile
 from ShipGeoConfig import ConfigRegistry
 from rootpyPickler import Unpickler
 from array import array
@@ -69,7 +70,6 @@ if not dy:
  except:
    tmp = None
 
-OutFile	      = "tst."+InputFile.split('/')[-1]
 if InputFile.find('_D')>0: withGeo = True
 
 def printMCTrack(n,MCTrack):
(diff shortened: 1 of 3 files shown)


ThomasRuf:
if i understand correctly, you removed physically the output file.
Q1: the same couldn't be done by commenting fRun.SetOutputFile(OutFile)?
Q2: did you checked what this file normally contains?

cheers,
Thomas

olantwin:
Q1: If I comment `fRun.SetOutputFile(OutFile)` there's a segmentation fault, but no error message. The Eve window shows up but is unusable.

Q2: As far as I can tell it contains the event header, and a list of branchnames and some additional metadata.

ThomasRuf:
what about having an option to specify the name of the output file?
OutFile = None
        if o in ("-o", "--output"):
             OutFile = a
             if a.lower()=='dummy': OutFile = "/dev/shm/output.root” 
if not OutFile:	      = "tst."+InputFile.split('/')[-1]

and for your specific case, you need to specify  -o dummy

ThomasRuf:
another suggestion by the FairRoot developers:
OutFile=ROOT.TMemFile("test", "RECREATE");

olantwin:
Concerning your first suggestion, I agree that giving the option to set the output file would be good, but I think for now I would suggest to default to a temporary file. Maybe when we start saving interesting information in it we can save it by default.

Since eventDisplay.py at the moment has a very limited number of options, I am very tempted to reimplement the options parsing in argparse (which I find a lot easier to maintain; even for trivial cases getopt can get very confusing). But this would be a different pull request. For now I'll add an option the old way.

I'll check whether TMemFile can be used in this case. I've used it in the past and discovered that it works for some FairRoot output files, but not for others, if I recall correctly.

olantwin:
TMemFile works, I will update the pull-request to use it instead of a temporary file in the case where no output file is set explicitly.

olantwin:
Implemented your suggestions with TMemFile as default. I have also fixed the other commandline options which were not correctly implemented (same problem as here: https://github.com/ShipSoft/FairShip/pull/32).

olantwin:
I have rebased this on the changes in #71, as those are necessary regardless of the changes here.

PR #72, opened 2018-02-16
Set TimeDet dimensions from the geometry config

olantwin:
This exposes the main settings to control the TimeDet dimensions to the
geometry config.

Note, as the depth of the detector depends in the spacing and bar depth it
needs to be adjusted via them, but I have added a the resulting depth to the
geometry config, so that it can be taken into account for the placement of
other detectors.

Currently TimeDet and the first part of the splitCal seem to overlap. This
needs to be fixed.

### DIFF
--- a/TimeDet/TimeDet.h
+++ b/TimeDet/TimeDet.h
@@ -47,6 +47,14 @@ class TimeDet: public FairDetector
 
     /** Sets detector position along z */
     void SetZposition(Double_t z) {fzPos = z;}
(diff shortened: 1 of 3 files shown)


olantwin:
Just some added info: If the x or y dimensions are adjusted this adjusts the overlap between bars. If they are set too large there is no overlap and the size will not be adjusted further. So the size that is set in `geometery_config.py` is the maximum size of the detector, but the actual size might be slightly smaller.

PR #74, opened 2018-02-20
[WiP] Upstream new muon shield

olantwin:
This pull request will upstream all changes necessary for the new muon shield and for the use of arbitrary muon shield configurations.

Many of the commits refactor the muon shield to facilitate later changes and allow for easier maintenance/understanding of the code. `TString` is used to replace `const char*` (almost?) everywhere, similarly I have tried to replace all C-style arrays with `std::array` or `std::vector` as suitable.

Additionally it includes the new hadron stopper.

To be done/decided before this can be merged:

- [x] Which muon shield should be the default, shall the new muon shield be hard-coded or read from file by default?
- [x] Only update hadron absorber for configs >= 8

Future things to consider:

* Should the hadron stopper implementation be split from the muon shield? As it's out of the scope of the optimisation and the design details will diverge further and further from the rest of the muon shield it might make sense to separate them.

### DIFF
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -280,8 +280,8 @@ with ConfigRegistry.register_config("basic") as c:
     # zGap to compensate automatic shortening of magnets
     zGap = 0.5 * c.muShield.dZgap  # halflengh of gap
     if muShieldDesign == 7:
-        c.muShield.dZ1 = 0.7*u.m
-        c.muShield.dZ2 = 1.7*u.m
+        c.muShield.dZ1 = 0.35*u.m + zGap
+        c.muShield.dZ2 = 2.26*u.m + zGap
         c.muShield.dZ3 = 2.0*u.m + zGap
         c.muShield.dZ4 = 2.0*u.m + zGap
         c.muShield.dZ5 = 2.75*u.m + zGap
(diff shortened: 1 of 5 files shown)


olantwin:
The Travis check has some networking issues. @ThomasRuf or @FonsRademakers, could you rerun it?

olantwin:
@ThomasRuf Rebase against your new changes and added the new configuration as a possible choice to `geometry_config.py`.

ThomasRuf:
Don't understand the question:  "Which muon shield should be the default, shall the new muon shield be hard-coded or read from file by default?" What means hard-coded or read from file? I would like to be able to run the 2016 version of the muon shield, and compare with any new version of the muon shield. What is now called the baseline muon shield should become the default for the 2018 geometry.

ThomasRuf:
should i start merging?

olantwin:
Now the hadron absorber is only updated for the new configurations.

Concerning your question:
There are two new configurations, 8 and 9. 8 is a fully configurable one via a geometry/config file that is read and is what is used during the optimisation. Configuration 9 sets the parameters for the new 2018 configuration in the code similar to how configuration 7 was defined. Configuration 7 is still usable.

How shall we proceed with configurations 1-4 and 6? 5 Should still work as the TP configuration, but I don't think any of the others have worked/have been maintained in the last years.

olantwin:
Let's wait for Travis to complete, but I think this should now be good to be merged.

olantwin:
Since some of these changes touch a lot of lines, I think it would be best to not squash, so that they remain understandable. Rebase and merge would probably be best.

PR #75, opened 2018-02-27
Fix filenames in README

olantwin:
Some filenames were not yet updated (or actually: partially updated, such that
input file and geofile didn't agree).


olantwin:
@ThomasRuf could we please merge this ahead of the Wednesday (ideally as soon as possible, as I hope people give FairShip a try before the starter kit).

PR #76, opened 2018-03-02
Give fastMuon default value

olantwin:
Without the default value, `if fastMuon` crashes, as the variable is undefined
if neither `--fastMuon` nor `--followMuon` are set.


ThomasRuf:
is there some typo, fastMuon FastMuon?

olantwin:
Oh, yes. Didn't notice that in the rush to fix it. Updated the pull request to only fix the original typo.

PR #78, opened 2018-03-20
Straighten coil gaps to simplify shapes

olantwin:
For the new muon shield configurations (8 and 9, the previous shield is unaffected) the coil gaps are straightened.

The effect on the muon flux should be negligible (but is being tested).

### Background

Originally a complicated shape was introduced to allow for enough space for the coils in the top and bottom corners of each magnet. Now it it clear that the coils don't need as much space and will be at the height of the beam axis (see the most recent designs from MISiS).

As a result the geometry can be simplified (which has the added benefit of simplifying the import into Fluka).

This geometry should now correspond exactly to what is implemented in the MISiS CAD designs.

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -159,13 +159,20 @@ void ShipMuonShield::CreateMagnet(TString magnetName,TGeoMedium* medium,TGeoVolu
     Int_t color[4] = {45,31,30,38};
 
     if (NotMagnet) {
-      coil_gap = gap;
-      coil_gap2 = gap2;
+       coil_gap = gap;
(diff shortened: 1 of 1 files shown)


olantwin:
@ThomasRuf If you could open up issues for this repository, that would be great.
I'd like to keep track of some future improvements of the muon shield code, which aren't urgent and don't affect the constructed geometry.

ThomasRuf:
sorry, no idea what you mean with "If you could open up issues for this repository".

ThomasRuf:
when i click on Issues, try to open, get, No results matched your search. No idea how to open an issue. Not very intuitive.
thomas

PR #79, opened 2018-03-20
Add QCD ffbar to Dark Photon production

amagnan:


ThomasRuf:
Using  pdgid for DP, 4900023, instead of 9900015 will break some generic signal analysis code, ShipAna.py, which assumes that a hidden particle id = 9900015. This could of course be fixed by providing a user argument to ShipAna, which defines which pid to use. But might be something people forget to do. Nevertheless, would like to understand why it is necessary to introduce a new number?

cheers,
Thomas

PR #84, opened 2018-04-18
Adjustable material frames around views for 2018 strawtubes geometry

DaniilSu:
They can be used with strawDesign=11 option

### DIFF
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -55,6 +55,8 @@ with ConfigRegistry.register_config("basic") as c:
     c.cave = AttrDict(z=0*u.cm)
     c.cave.floorHeightMuonShield = 5*u.m
     c.cave.floorHeightTankA = 4.2*u.m  # no overlapp with 4.5 -> 4.2
+    if strawDesign == 11:
+        c.cave.floorHeightTankA = 3.0*u.m # avoid overlap with Tr_1
     c.cave.floorHeightTankB = 2*u.m
 #
     #neutrino detector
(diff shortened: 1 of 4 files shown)


ThomasRuf:
I would suggest to merge design 10 and 11. Design 10 was not yet used, and was meant to be the design for the 2018 geometry. Therefore, design 11 should become design 10.

if you change a header file, you need to increase the ClassDef number, otherwise ROOT will have problems to read old data: ClassDef(strawtubes,4) -> ClassDef(strawtubes,5)

did you run an overlap check? run_simScript with debug =2  # 2 make overlap check

DaniilSu:
OK, I will make suggested changes. 
Yes, I performed the overlap check. I had to change floorHeightTankA to 3.0 m in order to avoid overlaps with decay volume.

DaniilSu:
By the way during the overlap check I've found this strange overlap:

Info in <TGeoNodeMatrix::CheckOverlaps>: Checking overlaps for tTauNuDet and daughters within 0.0001
Check overlaps:      [==========]  119002 [100.00 %]           00:00           
Info in <TGeoNodeMatrix::CheckOverlaps>: Number of illegal overlaps/extrusions : 2

=== Overlaps for FAIRGeom ===
 = Overlap ov00000: tTauNuDet/volTauNuMudet_1 overlapping tTauNuDet/Pillar1Vol_1 ovlp=10
 = Overlap ov00001: tTauNuDet/volTauNuMudet_1 overlapping tTauNuDet/Pillar1Vol_2 ovlp=10

Is it normal?

ThomasRuf:
no, it is not. i also saw it, and just informed Antonio.

PR #85, opened 2018-04-18
Fix for backward compatibility issue

DaniilSu:
Fix to previous commit which introduced adjustable frames around views of the straw tracker

### DIFF
--- a/ecal/ecalContFact.cxx
+++ b/ecal/ecalContFact.cxx
@@ -7,7 +7,7 @@
 
 ClassImp(ecalContFact)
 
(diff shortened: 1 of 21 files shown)


ThomasRuf:
I am confused. I would not have expected changes to strawtubes/strawtubes.cxx/h

DaniilSu:
I didn't make any changes in strawtubes.cxx/h 

It seems that git added changes from previous pull request to this one, which is strange, because you have already merged it to master...

DaniilSu:
Now it shows that only one file file was changed. Which is correctly, I think.

PR #87, opened 2018-04-19
Latest Muon flux files

evanherwijnen:
Including Goilath field map, new target and run_MufluxfixedTarget, updated geometry, drifttube scintillators


ThomasRuf:
for my curiosity, why do we have 6 different versions of tungsten and molybdenum? Is it the right time to change the target properties after we have made the big Yandex production? 

i am a bit worried about the changes in size/position and how this will affect the muon background simulation, where we have to position the muon/charm ... at the correct place. Defined by ship_geo.target.z0, which doesn't seem to be the right number anymore given the C++ changes:

Double_t zPos =  -4.59; 
top->AddNode(tTarget, 1, new TGeoTranslation(0, 0,fTargetZ - fTargetLength/2. + totLength+2.5)); 
in any case, there shouldn't be any hardcoded number. what is 2.5?

for run_simScript, particle gun use case: instead of having three different options, I would propose to have one option PD with an argument particleID.

ThomasRuf:
for PG, I propose to enlarge the use case and allow for an optional argument. If no argument given set pID = 22, as it was before.  

        if o in ("--PG",):
            simEngine = "PG"
            if a : pID=int(a) #not sure the check is correct for an existing argument, to be tested

if simEngine == "PG": 
  myPgun = ROOT.FairBoxGenerator(pID,1)

ThomasRuf:
for target density: What are the errors on the measured values?

evanherwijnen:
0.03 g/cm3 for TZM and 0.1 g/cm3 for W.

ThomasRuf:
aren't the measurements than agreeing within errors? main difference is compared to the default value we are using for tungsten.
Why are the dimensions so much different compared to the values we used in the simulation?

evanherwijnen:
I can take the average of those measurements.
The values are different because we don’t use Tantalum and have modified the size of W and TZM to keep the same number of interaction lengths for the target.

evanherwijnen:
Cleaner would be to have two files.
I’ll work on it.

evanherwijnen:
I solved this by adding a new option pID, now it is backwards compatible.

ThomasRuf:
Some more questions:
- why do we need two field map files?  field/GoliathFieldMap.root  files/GoliathFieldMap.root  

- run_simScript:
   can we remove print "pID= ",pID ?
  Estart and Eend are nowhere used as far as I can see. There is no explanation what they are, what units. 

- ShipTargetStation.cxx: still worried about the changes:
   Double_t zPos =  -4.59; 
  //Double_t zPos =  fTargetZ - fTargetLength/2.; 
  top->AddNode(tTarget, 1, new TGeoTranslation(0, 0,fTargetZ - fTargetLength/2. + totLength+2.5)); 
Don't understand these numbers, 4.59, 2.5. Why old zPos now commented?

My worry: this might contradict the value for start of target encoded in geometry_config.py, and the while muon background simulation will break down.

evanherwijnen:
Hi Thomas


-          files/GoliathFieldMap.root is the fieldmap in the format for VMC. I left it in case somebody wants to debug this, but it can be removed.

-          I removed the printout and added the use of Estart and Eend.

-          ShipTargetStation: fixed Double_t zPos=0. The commented line has been commented out since at least a year. The 2.5 is also a mistake, removed it. Sorry about that, these were remnants from when this file was used for both Ship and Muon flux targtets.

Cheers Eric

PR #90, opened 2018-04-25
Bug fixes for using VMC B fields.

jback08:
The VMC only correctly sets the magnetic fields for Geant4 when they are defined
in the C++ code where the geometry is defined. This means that additional
fields set using the addVMCFields() function python/geomGeant4.py are only
recognised by ROOT and not Geant4.

To correct this, we need to reuse the ConstructSDandField() function in TG4GeometryManager
so that fields defined in ShipFieldMaker are recognised by both ROOT and Geant4.
This requires that ShipFieldMaker inherits from TG4VUserPostDetConstruction(), where
we overload the virtual Construct() function to make the fields and assign them to volumes;
this effectively renames the earlier ShipFieldMaker::makeFields(inputFile) function.

Corrected bin calculations in ShipBFieldMap that can occur for certain binning choices.

Added field/convertMap.py as a general script for converting a field map text file
into ROOT format for FairShip use.


ThomasRuf:
trying to use VMC B field framework with a field map for the spectrometer magnet. 

First observation, centre of magnet needs to be inserted via txt file:
FieldMap Map files/MainSpectrometerField.root 0.0 0.0 3068.0 
this will cause problems. The centre of the magnet is defined in geometry_config as the mean between tracking station 2 and 3: c.TrackStation2.z + 0.5 * (c.TrackStation3.z - c.TrackStation2.z) . A better method would be to pass the centre of the field via fieldMaker = geomGeant4.addVMCFields('field/MainSpectrometerField.txt', centre, True) with centre = TVector3(0,0,ship_geo.Bfield.z)

Second: the field is created
ShipBFieldMap::readMapFile() creating field Map using file /media/ShipSoft/SHiPBuild/sw/ubuntu1710_x86-64/FairShip/master-1/files/MainSpectrometerField.root
x values: -300, 300, dx = 10
y values: -350, 350, dy = 10
z values: -500, 500, dz = 10
Total number of bins = 437431; Nx = 61, Ny = 71, Nz = 101
however, when trying to access it with bfield = ROOT.genfit.FairShipFields(), only zeros are returned:
>>> bfield.get(0,0,3068.,bx,by,bz)
>>> bx,by,bz
(0.0, 0.0, 0.0)

jback08:
We could add in the additional argument in addVMCFields to pass a (variable) position shift, but I think this would only work OK if we only had one volume defined in the text file; there is no way of knowing if other volumes defined in the text file also needs the same shift (or not).

Perhaps we can have some python code that can create the text file, writing out the centre positions using variable expressions, and this text file is then passed onto the addVMCFields() function?

Or we could modify the logic in ShipFieldMaker so that we have a function like defineVMCField() for each separate volume, but this will involve changes to the Construct function and potentially to the ordering of various internal function calls to make sure everything is defined before we pass this onto the VMC to Geant4 step via geom.ConstructSDandField() in addVMCFields().

It looks like the ROOT.genfit.FairShipFields() function only returns the global field, so if the field map is defined as a local one for the spectrometer then B = 0. If the defined field is global then there is a problem somewhere.

ThomasRuf:
changing the text file on the fly, I already do this for the ECAL geometry. But it is very unsafe. All geometry parameters are stored in a pickle file, which comes with the geofile. The information in the text file might not be in synch with the geofile. If we would have no changes in the geometry, be a running experiment, would be less of a problem. 

the comment about local / global, I don't understand. in the txt file, there is currently 
FieldMap Map files/MainSpectrometerField.root 0.0 0.0 3068.0 

will this make a global or local field?

jback08:
After defining 
FieldMap Map files/MainSpectrometerField.root 0.0 0.0 3068.0

a global field is set using

Global Map

and a local field is set using

Local volumeName Map

Otherwise the field map is not assigned to anything.

The 3068.0 offset should not be needed if the (local) field map is centred on the local co-ordinates of the spectrometer (with the caveat about mother-daughter volumes/relative positions mentioned earlier).

ThomasRuf:
I think local field will not work, since the magnetic field extends outside the volume of the spectrometer. So the txt file should in addition contain:
FieldMap Map files/MainSpectrometerField.root 0.0 0.0 3068.0
Global Map Wilfried

?? what is the meaning of the third argument, Wilfried? Just the name of the field?

jback08:
The example text file contains

Bell Wilfried 0.1436 3068.0 2 500.0

which defines a Bell-shaped field based on 

fMagField = ROOT.ShipBellField("wilfried", ...)
 
near the end of the python/shipDet_conf.py script. If you are replacing the Bell field with a field map then just ignore this. The addVMCFields() function will override any other global field settings if you use the Global keyword in the text file.

ThomasRuf:
this does not seem to be the case. I get

>>> gMC.GetMagField()
<ROOT.ShipBellField object ("wilfried") at 0x5596b75a37a0>
 
note, wilfried with lower case w. While Wilfried with upper case was used in the txt file.

ThomasRuf:
i think we don't understand the syntax. We use:
FieldMap Map files/MainSpectrometerField.root 0.0 0.0 3068.0
Global Map TP 

and get back:
Setting the global field
Adding field Map to global
Could not find the field TP
Settting Global field as the global field for gMC

something seems to be wrong

jback08:
Let me know the changes you have made to the FairShip code/scripts, and the field map files you are using (a download link?) and I will take a look at it.

jback08:
You just need

Global Map

since TP is not defined. 

The gMC.GetMagField() call returning the wilfried field looks like it is only accepting the definition done at the end of the python/shipDet_conf.py file. If the printout statement is at the end of this file then this is correct, since the VMC change is done later. If the printout is at the end of macro/run_simScript.py then it looks like there is a problem with setting the global field.

ThomasRuf:
is Map a key word or a name?

jback08:
Map is the name of the field you are defining. FieldMap is the keyword for a field map, Global is the keyword for the global field.

I am wondering if we should convert parsing the text file into actual C++ function calls that we use within the python script(s), or have both methods available.

ThomasRuf:
other question: Do I understand correctly, at the moment, there can only be one global field? If yes, we have an issue. One would have to combine mag field for dShip and iShip in one file, knowing their relative position before starting to create the geometry. I don't see at the moment how I can assign a field with large tails to a volume or assembly.

jback08:
The problem could be that the field map components are not ordered correctly: first z goes from zmin to zmax, then y is incremented, then x (repeating the z, then y, increments).

I will look into using function calls for defining fields, since this could allow us to use variables for setting offsets more easily, while allowing for the text file method as well.

By definition, there can only be one global field map in the VMC; see ConstructGlobalMagField() in

https://github.com/vmc-project/geant4_vmc/blob/master/source/geometry/src/TG4GeometryManager.cxx

However several fields can be combined into the single global one using linear superposition:

Global Field1 Field2 ... FieldN

Geant4 uses global coordinates to find the B field, independent of the position of volumes. What field is used depends on whether a suitable volume-field pairing is found at the given position or not.

PR #100, opened 2018-06-13
Fix missing header import

olantwin:
`#include <unordered_map>` was missing


ThomasRuf:
thanks. Any idea why this didn't caused a problem on my local machine, Ubuntu 17.10 and gcc 7.2, but on other machines, with same environment?

PR #102, opened 2018-06-13
Fix final overlap

olantwin:
Missed a factor of 2. Now the overlap check completes with none.

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -846,12 +846,12 @@ void ShipMuonShield::ConstructGeometry()
       if (fDesign >= 7 && fFloor > 0) {
 	// Only add floor for new shield
 	TGeoBBox *box5 = new TGeoBBox("shield_floor", 10 * m, fFloor / 2.,
-				      fMuonShieldLength / 2. - dZ);
+				      fMuonShieldLength / 2. - dZ - 1 * mm);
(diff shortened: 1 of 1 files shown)


ThomasRuf:
also there is no overlap reported I think there is a problem:
sensPlane_1                 : z=-6541.0000cm  dZ=    0.1000cm  [-6541.1000   -6540.9000]
CoatVol_1                    : z=-6808.0000cm  dZ=  276.0000cm  [-7084.0000   -6532.0000] 
AbsorberVol_1            : z=-6808.0000cm  dZ=  266.0000cm  [-7074.0000   -6542.0000] 

the CoatVol overlaps with the sensitive plane, for this reason no hits at all are registered. it is 10cm too long

olantwin:
The z dimensions should not have changed compared to the previous version (where there was already 20cm of concrete after the hadron absorber. Not sure whether you used that version for the fixed target production.

We should probably move the sensitive plane by 10cm in +z to accommodate for it.

ThomasRuf:
The Yandex production in December / January run with muon shield option 7. The sensitive plane was put after the last absorber magnet. At that time, no additional material existed. In order to have the new additional material, I had to change to option 9.  Don't like to add 10cm hardcoded. It is then better to add the sensitive plane after CoatVol_1.

olantwin:
Yes, I'd agree that we should not hardcode it. Moving after CoatVol_1 should work fine (but will only work for options > 7).

PR #104, opened 2018-06-21
[WiP] Update flux_map for new MuonBack weights

olantwin:
Should be a quite straightforward change to use the per-hit weights instead of the old per-event weights.

@DaniilSu: Could you please review to make sure this is correct?

### DIFF
--- a/macro/flux_map.py
+++ b/macro/flux_map.py
@@ -99,13 +99,11 @@ def main():
                             original_muon.GetStartY())
         except AttributeError:
             continue
-        weight = original_muon.GetWeight()
-        if weight == 0:
-            weight = 1.
         for hit in event.strawtubesPoint:
             if hit:
                 if not hit.GetEnergyLoss() > 0:
                     continue
+                weight = event.MCTrack[hit.GetTrackID()].GetWeight()
                 x = hit.GetX()
                 y = hit.GetY()
                 z = hit.GetZ()
@@ -129,6 +127,7 @@ def main():
             if hit:
                 if not hit.GetEnergyLoss() > 0:
                     continue
+                weight = event.MCTrack[hit.GetTrackID()].GetWeight()
                 x = hit.GetX()
                 y = hit.GetY()
                 px = hit.GetPx()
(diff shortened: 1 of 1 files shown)


DaniilSu:
Well, not exactly like this.

For the hit loops you should also demand this condition:
tid = hit.GetTrackID()
if tid > 0: weight = event.MCTrack[tid].GetWeight()
else: continue
if abs(pid) == 13:
	muon = True
	**muonid = tid**

For the part concerning original muons:
if muon:
	h['mu_p_original'].Fill(event.MCTrack[**muonid**].GetP(), weight)
	h['mu_pt_original'].Fill(event.MCTrack[**muonid**].GetPt(), weight)
	h['mu_ppt_original'].Fill(event.MCTrack[**muonid**].GetP(), event.MCTrack[muonid].GetPt(), weight)

Well, this is how I do the analysis and how I understood Thomas' hints.
May be he can also comment on this.

ThomasRuf:
hit.GetTrackID(), if the particle producing the hit falls below the cuts set for saving particles, it will not exist on the file, you will get a negative number. Which then causes an error when accessing MCTrack. This will specially happen for EM background with low energy electrons. In this case, one needs to go back to the MCTrack container and find the original muon and get the weight from this MCTrack.

olantwin:
Thanks to both of you, I will update this accordingly.

Unless the MCTrack access for negative numbers fails silently, I don't think it happens (but I'll make sure).

ThomasRuf:
if tid > 0: weight = event.MCTrack[tid].GetWeight()
else: continue 

this can only be a short term dirty solution. It will neglect all hits produced by particles below 100MeV,
fStack.SetEnergyCut(100.*u.MeV). 

A method is required to find the most probable muon in MCTrack and extract the weight. One should check all primary particles for muons, check that weight is != 1.

olantwin:
@ThomasRuf Maybe we can discuss a possible algorithm at the meeting?

ThomasRuf:
yes. ideally it should remember that it had been called for the same event, and return previous result. Depending on the number of minus ids, could be time consuming. If you know a way to do this …

olantwin:
Thanks!
Yes, it's something I've done before, so it's probably just a matter of remembering/looking up what I did.

PR #106, opened 2018-06-28
Place tShield in origin of SHiP coordinate frame explicitly

olantwin:
Previously due to a bug this was the case anyway (`totLength == 0.0`), but now this is done explicitly.

I guess one could modify all positioning to be relative to the centre of the `MuonShieldArea`, but this will be a bit more work.

### DIFF
--- a/passive/ShipMuonShield.cxx
+++ b/passive/ShipMuonShield.cxx
@@ -826,9 +826,8 @@ void ShipMuonShield::ConstructGeometry()
       Double_t dX1 = dXIn[0];
       Double_t dY = dYIn[0];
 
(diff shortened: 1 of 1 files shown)


ThomasRuf:
This does not change anything. The assembly is still at the wrong place:
MuonShieldArea_1            : z=    0.0000cm  dZ= 1802.0000cm  [-1802.0000    1802.0000] 

I am surprised that it works at all. Most of the nodes inside this assembly are in fact physically outside.

ThomasRuf:
the whole logic looks wrong. Things are placed inside the assembly using the global coordinates, instead of the local MuonShieldArea coordinates. Maybe it is better to remove this assembly at all and place everything to the top volume.

PR #107, opened 2018-07-02
Updating emulsion structure

antonioiuliano2:
Moreover, adding an option to separate active volumes for top and bottom layers


ThomasRuf:
just want to point out, if you change the material composition, old data might become invalid. I mean, if you try to read old data with new media file, the information will not fit. Since the same name is used, it will not cause any error. If this is a problem, a better option is to use a new name.

antonioiuliano2:
I see your point, but are medium and material informations not saved in the 'geometry' file? So, people can recover the emulsion composition from InspectMaterial() method after launching getGeoInformation.py. Of course, using a new name is a good idea, which solves the problem at the root.

ThomasRuf:
It would make sense if it is stored in the geofile. Could you check?

antonioiuliano2:
Just checked, it seems so. When I inspect the material with a new simulation:

Mixture NuclearEmulsion    Aeff=93.1469 Zeff=40.6412 rho=2.9 radlen=3.50876 intlen=52.2412 index=19
   Element #0 : AG  Z= 47.00 A=107.90 w= 0.588
   Element #1 : BR  Z= 35.00 A= 79.90 w= 0.321
   Element #2 : C  Z=  6.00 A= 12.01 w= 0.033
   Element #3 : O  Z=  8.00 A= 16.00 w= 0.017
   Element #4 : N  Z=  7.00 A= 14.01 w= 0.014
   Element #5 : H  Z=  1.00 A=  1.01 w= 0.000
   Element #6 : I  Z= 53.00 A=126.90 w= 0.024
   Element #7 : S  Z= 16.00 A= 32.06 w= 0.002
   Element #8 : SI  Z= 14.00 A= 28.08 w= 0.000
   Element #9 : NA  Z= 11.00 A= 22.99 w= 0.000
   Element #10 : SR  Z= 38.00 A= 87.62 w= 0.000

Instead, when I check on an old simulation geofile, but with the same new FairShip, I have:

Mixture NuclearEmulsion    Aeff=93.2653 Zeff=40.7133 rho=2.4 radlen=4.23008 intlen=63.3666 index=7
   Element #0 : AG  Z= 47.00 A=107.90 w= 0.599
   Element #1 : BR  Z= 35.00 A= 79.90 w= 0.323
   Element #2 : C  Z=  6.00 A= 12.01 w= 0.023
   Element #3 : O  Z=  8.00 A= 16.00 w= 0.029
   Element #4 : N  Z=  7.00 A= 14.01 w= 0.010
   Element #5 : H  Z=  1.00 A=  1.01 w= 0.000
   Element #6 : I  Z= 53.00 A=126.90 w= 0.015
   Element #7 : S  Z= 16.00 A= 32.06 w= 0.001
   Element #8 : SI  Z= 14.00 A= 28.08 w= 0.001
   Element #9 : NA  Z= 11.00 A= 22.99 w= 0.000
   Element #10 : SR  Z= 38.00 A= 87.62 w= 0.000

I can confirm that this is the old composition, density 2.4 g/cm^3. I do not know how the weights w are computed

PR #108, opened 2018-07-05
Fix bug and refactor getGeoinfo macro

shir994:
There was a bug in doloop, where key[0].GetNode was called, but key[0] was a string, not node. Also, copy/paste code was removed.


ThomasRuf:
Great, thanks. There is still something with using the option --volume, I think. It would also be good if in case the weight and capacity is asked for a specific volume, the calculation is not for volumes above in hierarchy, computation takes a long time.

PR #116, opened 2018-07-11
Updating geometry for July measurement.

antonioiuliano2:
A lot of modifications, moving from the initial simplified geometry setup to a setup more similar to the real one for the measurement of this month. Still simplified layout of detectors between Goliath and MuonTagger.

Change which involves also muon flux simulations: MuonTagger passive slabs now have the real thicknesses (80, 80, 40, 40, 40) and the hole in the middle.


ThomasRuf:
these changes are included in the pull request https://github.com/ShipSoft/FairShip/pull/119

PR #124, opened 2018-07-24
Implement RPC endpoints

nwojcicka:
Add method to `MuonTaggerHit` to lookup strip endpoints.

Currently z positions are hard-coded, but plan to look them up in the geometry in future.

We will try changing `drifttubeMonitoring.py` to make use of it.

@hushchyn-mikhail Could you test this and let us know whether it works for the pattern recognition?


ThomasRuf:
better not touching drifttubeMonitoring.py, i made a lot of changes which are not yet committed.

PR #125, opened 2018-07-24
Flip x for muon tagger segmentation

olantwin:
@nwojcicka : could you please have a look whether this makes sense?

### DIFF
--- a/charmdet/MuonTaggerHit.cxx
+++ b/charmdet/MuonTaggerHit.cxx
@@ -23,7 +23,7 @@ void MuonTaggerHit::EndPoints(TVector3 &vbot, TVector3 &vtop)
    /* gGeoManager-> */                 // Get Z position
    const float total_width = 182 * STRIP_XWIDTH + EXT_STRIP_XWIDTH_L + EXT_STRIP_XWIDTH_R;
    const float total_height = 114 * STRIP_YWIDTH + 2 * EXT_STRIP_YWIDTH;
-   const float x_start = -(total_width - EXT_STRIP_XWIDTH_R + EXT_STRIP_XWIDTH_L) / 2;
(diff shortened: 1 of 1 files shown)


nwojcicka:
@olantwin yes it makes sense to me - the x was the wrong way round before then.

ThomasRuf:
there is still something strange. Is it normal, that the start / end points are not constant in x (y)?
For hits with view = 1 (v  = (channelID-10000*s)/1000), 
example  xyz (bot), xyz (top)
 -41.4000091553 -49.4625015259 874.25 -42.2625045776 49.4625015259 874.25
y changes from -49.462 to +49.462, with varying x. slight tilt? 

for the hits with view = 0, 
 79.4499969482 18.1125011444 874.25 -79.3475036621 17.2500019073 874.25
they are tilted in both directions, x starts with 79.44 and goes to -79.34 and y from 18.11 to 17.25.

Is it correct that view = 0 are the horizontal strips, and view = 1 the vertical strips?

Do the tilts  come from the alignment?

olantwin:
We decided to use diagonally opposite corners of the strip for the first implementation.

We can easily modify this to use the midpoint, but then we lose a bit of information. Let me know what you prefer.

ThomasRuf:
i think this makes no sense. Don't see what information you are loosing. Most common use case, I want to know distance of point to the strip. This should not depend on if I am at the top or bottom point of the strip.

ThomasRuf:
how can i distinguish this from alignment correction?

ThomasRuf:
Anyhow, the following plot cannot be explained by this issue. This shows the distance to the track as function of the track position in X at station1 vor view = 1 (vertical strips). X=0 agrees with position of strip, but the strip at -100cm is off by -20cm and the strip at +100m is off by +20cm. Is there an issue with the overall length of a station? Distance between strips?  

image

PR #128, opened 2018-07-27
Fix strip pitch, use centre of strip instead of corners

olantwin:
@ThomasRuf : Fixes the strip pitch and uses strip centre.

@nwojcicka : Please check whether everything looks correct.

PR #129, opened 2018-07-27
Fix typos in #128

olantwin:


ThomasRuf:
it is just uncommon to start numbering with 1.

olantwin:
I agree, but I think having two competing naming conventions for the strips would be worse. This way at least everyone will agree on which strip is meant.

BTW, straw and drift-tube numbering also start at 1.

ThomasRuf:
in x, the station 1 is off by about -3.7cm. 
image

For y, don't get any peak. Either the y of the track is wrong, or of the RPC.

olantwin:
What does the distribution look like in y?

Can we do the alignment in drifttubeMonitoring for now?

ThomasRuf:
right vertical strips, left horizontal strips. First observation, most tracks are at positive y, which is strange. would point more to an issue with the track extrapolation. y is only known before the magnet. 
how sure are we with top and bottom for the numbering of the horizontal strips? 1 starts at top or bottom of a station?

image

olantwin:
For the left you're plotting residual y vs y(strip) or residual y vs y(track)?

ThomasRuf:
x axis is residual (distance to track), y axis is y coordinate of track

ThomasRuf:
which function? here the plot residual  vs y of strip
image

ThomasRuf:
one more plot, y of strip versus y of track, no correlation what so ever,

image

ThomasRuf:
y coordinate of station 1 versus y coordinate of station 2, nice correlation. AND, most entries at negative y, while in previous plots with y of track, most entries where at positive y. So the y of RPC and y of track have different signs. 

image

ThomasRuf:
here an event display which illustrates the problem
image
 nicely. In cyan, line and hits are yz projection.

olantwin:
I found the problem. In line 148 you need to use `(vtop[1]+vbot[1])/2` instead of `(vtop[0]+vbot[0])/2`, as the latter will always be 0 for horizontal strips (assuming no misalignment)

PR #133, opened 2018-08-23
Update broken external link

raux:
Updated a broken link reference that was moved from:

https://www-opera.desy.de/tracker.html

to

http://www-opera.desy.de/tracker.html


ThomasRuf:
Oliver, do you recommend to merge it?

olantwin:
There's no harm in merging it. Since there is no sensitive information transferred, HTTP is sufficient.

PR #140, opened 2018-09-14
Clustering

simoniel:
Reconstruction for SplitCal:
- clustering performed on digitised hits (cells) in python
- cluster class added 
Tested on simple photon events, it seems to work well so far

### DIFF
--- a/python/shipDigiReco.py
+++ b/python/shipDigiReco.py
@@ -291,6 +291,8 @@ class ShipDigiReco:
    # hit selection
    # step 0: select hits above noise threshold to use in cluster reconstruction  
    noise_energy_threshold = 0.002 #GeV
+   #noise_energy_threshold = 0.0015 #GeV
+   #noise_energy_threshold = 0.001 #GeV
    list_hits_above_threshold = []
    print '--- digitizeSplitcal - self.digiSplitcal.GetSize() = ', self.digiSplitcal.GetSize()  
    for hit in self.digiSplitcal:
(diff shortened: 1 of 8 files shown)


simoniel:
The only modified files are:
- python/shipDigiReco.py 
- splitcal/*

for the other files I had some CRLF/LF problems, but they are not really modified w.r.t the official master

olantwin:
Then I think we need to find a way to remove the CRLF/LF changes. I will try to figure out what the options are...

For troubleshooting this issue in the future, could you send me your `.../FairShip/.git/config` and `~/.gitconfig`?

olantwin:
First, I would suggest doing a `git rebase -i` master and dropping all of the commits that don't contain any changes you want to keep.

Then we can see about the remaining commits.

ThomasRuf:
where are we with this request? Any success in cleaning up?

olantwin:
I have not hear form @simoniel , but would hold off merging until at least the easy nonsense commits are removed.

simoniel:
Hi @olantwin
sorry for the very slow reply I was on vacation last week without internet, so I am seeing these messages only now.
At the start I followed the instructions for gitconfig file but it never worked for me and I still get problems with  CRLF/LF...  I will send you the files. I am usually not able to attend the software meeting since I have another meeting that I cannot skip at the same time. 
I will come back to this MR tomorrow 
Cheers,
Rosa

olantwin:
Hi @simoniel 

Which instructions for the `.gitconfig`?

At least now that the commits are in, rebasing should allow us to cleanly get rid of those commits that are only line endings. 

Are you at CERN today? I feel this is something that could be solved much more efficiently in person.

Cheers,
Oliver

olantwin:
@ThomasRuf The line endings issue is solved now. Could you please end an email to the software mailing list to recommend **NOT** using the `.gitconfig` that was recommended in the past (at least not on lxplus)? This is now the 4th or 5th person with the exact same git issues due to it.

@simoniel would also like to present the changes to the software meeting at some point. Maybe we can have one on a different day/at a different time sometime?

simoniel:
Hi @olantwin  @ThomasRuf 

is there any other problem with my commit? 
I am happy to present the logic of the code and what I have done in one meeting for a wider discussion

Cheers,
Rosa

ThomasRuf:
Yes, would be good to have a presentation in a software meeting. When are you available? 

Looking at the changes in shipDigiReco, don't understand the logic of dTree. Do you propose to write to output files? Why are these new containers not in the know cbmsim tree?

don't like ROOT.gROOT.ProcessLine( this struct needs just to save few variables in a flat ntuple ... , in production code. 

I have the impression that your changes in shipDigiReco will break any usage of existing files, or being able to continue using the ecal shashlik option.

ThomasRuf:
concerning Linkdef.h, why do you need to make these dictionaries? I don't think it is needed. Where is it used?

#pragma link C++ class std::vector<double >+; 
#pragma link C++ class std::vector<splitcalHit >+; 
#pragma link C++ class std::vector<splitcalHit* >+;

ThomasRuf:
coming back to processline. What you need is a recoSplitcal class with data members E,eta, phi, .., it looks very cumbersome to have this flat ntuple, c structure object. Doesn't look very flexible.  

Do you keep track which hits contribute to a cluster? How do you make the MC truth association?

olantwin:
> concerning Linkdef.h, why do you need to make these dictionaries? I don't think it is needed. Where is it used?
> 
> #pragma link C++ class std::vector+;
> #pragma link C++ class std::vector+;
> #pragma link C++ class std::vector<splitcalHit* >+;

I agree that this should not be needed, as ROOT nowadays should work fine with STL containers. Can you test whether it works when you remove these lines?

simoniel:
thank you for you comments! Indeed probably it would be easier to discuss the changes in a meeting. 
To start answer your comments:
- I do no think this changes will prevent anyone to use the old calorimeter, the splitcal digitisation  (a part few modifications) is already included in the current master (I committed it in april/may) and noone complained. I did not touch or modify any of the old digitasation/clustering I just created alternative functions.  
- the dtree is there mostly for my tests during development, trying to add the new branch to the cloned cmb tree did not work for me on lxplus (no errors or crashes but at the end no new branch in present in the tree). As soon as a create a new tree the same commands worked just fine. 
- the cluster struct is needed only for quickly filling the tree, there are a couple of lines commented where i try just to create a branch for the full cluster class (as done for the others) but it does not work. I think that the problem is that one of the data member of the cluster class is a vector of another class (splitcalHits) and this starts to become too complex for ROOT. Maybe I am wrong but it is the only difference I found and I needed something working quickly. If someone as a solution I am happy to give it a try. This is also the reason of the extra dictionaries in the Linkdef.h: I was trying to see if adding them would fix the problem with the branch but it did not work, so I fill the struct with just the variables i needed. Anyway one can remove those dictionaries safely (they are just a remnant of the test). 
- MCtruth association: I have implemented methods from the cluster to get the associated hits and form this to get the point. From here one should be able to get the truth information in the usual way. Implementing a method to get this directly is on my to-do-list together with other improvements (e.g. divide more clearly digi and clustering + tuning on the clustering itself). This commit is not intended as final version of the digi+clustering but more as a milestone. 
I am happy to discuss the changes further in a meeting!
Cheers,
Rosa

ThomasRuf:
Dear Rosa,
I should be available this week on Tuesday and Wednesday, most of the day. 
I would propose to split your pull request in two parts, one, which only affects splitcal code, and a second regarding changes to shipdigireco. For the latter, I propose to put the splitcal digi and splitcal reco code into shiptasks, two separate python files, which can be optionally called by shipdigireco, depending if splitcal is present or not. 

concerning your points above:
1) your changes in shipdigireco do not handle the case: old ecal, no splitcal. It immediately starts creating a dtree, adding splitcal branches, …  This needs to be changed. 

2) the issue of not seeing your new containers in the standard tree needs to be fixed, and the creation of a second tree should be suppressed. I can help here. 

3) The struct object needs to be replaced by a proper class. I can see problems if many jobs start to write header files to disk in the same directory.

PR #144, opened 2018-09-20
Don't write to 'dummy' files, just write to memory

olantwin:
This changes all instances where a "dummy" file is created for no reason to write a file solely in memory and never to disk.

This also allows the simplification of `macro/inspectGeant4Geo.py` as `atexit` is no longer needed.


ThomasRuf:
In principle a good idea. But given the memory problems we have from time to time, what does it cost?
thomas

olantwin:
All `dummy` files on my systems seem to be of order KB. I would estimate the upper bound of the memory cost to be about the same, probably somewhat less, as some of this size will be due to the (Fair)ROOT file format and not everything written to a file will disappear from memory immediately.

I think in no cases is anything actually written to these files by us directly.

PR #145, opened 2018-09-24
rpc digitisation

nwojcicka:
Muon tagger digitisation for the simulation data

### DIFF
--- a/python/MufluxDigiReco.py
+++ b/python/MufluxDigiReco.py
@@ -12,7 +12,6 @@ matplotlib.use('pdf')
 import matplotlib.pyplot as plt
 import numpy as np
 
-from numpy.linalg import inv
 from sets import Set
 
 stop  = ROOT.TVector3()
@@ -24,34 +23,34 @@ start = ROOT.TVector3()
 #sGeo = fgeo.FAIRGeom
 
 #function for calculating the strip number from a coordinate
-def StripX(x):
-
-        #defining constants for rpc properties
-        STRIP_XWIDTH = 0.8625 #internal STRIP V, WIDTH, in cm
-        EXT_STRIP_XWIDTH_L = 0.9625 #nominal (R&L) and Left measured external STRIP V, WIDTH, in cm (beam along z, out from the V plane)
-        EXT_STRIP_XWIDTH_R = 0.86 #measured Right external STRIP V, WIDTH,in cm (beam along z, out from the V plane)
-        V_STRIP_OFF = 0.2000
-	NR_VER_STRIPS = 184
-	total_width = (NR_VER_STRIPS - 2) * STRIP_XWIDTH + EXT_STRIP_XWIDTH_L + EXT_STRIP_XWIDTH_R + (NR_VER_STRIPS - 1) * V_STRIP_OFF
-        x_start = (total_width - EXT_STRIP_XWIDTH_R + EXT_STRIP_XWIDTH_L) /2
-
-        #calculating strip as an integer
-        strip_x = (x_start - EXT_STRIP_XWIDTH_L + 1.5 * STRIP_XWIDTH + V_STRIP_OFF - x)//(STRIP_XWIDTH + V_STRIP_OFF)
-
-        return int(strip_x)
-
-def StripY(y):
-
-        STRIP_YWIDTH = 0.8625 #internal STRIP H, WIDTH, in cm
-        EXT_STRIP_YWIDTH = 0.3  #measured external STRIP H, WIDTH, in cm
-        H_STRIP_OFF = 0.1983
-        NR_HORI_STRIPS = 116
-	total_height = (NR_HORI_STRIPS - 2) * STRIP_YWIDTH + 2 * EXT_STRIP_YWIDTH + (NR_HORI_STRIPS - 1) * H_STRIP_OFF
-	y_start = total_height / 2
-       
-	strip_y = (y_start - EXT_STRIP_YWIDTH + 1.5 * STRIP_YWIDTH + H_STRIP_OFF - y)//(STRIP_YWIDTH + H_STRIP_OFF)
+	def StripX(x):
 	
-	return int(strip_y)
+        	#defining constants for rpc properties
+        	STRIP_XWIDTH = 0.8625 #internal STRIP V, WIDTH, in cm
+        	EXT_STRIP_XWIDTH_L = 0.9625 #nominal (R&L) and Left measured external STRIP V, WIDTH, in cm (beam along z, out from the V plane)
+        	EXT_STRIP_XWIDTH_R = 0.86 #measured Right external STRIP V, WIDTH,in cm (beam along z, out from the V plane)
+        	V_STRIP_OFF = 0.2000
+		NR_VER_STRIPS = 184
+		total_width = (NR_VER_STRIPS - 2) * STRIP_XWIDTH + EXT_STRIP_XWIDTH_L + EXT_STRIP_XWIDTH_R + (NR_VER_STRIPS - 1) * V_STRIP_OFF
+        	x_start = (total_width - EXT_STRIP_XWIDTH_R + EXT_STRIP_XWIDTH_L) /2
+
+        	#calculating strip as an integer
+        	strip_x = (x_start - EXT_STRIP_XWIDTH_L + 1.5 * STRIP_XWIDTH + V_STRIP_OFF - x)//(STRIP_XWIDTH + V_STRIP_OFF)
+
+        	return int(strip_x)
+
+	def StripY(y):
+
+        	STRIP_YWIDTH = 0.8625 #internal STRIP H, WIDTH, in cm
+        	EXT_STRIP_YWIDTH = 0.3  #measured external STRIP H, WIDTH, in cm
+        	H_STRIP_OFF = 0.1983
+        	NR_HORI_STRIPS = 116
+		total_height = (NR_HORI_STRIPS - 2) * STRIP_YWIDTH + 2 * EXT_STRIP_YWIDTH + (NR_HORI_STRIPS - 1) * H_STRIP_OFF
+		y_start = total_height / 2
+       
+		strip_y = (y_start - EXT_STRIP_YWIDTH + 1.5 * STRIP_YWIDTH + H_STRIP_OFF - y)//(STRIP_YWIDTH + H_STRIP_OFF)
+	
+		return int(strip_y)
 
 class MufluxDigiReco:
     " convert FairSHiP MC hits / digitized hits to measurements"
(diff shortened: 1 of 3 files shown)


olantwin:
@ThomasRuf

Please hold off merging until I've checked that everything looks good.

ThomasRuf:
some comments from my side:
1) don't see a reason to change run_simScript
2) digitizeMufluxSpectrometer, although you only added some text, digitizing only abs(pid)==13 is a mistake. This needs to be modified. 
2) MufluxDigiReco.py, same mistake as above. Why only digitize hits produced by muons?
3) zcoord is not used, hardcoded numbers should be removed. 
4) don't understand why this is needed: strip < 0 and strip > 185 : break. Shouldn't the function StripX be save enough? Same for StripY.

python/MufluxDigiReco.py
```diff
@@ -151,6 +189,115 @@ def digitize(self):
...
+		clusters = [[DetectorID_list[0]]]
+		for x in DetectorID_list[1:]:
+			if abs(x - clusters[-1][-1]) <= 1: 
+				clusters[-1].append(x)
+			else: 
+				clusters.append([x])
+		cluster_size = [len(x) for x in clusters]
+		print cluster_size
+		for i in cluster_size:
+			rc = h['muontagger_clusters'].Fill(i)
```
olantwin: Could you add some example histograms to the pull request for the clusters? I find this counting algorithm hard to follow.

python/MufluxDigiReco.py
```diff
@@ -182,8 +329,8 @@ def digitizeMufluxSpectrometer(self):
             pid = MufluxHit.PdgCode()
             xcoord = MufluxHit.GetX()
             ycoord = MufluxHit.GetY()
-            if abs(pid)==13:
-                if (detector[0:8]=="gas_12_1"):
+            if abs(pid)==13:            #selecting particle type as muon
+                if (detector[0:8]=="gas_12_1")               :#identifying which volume and converting into station number and filling (what is h????)
```
olantwin: The colon should follow the parentheses.

nwojcicka:
Thanks for the comments, I agree and will fix these asap.

python/MufluxDigiReco.py
```diff
@@ -239,19 +238,10 @@ def digitizeMuonTagger(self):
...
-		elif station == 3: zcoord = 1004.25
-		elif station == 4: zcoord = 1049.25
-		elif station == 5: zcoord = 1094.25
-
                 #calculate strip
                 # x gives vertical direction
                 direction = 1
                 strip = StripX(xcoord) 
-		if strip < 0 and strip > 185 :
-			break
```
olantwin: I would move these checks into StripX/Y instead of removing them completely. Maybe as assertions.

python/MufluxDigiReco.py
```diff
@@ -24,34 +23,34 @@
...
+        	H_STRIP_OFF = 0.1983
+        	NR_HORI_STRIPS = 116
+		total_height = (NR_HORI_STRIPS - 2) * STRIP_YWIDTH + 2 * EXT_STRIP_YWIDTH + (NR_HORI_STRIPS - 1) * H_STRIP_OFF
+		y_start = total_height / 2
        
-	strip_y = (y_start - EXT_STRIP_YWIDTH + 1.5 * STRIP_YWIDTH + H_STRIP_OFF - y)//(STRIP_YWIDTH + H_STRIP_OFF)
+		strip_y = (y_start - EXT_STRIP_YWIDTH + 1.5 * STRIP_YWIDTH + H_STRIP_OFF - y)//(STRIP_YWIDTH + H_STRIP_OFF)
 	
-	return int(strip_y)
+		return int(strip_y)
```
olantwin: I think these are now all one indentation level too deep, if I'm not mistaken.

olantwin: The functions are defined in the global scope, but indented as if they weren't.

python/MufluxDigiReco.py
nwojcicka: I'm not sure how to limit the values without throwing an error - that's why this was in the main event loop before

nwojcicka:
> could you please remove the change in run_simScript.py?

I just removed it.

python/MufluxDigiReco.py
```diff
@@ -20,6 +22,33 @@
...
+	EXT_STRIP_XWIDTH_L = 0.9625 #nominal (R&L) and Left measured external STRIP V, WIDTH, in cm (beam along z, out from the V plane)
+	EXT_STRIP_XWIDTH_R = 0.86 #measured Right external STRIP V, WIDTH,in cm (beam along z, out from the V plane)
+	V_STRIP_OFF = 0.200
+	NR_VER_STRIPS = 184
+	total_width = (NR_VER_STRIPS - 2) * STRIP_XWIDTH + EXT_STRIP_XWIDTH_L + EXT_STRIP_XWIDTH_R + (NR_VER_STRIPS - 1) * V_STRIP_OFF
+	x_start = (total_width - EXT_STRIP_XWIDTH_R + EXT_STRIP_XWIDTH_L) /2
+
+        #calculating strip as an integer
+	strip_x = (x_start - EXT_STRIP_XWIDTH_L + 1.5 * STRIP_XWIDTH + V_STRIP_OFF - x)//(STRIP_XWIDTH + V_STRIP_OFF)
+      	return int(strip_x)
```
olantwin: This line needs to be indented less.

Also, could you add a check here to make sure only valid strip numbers are returned?

python/MufluxDigiReco.py
```diff
@@ -151,6 +186,104 @@ def digitize(self):
         self.digitizeMufluxSpectrometer()
         self.digiMufluxSpectrometerBranch.Fill()
 
+        self.digiMuonTagger.Delete()
+        self.digitizeMuonTagger()
+        self.digiMuonTaggerBranch.Fill()
+
+
+
+    def digitizeMuonTagger(self):
```
olantwin: The indentation of this function is very confusing. Every few lines there are a few spaces more or less.

python/MufluxDigiReco.py
```diff
@@ -151,6 +186,104 @@ def digitize(self):
...
+	    #translation to local frame
+	    #point_muonbox = array('d', [MuonTaggerHit.GetX(), MuonTaggerHit.GetY(), MuonTaggerHit.GetZ()])
+	    point_local = array('d', [0,0,0, 1])
+            master_matrix.MasterToLocal(point_muonbox, point_local)
+            
+	    xcoord = point_local[0]
+	    ycoord = point_local[1]
+	   	
+            #selecting muons
+            if abs(pid) == 13 :
```
olantwin: We should not select muons here.

python/MufluxDigiReco.py
olantwin: I don't see anything wrong with throwing an error (e.g. through an insertion), as silent failure could be worse. Just make sure it gives an understandable error (e.g. `assert should_be_true == True, "somethings wrong"`)

python/MufluxDigiReco.py
```diff
@@ -20,6 +22,33 @@
...
+	STRIP_XWIDTH = 0.8625 #internal STRIP V, WIDTH, in cm
+	EXT_STRIP_XWIDTH_L = 0.9625 #nominal (R&L) and Left measured external STRIP V, WIDTH, in cm (beam along z, out from the V plane)
+	EXT_STRIP_XWIDTH_R = 0.86 #measured Right external STRIP V, WIDTH,in cm (beam along z, out from the V plane)
+	V_STRIP_OFF = 0.200
+	NR_VER_STRIPS = 184
+	total_width = (NR_VER_STRIPS - 2) * STRIP_XWIDTH + EXT_STRIP_XWIDTH_L + EXT_STRIP_XWIDTH_R + (NR_VER_STRIPS - 1) * V_STRIP_OFF
+	x_start = (total_width - EXT_STRIP_XWIDTH_R + EXT_STRIP_XWIDTH_L) /2
+        #calculating strip as an integer
+	strip_x = (x_start - EXT_STRIP_XWIDTH_L + 1.5 * STRIP_XWIDTH + V_STRIP_OFF - x)//(STRIP_XWIDTH + V_STRIP_OFF)
+	assert (0 < strip_x > 185),"X strip outside range!"
```
olantwin: This is equivalent to `0 < 185 < strip_x`. I think you'll need to flip the `>` to be a `<`

python/MufluxDigiReco.py
```diff
@@ -20,6 +22,33 @@
...
+def StripY(y):
+
+	STRIP_YWIDTH = 0.8625 #internal STRIP H, WIDTH, in cm
+	EXT_STRIP_YWIDTH = 0.3  #measured external STRIP H, WIDTH, in cm
+	H_STRIP_OFF = 0.1983
+	NR_HORI_STRIPS = 116
+	total_height = (NR_HORI_STRIPS - 2) * STRIP_YWIDTH + 2 * EXT_STRIP_YWIDTH + (NR_HORI_STRIPS - 1) * H_STRIP_OFF
+	y_start = total_height / 2
+	strip_y = (y_start - EXT_STRIP_YWIDTH + 1.5 * STRIP_YWIDTH + H_STRIP_OFF - y)//(STRIP_YWIDTH + H_STRIP_OFF)
+	assert (0 < strip_y > 117),"Y strip outside range!"
```
olantwin: Same here

olantwin:
I think this is now pretty much ready?

@ThomasRuf: Shall we leave the "clustering" in?

@ThomasRuf: If you squash the changes on merge, run_simScript.py should not be touched.

ThomasRuf:
I will try it out.

ThomasRuf:
There are at least two issues with this code:
1) enormous amount of printout
[10079, 10080, 11075, 11076, 20079, 20080, 21076, 30080, 31075, 31076, 40079, 40080, 40081, 41075, 41076, 41077, 50077, 50078, 50079, 50080, 51075, 51076, 51077] 
[2, 2, 2, 1, 1, 2, 3, 3, 4, 3] …..

2)   after a few events:  assert (0 < strip_y < 117),"Y strip outside range!"
AssertionError: Y strip outside range!

I have to switch it off until this is fixed.

olantwin:
Sorry, I will prepare a pull request to fix this.

For 2., how do we want the digitisation to fail? Not digitise such an invalid hit? Return some special value?

ThomasRuf:
how can it happen that the digitization fails? Makes no sense. If the sensitive volume in the simulation is bigger than reality, than it should be modified, or no digitized hit made.

olantwin:
It was modified, so it should not happen anymore with new simulation. Will not make any digitized hits in that case and print a warning.

olantwin:
Fixed, see #149

PR #146, opened 2018-09-25
rpc dimensions fixed

nwojcicka:
Adjusted the muon tagger dimensions to match the real detectors.


ThomasRuf:
what will be the effect of this change? By how much will the positions of hits be moved?

nwojcicka:
It's only the size of the rpcs that will change (by 7 cm in x and 9 cm in y), so that the sizes of the detectors match. This is needed for the correct calculation of the strips.

ThomasRuf:
yes, I understand. So, by how much do the positions change?

nwojcicka:
by 7 cm in x and 9 cm in y

olantwin:
This only affects the simulation, as the RPC in simulation is too large. This results in non-nonsensical strip numbers, as no strips exist where the hits were simulated.

The positions do not change, only the transverse size. This is assuming that the RPC is correctly centered in X-Y.

PR #147, opened 2018-09-27
P.O.T. simulation for charm geometry

antonioiuliano2:
Similar macro as run_MufluxfixedTarget.py, which activates the volumes for the charm measurement instead of the ones for the muon flux measurement.

Moreover, needed to add a method in FixedTargetGenerator, that makes it look for charm target volume instead of the standard SHiP target. The charm target part is just as in Pythia8Generator.cxx.
If the method is not called, FixedTargetGenerator behaves as before.

Best regards,
Antonio


olantwin:
Could we not modify `run_MufluxfixedTarget.py` instead of creating a new macro?

antonioiuliano2:
> Could we not modify `run_MufluxfixedTarget.py` instead of creating a new macro?

Yes, I think we can. We can put similar conditions as python/charmDet_conf.py to distinguish the two cases. I will upload a draft by tonight.

ThomasRuf:
i planned to simplify mufluxfixedtarget. there is no need to instantiate all modules by hand, this can be done by just using the charmdet config file.

antonioiuliano2:
Thank you very much for the commit to MufluxfixedTarget, Thomas. This really simplifies the structure.
I have updated this branch and added the if condition for charm geometry.

muonShieldOptimization/run_MufluxfixedTarget.py
```diff
@@ -20,6 +20,7 @@
 npot  = 5E13
 nStart = 0
 
+CharmdetSetup = 0 # 1 charm cross section setup, 0 muonflux setup
```
olantwin: This is unnecessary. You should be able to use `args.CharmdetSetup` everywhere. You can also give `ap.add_argument` an optional `default=0` argument.

antonioiuliano2: Ok, I can edit that. I saw that the other arguments are always stored in an ancillary variable, so I just assumed that this was the standard policy.

olantwin: I think it's just inherited from the way `getopt` did it. I would recommend using the `default=` option to `add_argument` and using `args.SomeArgument` directly if the scope allows.

antonioiuliano2: I see, thank you for the explanation. I have changed this part as you suggested

olantwin: Thanks!

PR #149, opened 2018-09-28
Fix rpc digitisation verbosity, skip invalid strips instead of erroring out; some cosmetics.

olantwin:


olantwin:
Also just noticed another issue with the expansion if the TClonesArray (fixed in the last commit)

PR #156, opened 2018-10-02
splitcal classes

simoniel:
commit of only the splitcal C++ classes


ThomasRuf:
splitcalCluster.cxx/.h seems to be missing

PR #158, opened 2018-10-05
New muflux rpc

evanherwijnen:
Segmentation of Muon RPCs according to survey measurements and info from Alessandra.
Correct rotation of Goliath (it was upside down).
Other detectors did not move.


ThomasRuf:
changes in media.geo are not based on latest committed version. I will rewind after merging.

PR #159, opened 2018-10-09
Adding nudet magnet field map

antonioiuliano2:
Inserted nudet magnet field map to FairShip files folder. Then added to the Main Spectrometer field in python/geomGeant4.py to create the global field.
Commented all the previously used uniform fields in charmdet/EmulsionMagnet.cxx

Best regards,
Antonio


ThomasRuf:
I am only wondering if instead of commenting the lines for a constant field, it would be more useful to add an option nutauWithConstField=False, and protect the lines with an If statement. In case, one wants to go back or compare.

olantwin:
I would agree. Either make it optional, or remove the old lines. Commenting the lines serves no purpose, as one can always go back to previous versions of the file thanks to git.

antonioiuliano2:
Thank you very much for your advices. I have added the option for the previous constant fields

ThomasRuf:
fixed it, will commit soon:
      withConstField = False
      if hasattr(shipGeo.EmuMagnet,'WithConstField'): withConstField = shipGeo.EmuMagnet.WithConstField
      if not withConstField:

PR #161, opened 2018-10-15
Use existing environment variables

olantwin:
... to make the script more readable and avoid repetition.

Still don't fully understand why `${SHIPBUILD}/sw/SOURCES/FairRoot/May30-ship/May30-ship/base` is needed.

I hope we can move these into the aliBuild recipes eventually.

### DIFF
--- a/localBuild.sh
+++ b/localBuild.sh
@@ -91,7 +91,7 @@ cp "$SOURCEDIR"/config.sh  config.sh
   echo "export LD_LIBRARY_PATH=\${LD_LIBRARY_PATH}:\${SHIPBUILD}/$architecture"
   echo "export LD_LIBRARY_PATH=$INSTALLROOT/lib:\${LD_LIBRARY_PATH}"
   # ugly fix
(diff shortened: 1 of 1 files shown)


ThomasRuf:
There are many things which I don't understand with setting up the environment on lxplus. I am happy for any suggestion how to improve it. Any modification has to pass the following tests on lxplus and lxplus7:

./localBuild.sh succeeds without errors
source FairShipRun/config.sh sets up the environment
the following three scripts run without any errors:
python $FAIRSHIP/macro/run_simScript.py
python $FAIRSHIP/macro/ShipReco.py -f … -g geofile …
python $FAIRSHIP/macro/ShipAna.py -f … -g geofile …

olantwin:
I completely agree.

This particular change points to the exact same directories (only difference: your version points to `latest`, while using `PACKAGE_ROOT` points to the actual version loaded, which in most cases will be symlinked to by `latest`).

The `${SHIPBUILD}/sw/SOURCES/FairRoot/May30-ship/May30-ship/base` part is untouched.

PR #163, opened 2018-10-17
[WiP] Updates to the online code for the muon flux measurement

olantwin:
- all hits now include information about the original channel and TDC (Hit::GetTDC()), as well as the flags set by the DT DAQ, and can be interrogated to determine whether a hit is leading or trailing edge (Hit::isLeading() and Hit::isTrailing())

- All beam counter (=S1), scintillator (=S2) hits, master triggers (MT), and triggers are now available in the Digi_BeamCounters, Digi_Scintillators, Digi_MasterTrigger and Digi_Triggers branches respectively, and their signal width can be determined using the two edges of the signals. The times are converted to ns, but not shifted in time.

- The flags (included in each hit), contain information on possible TDC errors. There is a method for all hits to check whether the TDC they are on reported problems (Hit::TDCGood()). However, the hits might still be usable even if the TDC has problems. Maybe Daniel knows in which cases we can still use hits on TDCs with problems (e.g. if we have the corresponding triggers?).

- MufluxSpectrometerHits now have a method (Hit::hasDelay()) to check whether their delay is properly calculated or guessed (in the case where MT or the trigger on TDC4 is missing)

- Late MufluxSpectrometerHits are saved to a separate branch (Digi_LateMufluxSpectrometerHits) and set to invalid (The method Hit::isValid() mirrors functionality for MC data).


ThomasRuf:
These are a lot of changes, many not only related to the online code but to the "offline" too. Is it expected that the changes are transparent for current users of the offline objects?
For example, when looping over DT hits, I now have to ask if a hit is leading or trailing edge? I have many of such loops.

olantwin:
I will check how this affects other offline code, so please hold off merging until then (in particular the digitisation code might need updating).

Separating the changes to online and offline code here completely is not possible, as I need a destination for a lot of the information, but I'm open to suggestions on how to improve the format.

Can you think of a good alternative for the leading vs. trailing edge checking?

I think a check for `::isValid()` needs to be made in MC data anyway, as it's used to suppress multiple hits per tube there.

ThomasRuf:
One more question, with the changes applied, would it still be possible to analyze existing data files, or does it require new converted data?

ThomasRuf:
For leading and trailing edge, I propose the following:
We add an additional data member to the hit class, which contains the TDC value of the trailing edge, and add a get method to retrieve the trailing edge.

Like this, it is backward compatible, by default one gets the TDC value of the leading edge. In addition, one gets for free the value of the corresponding trailing edge. No checking needed, and no loop over hits to find hits with same detID.

olantwin:
> One more question, with the changes applied, would it still be possible to analyze existing data files, or does it require new converted data?

If I understand ROOT I/O correctly, ROOT should be able to tell apart the classes by comparing their hashes and schema versions, which I incremented.

olantwin:
> For leading and trailing edge, I propose the following:
> We add an additional data member to the hit class, which contains the TDC value of the trailing edge, and add a get method to retrieve the trailing edge.

I think then it might be better to just add the measured width (i.e. `trailing-leading`) to the hit classes, as that's the quantity we usually care about (and the raw trailing edge time can be easily reconstructed from `leading+width`).
 
> Like this, it is backward compatible, by default one gets the TDC value of the leading edge. In addition, one gets for free the value of the corresponding trailing edge. No checking needed, and no loop over hits to find hits with same detID.

Essentially this shifts the complexity into the conversion, but I already have some ideas how to do it efficiently.

ThomasRuf:
Yes, but the conversion is done once, and looping over hits is done thousand times. I don't know how the raw data is sorted, but if leading and trailing edge are randomly distributed, I would try an algorithm which creates a hit object the first time the detector ID is seen, together with a list which keeps track of the pointer to the hit as function of detID, and next time rawdata with same detID is seen, the hit is updated.

olantwin:
Thanks, but I don't think that's practical for the drift tubes, as I need to find e.g. triggers for each TDC etc. before I can create a hit object.

But as we know the relationship between channel IDs and I have to sort them anyway to take only the first hit per channel, I have an implementation now, that I'm testing.

olantwin:
Thanks, but I don't think that's practical for the drift tubes, as I need to find e.g. triggers for each TDC etc. before I can create a hit object.

But as we know the relationship between channel IDs and I have to sort them anyway to take only the first hit per channel, I have an implementation now, that I'm testing (see latest commit).
I'll expand the implementation to triggers as well.

olantwin:
Thanks, but I don't think that's practical for the drift tubes, as I need to find e.g. triggers for each TDC etc. before I can create a hit object.

But as we know the relationship between channel IDs and I have to sort them anyway to take only the first hit per channel, I have an implementation now, that I'm testing (see latest commit).
I'll expand the implementation to triggers as well.

olantwin:
Thanks, but I don't think that's practical for the drift tubes, as I need to find e.g. triggers for each TDC etc. before I can create a hit object.

But as we know the relationship between channel IDs and I have to sort them anyway to take only the first hit per channel, I have an implementation now, that I'm testing (see latest commit).
I'll expand the implementation to triggers as well.

olantwin:
@owtscharenko : you should now be able to work relative to the master branch of FairShip.

owtscharenko:
Very nice. The question is how to get the geometry from Antonios fork. Maybe rebase on master now?

olantwin:
Your personal changes I would interactively rebase on the current master (just remove all commits I added in online, as they were squashed into a single commit in master) or on @antonioiuliano2's branch once he has rebased on the current master.

PR #171, opened 2018-11-06
Fixing a bug in neutrino field map

antonioiuliano2:
Checking the implementation of the new magnetic field map in the nutaudet Emulsion Magnet, I have noticed that the rotations have been implemented wrongly. To avoid further errors, I have recreated the map in FairShip reference, in order to avoid the need for rotations entirely.

I will continue checking.

Sorry for the mistake and best regards,
Antonio


antonioiuliano2:
I have also set as default the previous configuration with constant field, as a safety measurement until we are done testing the map.

Please let me know if you agree.
Best regards,
Antonio

PR #172, opened 2018-11-08
New PatRec WITHOUT dist2wire

hushchyn-mikhail:
This PR was used to the report in the attachment.
[reportship24oct2018.pdf.pdf](https://github.com/ShipSoft/FairShip/files/2563745/reportship24oct2018.pdf.pdf)

### DIFF
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -12,7 +12,7 @@ dy  = None
 saveDisk  = False # remove input file
 pidProton = False # if true, take truth, if False fake with pion mass
 realPR = ''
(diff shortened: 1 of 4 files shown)


ThomasRuf:
there is a bit of inflation of names
"Prev", "FH", "AR", "Baseline", "TemplateMatching", "TemplateMatchingShort", "TCV0", "TCV1", "TCV2" 
a normal user will be overcharged of all these options. Could a short description for each option be added?

Second:  in  python/shipDigiReco.py, almost every line is changed according to git. It is impossible to see what are the real changes. I suspect there are problems with tabs, file format, ... Could you please fix this?

PR #173, opened 2018-11-09
New PatRec without dist2wire

hushchyn-mikhail:


hushchyn-mikhail:
Dear Thomas,

The pull request is recreated. Previously there was a problem with indentations in files. 
By default, I am using 4 spaces for one indent. However, in FairShip files I saw several different indents: 1, 2 and 3 spaces. Probably, we have to choose one of them. 

I also added very short description of realPR options.

ThomasRuf:
macro/ShipReco.py : OK

python/shipDigiReco.py, findTracks: Still a lot of changes, probably because changing style, two columns instead of one? In any case, makes it difficult to see the real changes. 
why is ut.writeHists(h,"recohists.root")  been removed?

python/shipStrawTracking.py, def extrapolateToPlane(fT,z): why is it needed to make a local copy of exisiting code?

hushchyn-mikhail:
python/shipDigiReco.py, findTracks was changed a lot. It was easier to replace the whole function, that went line by line saving different indentations. As I wrote earlier, the file has no one indent, but python editor needs one. I propose to select one indent, and apply it to the whole file.

ut.writeHists(h,"recohists.root") was removed, because all PR quality metrics are calculated in python/shipStrawTracking.py. If ut.writeHists(h,"recohists.root") is needed, I can return it. 

python/shipStrawTracking.py, def extrapolateToPlane(fT,z): I made a copy, because import TrackExtrapolateTool gave me an error in line 6. I was not sure how to fix it, so I just made my own copy with fixes.

ThomasRuf:
for extrapolateToPlane: it would be better to fix the error, instead of copying code and apply local fix. Could you post the error which you encountered? 

For ut.writeHists(h,"recohists.root"), probably ok to remove it for the moment, if only the PR algorithm produces histograms. As soon as other algorithms produce monitor histograms, we might need a central manager for this. 

findTracks():  the comment about indents, I don't understand. I would like to understand if the logic has changed for using the fake PR.

hushchyn-mikhail:
Recreate the pull-request.

PR #174, opened 2018-11-15
Docker howto

hushchyn-mikhail:


Dockerfile
```diff
@@ -1,6 +1,16 @@
 FROM olantwin/ship-base:20180627-online
 
-ADD . /FairShip
+# Copy FairShip scripts
+COPY . /FairShip
```
olantwin: Why do you replace the `ADD` here? What are the advantages of `COPY`?

Dockerfile
```diff
@@ -1,6 +1,16 @@
...
 
-RUN aliBuild -c shipdist/ --defaults fairship build FairShip --no-local ROOT \
-	&& aliBuild clean --aggressive-cleanup
```
olantwin: Why remove the cleanup? I prefer keeping the image size as small as possible.

Dockerfile
```diff
@@ -1,6 +1,16 @@
...
+# Build FairShip
+RUN aliBuild -c shipdist/ --defaults fairship build FairShip --no-local ROOT
+
+# Additional library for OpenGL
+RUN yum install -y mesa-dri-drivers
+
+# Fix problems with fonts
+RUN yum install -y dejavu-lgc-sans-fonts
```
olantwin: I think adding the packages should happen in the base image to minimise build times: <https://github.com/olantwin/ship-base>

Dockerfile
hushchyn-mikhail: This is nothing more than a habit, and I found it in tips https://nickjanetakis.com/blog/docker-tip-2-the-difference-between-copy-and-add-in-a-dockerile.

Dockerfile
hushchyn-mikhail: Agree. 

Ship-base Dockerfile is great. Can you add this link into the FairShip Dockerfile or ship-base on dockerhub? It is very useful.

Dockerfile
hushchyn-mikhail: Agree. I'll put it back. I removed it for tests, but now we can put it back.

hushchyn-mikhail:
About the ```ENTRYPOINT```. Yes, we can warn users about it. But now I can not image how can it lead to a problem.

olantwin:
> About the `ENTRYPOINT`. Yes, we can warn users about it. But now I can not image how can it lead to a problem.

There might be some cases where users already call `alienv` in their scripts, which could cause trouble (but should not fail silently in that case, so should be easy to fix).

Dockerfile
olantwin: I hope we can also move it to ShipSoft eventually (@FonsRademakers ?)

I'll create a pull-request to your repo adding the info. (and should probably add some more info on Dockerhub as well)

Dockerfile
olantwin: This is a good point. This is indeed a good change!

Dockerfile
olantwin: Thanks!

Dockerfile
olantwin: Also a good time to maybe rebuild ship-base...

PR #178, opened 2018-11-20
Muon flux drift tubes PR

hushchyn-mikhail:
- More tracks recognized
- Hits sharing
- Multiple combinations of tracks before and after the magnet
- Debug option

Tested on SPILLDATA_8000_0519173180_20180723_075716.root file.

### DIFF
--- a/python/MufluxPatRec.py
+++ b/python/MufluxPatRec.py
@@ -11,7 +11,7 @@ def initialize(fgeo):
     pass
 
 
-def execute(SmearedHits, TaggerHits, withNTaggerHits, withDist2Wire):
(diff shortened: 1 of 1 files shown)


ThomasRuf:
It would be good to have the option for only sharing hits before the magnet. After the magnet, there is no use case, as far as I know. With the current setting, we get a lot of ghosts.
image

ThomasRuf:
MufluxPatRec.py:332: RuntimeWarning: divide by zero encountered in double_scalars
  momentums.append(1.03 / (alpha_12 - alpha_34))

need a protection against such cases. 

Does MufluxPatRec work for zero field data?

PR #183, opened 2018-12-04
Combination of muon tagger tracks with drifttubes tracks

hushchyn-mikhail:
Starting from a muon tagger track try to combine it with drifttubes tracks.


hushchyn-mikhail:
Drifttubes tracks are recognized tracks returned by MufluxPatRec.execute()

ThomasRuf:
although the naming drifttubes tracks is ambiguous, could be confused with fitted tracks.

PR #188, opened 2018-12-05
Remove unmaintained (and almost certainly broken) vagrant files

olantwin:
These files serve a similar purpose to what can be done with the Dockerfiles, but have never been updated to the build procedure using aliBuild (they were last modified 4 years ago). This makes them almost certainly broken, and as nobody has ever complained, unused.

I suggest we remove them, unless there's a volunteer to fix and maintain them (as usual, if anyone wants to look at them, they are easily accessible by going back to earlier commits, so no information is lost by removing them).


ThomasRuf:
this is in principle fine with me, only comment I have about "easily accessible by going back to earlier commits", this I guess requires the knowledge that these files ever existed. Or is there a way to see all files which ever had been in the git repository?

olantwin:
If you do this, it will give you a list of commits including a file (also accepts wildcards etc., so you don't need the exact path):

```
git log --all --full-history -- <path-to-file>
```

I am not sure whether GitHub has any special interface, but since this pull request exists, anyone searching the repository for "vagrant" or something similar will see this pull request and commit, and can revert the change, if they need to.

In any case, the FairShip versions that Vagrant might work with (e.g. before the switch to aliBuild etc.), will include the Vagrantfile, so if you check out one of those commits, the files will be available.

olantwin:
We could also add a page to the Wiki with removed features/scripts/things letting people know about things that could potentially be useful for reference at some point, but were removed for one reason or another.

In this case however, if one wanted to restore support for Vagrant one would probably need to start from scratch anyway, as too much has changed.

PR #191, opened 2019-01-23
Blacklist disconnected channels

olantwin:
This blacklists all channels that are not connected, so that they can't accidentally register as valid hits on valid detector IDs.

### DIFF
--- a/online/ShipOnlineDataFormat.h
+++ b/online/ShipOnlineDataFormat.h
@@ -44,6 +44,7 @@ struct ChannelId {
       bool scintillatorA = false;
       bool scintillatorB = false;
       bool master_trigger = false;
+      bool blacklisted = false;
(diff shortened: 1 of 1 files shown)


ThomasRuf:
this will be in action for new converted data?

olantwin:
I will hold off on re-converting anything until I (or someone else?) figures out why some detector IDs are duplicate (so that only one re-conversion needs to happen. I am re-checking the algorithm, but there's nothing suspicious or too-clever that could explain why more than one hit occurs for some detector IDs.

In the meantime, I will continue converting any still unconverted files with this blacklist, as this blacklist definitely doesn't hurt.

olantwin:
> this will be in action for new converted data?

So yes, any converted data starting today.

PR #193, opened 2019-01-29
Fix source nutaufieldmap

antonioiuliano2:
Dear all,

there was a bug in the source file of the field map. Here is a brief description of the issue:

The txt file had a different reference system with respect to the FairShip one:
x - beam axis (instead of z)
z - height (instead of y)
y - width (instead of x)
Before, I had rotated the coordinates in the standard FairShip reference, but I did not notice that the coordinates must increase in a specific order (z increase first, then y, finally x), in order to the map to work correctly. Therefore, the actual FairShip map had still the axes inverted.

I have fixed this issue and reloaded the root file with the field map, now active by default.

Best regards,
Antonio


olantwin:
Do I understand correctly that this bug did not affect anyone by default, as the field was constant by default?

antonioiuliano2:
Yes, it is as you say.

Since now it should work properly (hoping no new bugs are discovered), I have enabled it by default in this commit

PR #195, opened 2019-01-31
Fix delay issue on lxplus

olantwin:
Rewriting the ternary statement as an if statement fixed the issue for some reason on lxplus. The two statements should have been equivalent (compiler differences maybe?).

I will check other ternary statements to see whether this is just a problem in this case or in general.

### DIFF
--- a/online/DriftTubeUnpack.cxx
+++ b/online/DriftTubeUnpack.cxx
@@ -181,8 +181,14 @@ Bool_t DriftTubeUnpack::DoUnpack(Int_t *data, Int_t size)
       if (detectorId == 0) {
          // Trigger
          trigger++;
-         trigger_times[TDC] =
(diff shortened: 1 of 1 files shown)


olantwin:
Just a quick update: I checked all other ternary expressions, and they give the expected results, so there are no similar bugs in the conversion.

Maybe the interplay of several `std::map` operations + ternary in this case caused the issue with the `lxplus` combination of compiler/STL?

ThomasRuf:
will you restart the conversion to ROOT? Remove the existing files first.

olantwin:
I will not modify the files right now. I will have another look whether I can track down the duplicate "first" hits and then restart the conversion, writing to my local eos space first.

Once everything looks alright I will warn everyone and replace the current converted files.

olantwin:
I'll probably restart the conversion tomorrow, so all files should be re-converted by Monday.

PR #196, opened 2019-02-11
Fix bug with muontagger endpoints

evanherwijnen:


ThomasRuf:
Don't understand. For horizontal strips, we need left and right end points, y left == y right, x left = - x right. For vertical strips, we need, x top == x bottom, y top = - y bottom. Or?

evanherwijnen:
My fix takes also the width into account. The printout from drifttubeMonitoring/debugRPCstrips() is now correct.

ThomasRuf:
I think the strips are not anymore vertical or horizontal, but inclined by width/length.

evanherwijnen:
True. Can you please remove this pull request, I'll fix it.

PR #197, opened 2019-02-11
New muontagger fix

evanherwijnen:
Fixed MuonTagger and MuonTaggerHit endpoints bug.


ThomasRuf:
I still have a problem with this. A top(left) and bottom(right) point for me is changing one coordinate, keeping two coordinates constant.  With Double_t bot[3] = {-S->GetDX(),-S->GetDY(),-S->GetDZ()}; all three coordinates are changed.

evanherwijnen:
Yes but depending on the orientation the routine returns the average of top&bot, thus the coordinate is not changed. However I just noticed there is a typo that I have to fix, so please delete this pull request also.

PR #200, opened 2019-02-13
Pixel Detectors taken out of the Spectrometer

matclim:


ThomasRuf:
May I ask what checks had been performed to verify that he changes don't brake the running code?
Does run_simScript and MufluxFixedTarget work?

matclim:
I ran run_simScript (python run_simScript.py --charm 1 -A charmonly --CharmdetSetup 1 -n X) and MufluxReco (python MufluxReco.py -f ship.conical.Pythia8CharmOnly-TGeant4.root -g geofile_full.conical.Pythia8CharmOnly-TGeant4.root) and both work. I haven't run MufluxTarget since I don't think it is relevant to the charm measurement ?

matclim:
After closer examination, the MufluxfixedTarget doesn't work with those changes. I'll edit, check and do another request

evanherwijnen:
Your code in charmDet_conf.py is incompatible with the muon flux setup.
You have to make sure all pixel related code is not active when muflux==True.
The following code fixes the problem:  
if (ship_geo.MufluxSpectrometer.muflux==False): 
    PixelModules = ROOT.PixelModules("PixelModules",ship_geo.PixelModules.DX, ship_geo.PixelModules.DY, ship_geo.PixelModules.DZ,ROOT.kTRUE)
    PixelModules.SetSiliconDZ(ship_geo.PixelModules.DimZSi)
    detectorList.append(Box)
    detectorList.append(PixelModules)
    for i, (x, y, z) in enumerate(zip(ship_geo.PixelModules.xSi,ship_geo.PixelModules.ySi,ship_geo.PixelModules.zSi)): 
      PixelModules.SetSiliconStationPositions(i, x, y, z)

olantwin:
> After closer examination, the MufluxfixedTarget doesn't work with those changes. I'll edit, check and do another request

You don't need to create a new pull-request. You can just keep pushing commits to this branch, and the pull-request will be updated.

ThomasRuf:
could you please remove files which are not for general interest:

.gitignore  with macro/mateiresults


I suspect that charmdet/Spectrometer.h needs an increase in class number, since you changed header file, otherwise, problems with backward compatibility:
ClassDef(Spectrometer,1) -> ClassDef(Spectrometer,2)

olantwin:
I would also prefer if you could delete the now obsolete code in `Spectrometer.cxx/h` instead of commenting it out.  Having nearly identical code, once commented and once active in two different files could lead to confusion later.

matclim:
I believe everything is fixed

PR #203, opened 2019-02-14
Muontagger ep

evanherwijnen:
Hopefully the muontagger endpoints are now ok.


ThomasRuf:
if it is on purpose that  vtop (x or y) is set to Gbot, then it is probably ok.

evanherwijnen:
yes this is how it should be.

PR #205, opened 2019-02-20
Pattern Recognition without dist2wire

hushchyn-mikhail:


ThomasRuf:
these two change have serious implications:
    # if nM < 25 : continue                          # not enough hits to make a good trackfit 
    # if len(stationCrossed[atrack]) < 3 : continue  # not enough stations crossed to make a good trackfit 

trackfit is done for any list of measurements. This will probably increase the number of track failures, increase the number of bad tracks in the output file. For physics analysis, I don't see the use of taking track candidates which only cross two stations, or have not enough measurements. Would require to check this during analysis, otherwise weird momentum, weird inv mass, ...
Could you please outline the use cases for these changes?

hushchyn-mikhail:
nM < 25:
We found that track nmeas cuts are present in ShipAna.py. And these cuts are adjustable. During the SST geometry optimization we found that nmeas < 20 is better from analysis point of view. So it reasonable to remove this condition from the track fit step and keep it in analysis step. This gives possibility to select nmeas cuts during analysis without rerunning track fit.

if len(stationCrossed[atrack]) < 3:
Here I do not have strong reasons, except that this condition can be used in steps after track fit, if needed, without rerunning track fit.

ThomasRuf:
you need to distinguish between expert and normal user. A normal user will just take all fitted tracks and not necessarily start to apply trivial quality cuts.
changing nM to 20 is probably ok, but I would like to see the mass resolution and vertex resolution for HNLs made with a least one track of less than 25 measurements.

hushchyn-mikhail:
Ok, then I propose to return the conditions (if nM < 25, if len(stationCrossed[atrack]) < 3). Then change them after analysis study, if needed, with separate pull-request. Is it ok?

ThomasRuf:
ok

Von: hushchyn-mikhail <notifications@github.com>
Gesendet: Dienstag, 5. März 2019 16:56
Betreff: Re: [ShipSoft/FairShip] Pattern Recognition without dist2wire (#205)


Ok, then I propose to return the conditions (if nM < 25, if len(stationCrossed[atrack]) < 3). Then change them after analysis study, if needed, with separate pull-request. Is it ok?

hushchyn-mikhail:
Done

ThomasRuf:
There is one more issue:
 # pdg    = self.sTree.MCTrack[atrack].GetPdgCode()
   pdg = 13

you hardcode the type of each track to be a mu-.  The problem with this, the energy loss of electrons / muons / pions is different. When the trackfit returns the momentum at the first measurement, the energy loss is corrected for. With pdg = 13, this will give the wrong answer for electrons. Did you checked by how much the momentum resolution, HNL mass resolution, VX resolution, changes for the channels with electrons due to this change? 
Ideally, the particle type assignment should come eventually from the particle ID algorithm, but this still needs to be implemented proper.

hushchyn-mikhail:
I hardcoded this because we can not use MC information during track fit. That was the only reason.
I'll return it.

hushchyn-mikhail:
Done

ThomasRuf:
What you say is in principle true. But it requires somebody to implement the proper steps to derive the particle type from the data. Until this is not implemented, we have to use the MC shortcut.

ThomasRuf:
please remove
self.sTree.GetCurrentFile().Close() 
this is now handled by pyExit()

hushchyn-mikhail:
Done

PR #207, opened 2019-03-04
Minor change to script charmdet/runReconstruction.py

StBies:
# Purpose
So far, the above mentioned script counts python processes for a hardcoded username. When this script will regularly be called by different users, it is more convenient, if the script reads the username itself.

# Included commit messages
Read below for the included commit messages:
Changed script charmdet/runReconstruction to use username of script caller instead of hardcoded one.

Changed script charmdet/runReconstruction to ensure an integer value of CPU threads will be requested.


StBies:
By the way, this only works on UNIX like systems

PR #209, opened 2019-03-06
Dark Photon production: fix in pbrem and qcd production, addition of rate calculation tools

amagnan:
- addition of a python file to calculate overall rates in different DP production modes
- fix in Pythia parameters, to allow proper decay of the DP for masses below 0.7 GeV in pbrem mode. The DP is now declared as non-resonant (there is a hardcoded limit in mass for resonant particles at 0.4 GeV -> OK for QCD, for which the DP has to be a resonance, and which is generated anyway for DP masses above 1.4 GeV.
- For QCD, the decay channels were declared with meMode = 102, but further investigations seem to indicate strange behaviour in the choice of decay channels (substantially more decays to ee,mumu compared to what is hardcoded in the BR given as input), closer to the input values for meMode=103.

### DIFF
--- a/python/dpProductionRates.py
+++ b/python/dpProductionRates.py
@@ -1,0 +1,98 @@
+import ROOT,os,sys,getopt,math
+import shipunit as u
+import proton_bremsstrahlung
+
+
+PDG = ROOT.TDatabasePDG.Instance()
+protonFlux = 2e20
+
+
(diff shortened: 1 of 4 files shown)


JLTastet:
`dpProductionRates.py` does not seem to be used anywhere. Is this intended?

PR #213, opened 2019-03-18
Include drift tube geometry for charm based on survey 

dbick:


evanherwijnen:
I checked that these changes do not conflict with the muonflux setup.

PR #215, opened 2019-03-24
Fix of compatibility issue of CharmTarget option

antonioiuliano2:
Dear all,

in the recent CharmTarget commit I have done the error of changing a name of a method, thus ruining backward compatibility with previous simulations. 

I am truly sorry for the inconvenience, and this patch should fix it. Also, since Box class is not used in the MuonFlux configuration, I propose to call its methods only for the Charmdet configuration. This should avoid future conflicts between the two configurations.

Best regards,
Antonio

### DIFF
--- a/charmdet/Box.cxx
+++ b/charmdet/Box.cxx
@@ -164,9 +164,9 @@ void Box::SetTargetDesign(Bool_t Julytarget){
   fJulytarget = Julytarget;
 }
 
-void Box::SetTargetNumber(Int_t CharmTargetNumber){		
-  nrun = CharmTargetNumber;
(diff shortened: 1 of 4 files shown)


ThomasRuf:
I tried, however, new issue coming up:
MufluxSpectrometer.SetT3(ship_geo.MufluxSpectrometer.SurveyCharm_T3ax,ship_geo.MufluxSpectrometer.SurveyCharm_T3ay,ship_geo.MufluxSpectrometer.SurveyCharm_T3az,0)
AttributeError: 'AttrDict' object has no attribute 'SurveyCharm_T3ax'

I think we definetely need a clear separation between charm and muflux setup. Don't understand why I need SurveyCharm for muflux.

PR #217, opened 2019-03-25
Adopt to the change of time units convention

DaniilSu:
Not completely sure, if t0 is affected. Is it also in sec from now?

### DIFF
--- a/strawtubes/strawtubesHit.cxx
+++ b/strawtubes/strawtubesHit.cxx
@@ -40,7 +40,7 @@ strawtubesHit::strawtubesHit(strawtubesPoint* p, Double_t t0)
      Double_t sigma_spatial = module->StrawSigmaSpatial();
      module->StrawEndPoints(fDetectorID,start,stop);
      Double_t t_drift = (fabs( gRandom->Gaus( p->dist2Wire(), sigma_spatial ) )/v_drift)*1.E+09; // in ns
-     fdigi = t0*1.E+09 + p->GetTime() + t_drift + ( stop[0]-p->GetX() )/ speedOfLight; // in ns
+     fdigi = t0 + p->GetTime() + t_drift + ( stop[0]-p->GetX() )/ speedOfLight; // in ns
      flag = true;
 }
 void strawtubesHit::StrawEndPoints(TVector3 &vbot, TVector3 &vtop)
--- a/strawtubes/strawtubesHit.cxx
+++ b/strawtubes/strawtubesHit.cxx
@@ -39,8 +39,8 @@ strawtubesHit::strawtubesHit(strawtubesPoint* p, Double_t t0)
      Double_t v_drift       = module->StrawVdrift();
      Double_t sigma_spatial = module->StrawSigmaSpatial();
      module->StrawEndPoints(fDetectorID,start,stop);
-     Double_t t_drift = fabs( gRandom->Gaus( p->dist2Wire(), sigma_spatial ) )/v_drift;
-     fdigi = t0 + p->GetTime() + t_drift + ( stop[0]-p->GetX() )/ speedOfLight;
+     Double_t t_drift = (fabs( gRandom->Gaus( p->dist2Wire(), sigma_spatial ) )/v_drift)*1.E+09; // in ns
+     fdigi = t0*1.E+09 + (p->GetTime())*1.E+09 + t_drift + ( stop[0]-p->GetX() )/ speedOfLight; // in ns
      flag = true;
 }
 void strawtubesHit::StrawEndPoints(TVector3 &vbot, TVector3 &vtop)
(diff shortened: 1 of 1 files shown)


ThomasRuf:
I have not touched, t0 should still be in ns

DaniilSu:
Also in strawtubes.cxx I saw that the time of the Hit is manually corrected to ns (line 108):
fTime   = gMC->TrackTime() * 1.0e09;
Should I change it to sec? Does it mean that in the output file we get ns instead of sec for the time of the hit?

ThomasRuf:
We might need to distinguish between time of flight for mctracks, points, and digi, tdc. First in natural fairroot units, others in ship ns?

DaniilSu:
Well, according to FairRoot class reference:

NewDetectorPoint::NewDetectorPoint	(	Int_t 	trackID,
Int_t 	detID,
TVector3 	pos,
TVector3 	mom,
Double_t 	tof,
Double_t 	length,
Double_t 	eLoss 
)		

-- | --
trackID | Index of MCTrack
detID | Detector ID
pos | Ccoordinates at entrance to active volume [cm]
mom | Momentum of track at entrance [GeV]
tof | Time since event start [ns]
length | Track length since creation [cm]
eLoss | Energy deposit [GeV]

But for MCtracks the time units are different:

virtual void FairGenericStack::PushTrack	(	Int_t 	toBeDone,
Int_t 	parentID,
Int_t 	pdgCode,
Double_t 	px,
Double_t 	py,
Double_t 	pz,
Double_t 	e,
Double_t 	vx,
Double_t 	vy,
Double_t 	vz,
Double_t 	time,
Double_t 	polx,
Double_t 	poly,
Double_t 	polz,
TMCProcess 	proc,
Int_t & 	ntr,
Double_t 	weight,
Int_t 	is,
Int_t 	secondparentID 
)		

Virtual method PushTrack. Add a TParticle to the stack. This function has an extra argument wrt to the function defined in the base class.
-- | --
toBeDone | Flag for tracking
parentID | Index of mother particle
pdgCode | Particle type (PDG encoding)
px,py,pz | Momentum components at start vertex [GeV]
e | Total energy at start vertex [GeV]
vx,vy,vz | Coordinates of start vertex [cm]
time | Start time of track [s]
polx,poly,polz | Polarisation vector
proc | Production mechanism (VMC encoding)
ntr | Track number (filled by the stack)
weight | Particle weight
is | Generation status code (whatever that means)
secondparentID | used fot the index of mother of primery in the list

So, the sec only appear in the methods related to MCTracks, the ones related to the description of the detectors use ns for time.

ThomasRuf:
How do we get this sorted out? we have the standard workflow Pythia->Geant4->MCPoints and MCTracks. But we also have MCPoints->MCTracks->Geant4->MCpoints and MCTracks for the muon background.

olantwin:
Maybe the FairRoot team has some suggestions?

Even if we have a usecase they did not originally intend, it's probably good to let them know that we use it this way, so they don't end up accidentally breaking it.

ThomasRuf:
One step back. In HNLPythia8Generator, we use:

convert pythia's t[mm/c] to [s]
const Double_t cm = 10.; // pythia units are mm
const Double_t c_light = 2.99792458e+10; // speed of light in cm/sec
cpg->AddTrack(..,tS/cm/c_light,..).

I think, for some strange reason, this gave always consistent results, time of MCTrack in ns and and time of hit in ns. So, somewhere, there must be another conversion from s to ns?

In FixedTargetGenerator, no conversion was done, time from Pythia8 directly copied to FairRoot. This is now corrected, but of course only for newly generated data. However, only affects MCTracks decaying in Pythia8 (mainly pi0).
tof = fPythia->event[ii].tProd() / (10*c_light) ; // to go from mm to s
cpg->AddTrack(...,tof,...)

In MuonBackGenerator, we read previously simulated data, take time from MCTrack   
tof =  track->GetStartT();
cpg->AddTrack(...,tof,...)

According to the above, this must be wrong, since time is given in ns, but seconds are expected.

=======

change of shipunits affects:
geometry/geometry_config.py:    c.strawtubes.v_drift = 1./(30*u.ns/u.mm) # for baseline NA62 5mm radius straws)
geometry/charm-geometry_config.py:    c.MufluxSpectrometer.v_drift = 1./(72*u.ns/u.mm) # 1300 ns max for 36.3 mm drifttubes
python/shipDigiReco.py:  if n>0: t0 = t0/n - 73.2*u.ns


Wondering, if we should revert the change in shipunits for time and go back to 
u.ns = 1 and  u.s = 1E9?

ThomasRuf:
It seems what is now committed makes no sense. Just checked a newly generated file. A MCTrack with ~100GeV travelling 9.9cm gets a ToF of 330357344.0 . Would expect 9.9cm/c = 3.303e-10seconds or 0.3303ns. Don't know where the mistake happens.

ThomasRuf:
I found where this other conversion from time is done:

a) in the detector classes, for example veto.cxx
 fTime   = gMC->TrackTime() * 1.0e09;

b) in shipdate/ShipMCTrack.cxx:
ShipMCTrack::ShipMCTrack(TParticle* part)
  : TObject(),
...
    fStartT(part->T()*1e09),
which means, natural time unit for FairShip is ns. 
So, I propose to revert the change of shipunits, stay with 1ns = 1, 1sec = 1E9. 
And correct the missing conversion when reading FairShip events in muonBackGenerator.

ThomasRuf:
I close the pull request. We stay with 1ns for FairShip objects. So, no changes required except when converting from Pythia8/FairShip to Geant4.

PR #222, opened 2019-04-30
All Pixel Modules Sliced in preparation for Digitization

matclim:
In order to digitize the PixelModules, several energy depositions are required in the Z direction for each module. To perform this, each module was sliced in 10 since no way of changing the GEANT4 step was found. Some general code cosmetics were also changed.


matclim:
I noticed an issue with modulé 0 position, will reopen the request once it is fixed

PR #225, opened 2019-05-09
corrected constants and coupling used in Width_N_L fucntion

alexmarshallbristol:


ThomasRuf:
Is this commit confirmed by the physics group? I haven't received any green light from Nico Serra.

PR #226, opened 2019-05-09
Millepede library implementation and wrapper class for charmdet

StBies:
# Contents
This contains:

- The code for Millepede as a seperate library for FairShip including CMake script

- A wrapper class to call Mille from within python in the charmdet library

- Documentation in doxygen format


ThomasRuf:
Does it compile on lxplus?

StBies:
> Does it compile on lxplus?

Yes, just checked using the way described on the main project's readme with localBuild.sh, see attached screenshot.

screenshot

PR #229, opened 2019-05-27
Adding information about lxplus installation

antonioiuliano2:
Hi to all you FairShip developers,
today I have been contacted by a colleague who received an error, after trying to launch the full installation on lxplus. Therefore, I consider it better to clearly explain in the README that the full installation should not be done there.
Also, I have suggested the usage of lxplus6, since I noticed the display not working in lxplus7.

Best regards,
Antonio

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Additionally for developers:
     ```bash
     export SHIPBUILD=/cvmfs/ship.cern.ch/SHiPBuild
     ```    
-
+   on lxplus you should not follow steps 3 and 4, since the software is already installed in the previous folder.
+   The usage of lxplus6 is currently recommended.
 1. Install FairShip
     ```bash
     git clone https://github.com/ShipSoft/FairShip.git
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ Additionally for developers:
     ```bash
     export SHIPBUILD=/cvmfs/ship.cern.ch/SHiPBuild
     ```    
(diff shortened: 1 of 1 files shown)


olantwin:
I would maybe rephrase the recommendation for lxplus6:

> There is currently a known issue with the event display on CC7. As a workaround, lxplus6 can be used until this issue is resolved.

Do we currenlty know why it doesn't work?

In general we should encourage use of lxplus 7, as lxplus 6 won't be supported forever, and we'll only be able to discover and fix these issues if people use CC7.

ThomasRuf:
on cc7, the system libraries changed since building. Currently SHiPBuild points to
SHiPBuild -> /cvmfs/ship.cern.ch/July2018/SHiPBuild
did anybody tried with /cvmfs/ship.cern.ch/Nov2018/SHiPBuild? 

The full installation on lxplus fails because a normal user will not have enough memory for the compilation. However, I thought this issue is being mitigated when we changed to CMAKE_BUILD_TYPE: "RELEASE".

antonioiuliano2:
Hi,

I have tried, unfortunately I continue to get the same error:

libpng warning: Application built with libpng-1.2.55 but running with 1.5.13
TASImage::GetMask:0: RuntimeWarning: No image

Best regards,
Antonio

Il giorno lun 27 mag 2019 alle ore 11:38 Thomas Ruf <
notifications@github.com> ha scritto:

PR #232, opened 2019-05-29
Use CMake from SHiPBuild

olantwin:
On lxplus, localBuild doesn't load CMake from SHiPBuild, as it's only a build dependency of FairShip, but not a dependency.

This loads CMake explicitly. Ideally we'd load all *build* dependencies automatically as well.

@antonioiuliano2

### DIFF
--- a/localBuild.sh
+++ b/localBuild.sh
@@ -20,7 +20,7 @@ fi
 echo "Setting environment for ${architecture} for $SHIPBUILD"
 echo "SHIPBUILD=$SHIPBUILD" > config.sh
 
-"$SHIPBUILD"/alibuild/alienv -a "${architecture}" -w "$SHIPBUILD"/sw printenv FairShip/latest >> config.sh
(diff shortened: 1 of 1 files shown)


olantwin:
@antonioiuliano2: Did this fix work for you?

I can't test myself, as LHCb pollutes my environment...

antonioiuliano2:
Yes, I can confirm that the CMake fix works for me.

Thank you very much.

Best regards,
Antonio

Il giorno mar 4 giu 2019 alle ore 14:07 Oliver Lantwin <
notifications@github.com> ha scritto:

olantwin:
In that case, Thomas, could you please merge? Otherwise users not on LHCb will experience trouble building FairShip on lxplus.

ThomasRuf:
Oliver, you should have one account for LHCb and another for SHiP.

olantwin:
Initially I had some issues with LHCb software, but now both stacks work fine on the same computing account (as far as I can tell).

But the only relation to this issue is that LHCb caused me to have a newer CMake, so I didn't notice that localBuild.sh didn't add the SHIPBUILD CMake to the path (which only became noticeable once we actually required the correct CMake version with #230).

PR #236, opened 2019-06-21
[LINT] Unused variable (DO NOT MERGE)

siscia:
Hi all,

a standard technique when reviewing code is to look for unused variables and declaration.

### Why? They do no harm, right?

Yes, it is correct that they do no harm, however, are usually symptomatic of other issues.

First of all, if the variable is not used, why do we compute them in first place? It is a waste of resources.

Then they may hide some issue in the code? Maybe we need to actually use them, but for some reason, we forgot to. Maybe we used a variable with a similar name but that contains a wrong value.

Hence it is important to have an eye on unused variables, [very expert programmers wasted hours because they forget to run an unused variable analysis on their code](https://twitter.com/antirez/status/969169622430879744).

### How to find unused variables

Fortunately is possible to automatically find unused variables, indeed this same PR was generated using an automated tool. Underneat it uses a simple graph algorithm.

The PR itself is safe to merge, the behaviour of the software does not change after merging, however is much more important to understand **why** the unused variable is there rather than just remove it.

Again, unused variables cause no harm, but they are sympomatic of other, underlining issues.

### Can you remove all the unused variables?

I could in theory, but I am afraid to don't know the code base and the domain well enough to understand each and every case, somebody more expert should take a look.

### It is a lot of work!

I know, but usually is worthed. If we are interested in this kind of work we could procede step by step, one folder at the time, so that in less than a week everything get's done.

### Even if we do this work, other people will re-introduce unused variables.

The solution is to run the same LINT before to merge each pull request, in this way we are sure that no unused variable is added to the codebase.

We can done this in gitlab for free.

### But some unused variables are generally useful, we may not need them now, but for sure they will be necessary next month.

I agree, the best thing we can do is to place all of them in few files and we will instruct the linter to don't run on those file.

In this way we respect the [single responsability principle](https://redbeardlab.com/2019/01/24/cb2-single-responsibility-principle/) and we will still keep the advantages of removing unused variables from the code.

---

### DIFF
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -196,9 +196,9 @@ def RPCPosition():
         hit = ROOT.MuonTaggerHit(detID,0)
         a,b = correctAlignmentRPC(hit,v)
         RPCPositionsBotTop[detID] = [a.Clone(),b.Clone()]
-        x = (a[0]+b[0])/2.
-        y = (a[1]+b[1])/2.
-        z = (a[2]+b[2])/2.
+        (a[0]+b[0])/2.
+        (a[1]+b[1])/2.
+        (a[2]+b[2])/2.
        # print "Postion for view {} and channel {}: ({}, {},{})".format(v,c,x,y,z)
 
 def GetRPCPosition(s,v,c):
@@ -260,8 +260,8 @@ def loadRPCtracks(n=1,draw=True,writentuple=False,fittedtracks=False):
   if fittedtracks:
    mH,bH = getSlopes(clustersH,0) 
    mV,bV = getSlopes(clustersV,1)   
-   trackH = ROOT.RPCTrack(mH,bH)
-   trackV = ROOT.RPCTrack(mV,bV)
+   ROOT.RPCTrack(mH,bH)
+   ROOT.RPCTrack(mV,bV)
   #print "Line equation along horizontal: {}*z + {}".format(mH,bH)
   #print "Line equation along vertical: {}*z + {}".format(mV,bV)
    theta = ROOT.TMath.ATan(pow((mH**2+mV**2),0.5))
--- a/charmdet/drifttubeMonitoring.py
+++ b/charmdet/drifttubeMonitoring.py
@@ -1,5 +1,8 @@
 #import yep
-import ROOT,os,time,sys,operator,atexit
+import ROOT
+import operator
+import os
+import time
 ROOT.gROOT.ProcessLine('typedef std::unordered_map<int, std::unordered_map<int, std::unordered_map<int, std::vector<MufluxSpectrometerHit*>>>> nestedList;')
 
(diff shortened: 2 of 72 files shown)


ThomasRuf:
Hi Simone,
looking at some of the modified lines, in fact, in most cases the whole line could go. 
For example: fitStatus = fittedTracks[tr].getFitStatus(), if fitStatus is not used, no reason to call the method getFitStatus(). 
On the other hand, rc = sTree.GetEvent(n)  is very useful in case you make a cut and paste and use the code in an interactive session. If you don't do this, you are swamped with the printout of how many bytes are read for each event. Similar when using the Fill method for histograms.

another observation, why making out if  import ROOT,os,sys,getopt three lines with import? Only because getopt is dropped? 

I agree with you, nice exercise to make the code more efficient.

siscia:
>  On the other hand, rc = sTree.GetEvent(n) is very useful in case you make a cut and paste and use the code in an interactive session. If you don't do this, you are swamped with the printout of how many bytes are read for each event. Similar when using the Fill method for histograms.

Thanks, I don't know what are the dynamics for this type of code.

Still, there are some places that I will urge somebody to check since they can really **hide bugs**, is not only a matter of performance, namely:

https://github.com/ShipSoft/FairShip/pull/236/files#diff-f8e244304fefdeb5b360f6c4e9494697L219
https://github.com/ShipSoft/FairShip/pull/236/files#diff-f8e244304fefdeb5b360f6c4e9494697L484
https://github.com/ShipSoft/FairShip/pull/236/files#diff-1c30e3b00b76eb3b2184ed83c457749dL399
https://github.com/ShipSoft/FairShip/pull/236/files#diff-7a5554a05333ad8140fda5dd04b1de1aL385
https://github.com/ShipSoft/FairShip/pull/236/files#diff-1972158418f8a656096314a81fd40818L268
https://github.com/ShipSoft/FairShip/pull/236/files#diff-8358a1d90376ce37fd95110cacc941b3L174

> another observation, why making out if import ROOT,os,sys,getopt three lines with import? Only because getopt is dropped?

In general is considered a bad practise to have import to differen modules in the same line with a comma, so when the linter goes to re-write the line, it follows the standard and write them on separated lines. Nothing to worry about.

---

When dealing with a lot of code, also professional software developer, try to get as much help as possible from automatic tools. 

Adding extra steps to the manual procedure is a price I will consider to pay, adding as much automation as possible really helps, especially when a lot of people from around the world need to contribuite to the same code base.

I will look to find a solution that can accomodate both needs and I will report back :)

olantwin:
Looking forward to the introduction of this and other linters!

I didn't have time to check all the changes (mostly look like changes that should not change behaviour), but noticed:

* `housekeeping/cpplint` probably shouldn't be changed, but just straight up removed
* same for `macro/obsolete`

ThomasRuf:
outdated

PR #240, opened 2019-07-04
Fix position of Pixel Volumes in geometry

owtscharenko:
Please excuse the second pull request today.
This one fixes the positions for the Volumes used for Simulation and the event display.


ThomasRuf:
Hi,
It is difficult for me to find out what really has changed in charm-geometry_config.py, except of removing many blanks. 

regarding units, my I mention that shipunits.py exists. This should be used to make it obvious which units are used, and it provides automatic translation to standard FairShip units used internally, see many examples in geometry/geometry_config.py.

owtscharenko:
Hi,

I changed the values for the single PixelModules.ySi() in lines 343, 347, 351 and so on. With the latest commit I used shipunit.py

ThomasRuf:
OK. Is there any possibility that you can commit a file with only these changes?

owtscharenko:
I removed the whitespace changes and updated my fork. Is this better?

ThomasRuf:
yes, much better.

PR #241, opened 2019-07-09
Change the way unpackers are selected

olantwin:
**Please do not merge before this has been tested!**
This is a first step to simplify the way unpackers are handled, allowing one to re-use unpackers for several partitions (as necessary for the pixel)

This also changes the way the unpackers need to be defined in the unpacking macro. The relevant changes are here: https://github.com/olantwin/muon_flux_online/tree/unpackers (I wonder whether the unpack.py script should be added to FairShip/macro dir, as it now doesn't need to compile and only depends on standard FairShip).

@owtscharenko @matclim @antonioiuliano2 : Would very much appreciate your feedback and testing: This should only affect the pixels and should not change anything for the other partitions.
 Once tested, I might clean it up a bit more, as I left the old way of adding unpackers in place to not break compatibility with the FairRoot FairOnlineSource.

### DIFF
--- a/online/PixelUnpack.cxx
+++ b/online/PixelUnpack.cxx
@@ -72,7 +72,7 @@ Bool_t PixelUnpack::DoUnpack(Int_t *data, Int_t size)
    for (auto &&hit : hits) {
       auto hitData = reinterpret_cast<HitData *>(&(hit.hitTime));
       auto channelId = reinterpret_cast<ChannelId *>(&(hit.channelId));
-      auto detectorID = (fPartitionId%0x0800) * 10000000 + 1000000 * hitData->moduleID + 1000 * channelId->row + channelId->column;
+      auto detectorID = (df->header.partitionId%0x0800) * 10000000 + 1000000 * hitData->moduleID + 1000 * channelId->row + channelId->column;
       auto tot = hitData->tot;
       new ((*fRawData)[fNHits]) ShipPixelHit(detectorID, tot); //tot is measured in steps of 25 ns
       fNHits++;
--- a/online/ShipTdcSource.cxx
+++ b/online/ShipTdcSource.cxx
@@ -1,3 +1,4 @@
+#include <stdexcept>
 #include "ShipTdcSource.h"
 #include "FairLogger.h"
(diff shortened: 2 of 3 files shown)


antonioiuliano2:
Dear Oliver,

thank you for your update. However, when I try to launch it with input file 
/eos/experiment/ship/data/charmxsec/DATA_8000_charmTest/rawdata/RUN_8000_2800/SPILLDATA_8000_0522408565_20180730_194153.raw

I get an error, followed by a segmentation violation : 

  File "/cvmfs/ship.cern.ch/SHiPBuild/sw/slc6_x86-64/ROOT/v6-14-00-ship-1/lib/ROOT.py", line 613, in __getattr2
    raise AttributeError( name )
AttributeError: SciFiUnpack
[INFO] PixelUnpack: Delete instance 

 *** Break *** segmentation violation

It seems your FairShip/unpackers fork has not yet been updated with SciFiUnpack, since I do not see the class in the online folder (but they are present in the ShipSoft master branch).

Best regards,
Antonio

olantwin:
Hi Antonio,
Of course you're right. Thanks for catching that!

I rebased the branch (so it should now include the SciFiUnpacker). It can however also be fixed by just replacing the SciFiUnpacker with a `ROOT.DummyUnacker(0x900)`, of one wants to isolate the pixel changes.

antonioiuliano2:
Hi Oliver,
Thank you for the update. However, it still does not work because the SciFiUnpack constructor requires the partitionID as an option (just as it happened with the master branch of muon_flux_online). 

I have added it (https://github.com/olantwin/muon_flux_online/pull/2), and now it seems to work (at least I get no errrors).
Best regards,
Antonio Iuliano

olantwin:
Thanks for your comment!
We should probably give SciFiUnpack a constructor without partition ID as well, as it in practice only ever has one partition ID.

ThomasRuf:
is this now being tested? Can I merge it?
thomas

olantwin:
@owtscharenko @antonioiuliano2 

I think any issues we discovered while testing are now fixed.

PR #242, opened 2019-07-10
Fix: x,y positions  

anabarbararc:
Fix on x and y positions on SciFiHit class


olantwin:
> I was just noticing that there are several _quite optimistic_ casts in the code.
> 
> Maybe is worth to check on those also.

Could you give an example?

In the diff I don't see any.

siscia:
Sorry,

like the very first line:

```c++
double mult = int((ch_layer)/64);
```

you have a double and you are putting in an int.
In the specific case we are lucky, double is always able to hold an int (is called a widening cast, the target type, double, is bigger than the original type int).

If we scroll a little up, here: https://github.com/ShipSoft/FairShip/blob/master/charmdet/SciFiHit.cxx#L25-L29

```c++
  unsigned int ch;
  unsigned int boardId;

  boardId = (fDetectorID) / pow(10,5);
  ch = (fDetectorID - boardId * pow(10,5));
```

look what happen in boardId.
Inside boardId we put (fDetectorID / 100000) where fDetectorID is an integer as well.

In this case, all the fDectectorID between 0 to (10E5)-1 will produce a boardId equal to zero.
All those between 10E5 and 2*10E5 will produce 2, and so on.

We can be lucky and never find situations where this is a problem, or we can be careful and try to avoid cast as much as possible.

If we use casts too freely, everybody will just start to add cast to quickly solve minor problems, resulting in very hard to trace bugs.

Casts produced bugs are one of the worst kind of bugs, because they are technically legal, but they often produce results that are wrong.

Cast is allowed in C/C++ for situations where you work very very close to the hardware, like microprocessor, we are definitely not in those, and if we can avoid them would be much better.

olantwin:
I would call this a conversion (some explicit, some implicit), not a cast (unless I'm mistaken, casting usually requires a pointer/address).

But you're right, these conversions could cause some nasty surprises!

ThomasRuf:
among many warnings, there is also an error and FairShip does not compile anymore:

/tmp/trufship/tmp.3lFVv4AICq/sw/SOURCES/FairShip/master/0/charmdet/SciFiHit.cxx:69:72: error: invalid operands of types 'double' and 'int' to binary 'operator%'
   double shift = int(mult/2) * gap_die + int(mult/2) * gap_SiPM + (mult%2) * gap_die;
 
i will revert the pull request, if there is no bug fix in the next 20minutes.

olantwin:
Could you try changing `double mult=...` back to `int mult=...`? Just looking at the code, that might fix the issue...

olantwin:
Not sure why that change was made, but doesn't seem like `mult` needs to be (or should be) a `double`

anabarbararc:
I just saw all the messages! 
I am going to check those issues! Thanks for pointing them out!

PR #248, opened 2019-07-12
Add CI in gitlab

siscia:
This PR add a simple CI that runs on gitlab

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@ cc7:
   image: alisw/alidock:cc7
   stage: build
   before_script:
-    - pip install alibuild --upgrade --user
+    - pip install --upgrade --force-reinstall alibuild
     - ls /cvmfs/ship.cern.ch
     - cd ..
     - git clone https://github.com/siscia/shipdist
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@ cc7:
   image: alisw/alidock:cc7
   stage: build
   before_script:
-    - pip install alibuild --upgrade
+    - pip install alibuild --upgrade --user
     - ls /cvmfs/ship.cern.ch
     - cd ..
     - git clone https://github.com/siscia/shipdist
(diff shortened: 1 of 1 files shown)


siscia:
# What is a CI and why we are interested in it.

CI stands for **C**ontinuos **I**ntegration and is a tool that allow to check that the code is in good shape.

As an example it is possible to run a build of FairShip automatically for every pull request, this help us making sure that the code compile without major problems before to merge a new pull request.

It happen for whatsoever reason that we can be distracted and we end up with a PR that doesn't compile. This helps everybody, both the maintainer, that can see quickly that everything is in order in the PR but also the contributor, that has another check in its own code.

Moreover it provides a simple and unique way to build the software.

# What else is possible

We can enhance the CI to:
1. run automatic tests 
2. check the quality of the code with linters
3. run the three steps of Simulation, Recostruction and Analysis, store the results and analyze the result before to merge a PR
4. publish on CVMFS (need a bit of time for this).

# CERN and CI infrastructure

CERN provide us a free to use CI infrastructure, but it is inside `gitlab.cern.ch`.

I create a public repository that mirror this one and that runs the CI on the CERN infrastructure, the repo is this one: https://gitlab.cern.ch/smosciat/FairShip

We should be able to run the CI with PR coming from Github, so after any PR in Github, we trigger the CI in gitlab, but I believe we need this PR merged first.

Unfortunately, is not possible to get directly a check behind each PR in Github from the CERN CI infrastructure.

However is possible to simply visit the CI page on gitlab to get the same checks, more tedious but it works.

# How does all this look like

For each run of CI we obtain something like this: https://gitlab.cern.ch/smosciat/FairShip/-/jobs/4802688

There is also an interface to see all the CI jobs: https://gitlab.cern.ch/smosciat/FairShip/pipelines

# How long does it take to run

After the last enhancement to the build ~5 minutes to complete a FairShip build, most of it is to set up the CI environement so it is not going to be much faster.

owtscharenko:
we could also use travis CI this is what we do in Bonn for our GitHub repositories

siscia:
Indeed I like travis a lot as well.

However the cern gitlab one has a very nice integration with cvmfs and reasonable fast machines, so I suggested that one... :)

PR #255, opened 2019-07-30
(WIP) First steps towards python3 compatibility

antonioiuliano2:
Dear FairShip expert,

during the last ShipSoft meeting, Oliver (@olantwin) proposed to start checking our code to ensure python3 compatibility: https://indico.cern.ch/event/833735/contributions/3493690/attachments/1878722/3094666/talk.slides.pdf

I then tried to start to do it  on two simple scripts I often use, getGeoInformation.py and CharmDetHitPosition.py, by updating the print functions and checking divisions between integers. Since, as suggested by Oliver, I imported the __future__ modules, so the code works on python2.7 without visible changes

I do not know if it is planned to put these updates and relative tests on a separate branch, so I wait for your suggestions on the matter.

Best regards,
Antonio Iuliano

### DIFF
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -1,9 +1,12 @@
 from __future__ import print_function, division
+from future import standard_library
+standard_library.install_aliases()
+from builtins import range
 import ROOT,os
 #ROOT.gROOT.ProcessLine('typedef std::unordered_map<int, std::unordered_map<int, std::unordered_map<int, std::vector<MufluxSpectrometerHit*>>>> nestedList;')
 import numpy
 from decorators import *
-import __builtin__ as builtin
+import builtins as builtin
 ROOT.gStyle.SetPalette(ROOT.kGreenPink)
 PDG = ROOT.TDatabasePDG.Instance()
 # -----Timer--------------------------------------------------------
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -410,8 +410,8 @@ def getSlopes(clusters,view=0):
 GetPixelPositions(nevent)    
 GetSciFiPositions(nevent)
 GetDTPositions(nevent)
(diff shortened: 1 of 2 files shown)


ThomasRuf:
Hi Antonio,
I created a few days ago the branch SHiP-2018, for working with the old SHiP geometry. The master branch should be used to move towards the SHiP-2020 geometry. So, it looks fine for me to integrate these changes into the master branch.

siscia:
Ciao Antonio,

I worked on it very quickly.

Let me suggest a couple of things to save you time, since I already faced these issues.

All the manual work can be done by automatic tools that Oliver pointed out and that you hopefully used as well. The conversion works quite well.

Python3 introduces an error in formatting if we mix spaces and tabs, and we do.

Overall we obtain good results running both futurize (https://python-future.org/overview.html#automatic-conversion-to-py2-3-compatible-code) for the mere migration to compatible code. And `autopep8` for the formatting part.

Good practises demand that pyhton imports are all at the beginning of file, which is not the case in our codebase. `autopep8` will try to fix those as well, breaking the code. The solution is to just fix the formatting issues with `autopep8` and it can be done with `autopep8 --select E1 file_to_fix.py`

Both tools can take as input a whole directory, which it may be interesting in your case.

Cheers,
Simone

antonioiuliano2:
Dear Simone,

thank you very much for your advices!
Since I am not familiar with the automatic tools, I was afraid to use them. 
I have now launched the following commands for the two scripts:

* futurize -w myscript.py
* autopep8 --select E1 myscript.py

futurize basically fixed some imports, while autopep8 did nothing.

siscia:
Ciao!

don't be afraid! We are using git for a reason.
It allows us to make big changes, test things, and without fear going back and forth between changes.

I am afraid you are not at CERN, otherwise, I would have invited you over and we could have played with git together.

Said so, it is possible that `myscript.py` already follows the things that `autopep8 --select E1` try to fix, and so it does not fix anything.

Cheers,
Simone

Issue #254, opened 2019-07-29
Some nutau volumes use medium vacuum instead of air


antonioiuliano2:
Dear FairShip experts,

checking the geometry implementation of the scattering and neutrino detector, I have noticed that some volumes use 'vacuum' as medium (see for example volTarget). However, the cave medium is air, not vacuum, as built by python/shipDet_conf.py.

Should I change the medium in these volumes, in order to better reproduce cave medium?

Best regards,
Antonio Iuliano

ThomasRuf:
Maybe using a volume is the wrong approach. If you only need an object which contains other volumes, an assembly might be the better option. 

thomas

antonioiuliano2:
Dear Thomas,

thank you very much for your advice, you actually pointed me to the recommended way to address these situations

However, since I have followed the wrong volume approach until now, I am going to need some time to check where all the virtual volumes are and replace them with assemblies (both in nutau and charmdet folder, at this point).

I will contact you as soon as I have any update.
Best regards,
Antonio Iuliano

antonioiuliano2:
Thank you for your advice,
I have just opened a pull request with these changes

PR #256, opened 2019-08-01
Replacing virtual volumes with assemblies

antonioiuliano2:
Dear FairShip experts,

following @ThomasRuf 's advice in https://github.com/ShipSoft/FairShip/issues/254, I have replaced the vacuum/air TGeoVolume containing other volumes with TGeoVolumeAssembly objects, in the scripts from the charmdet and nutaudet folders I have worked on.

Now there are no more unphysical vacuum volumes, so there should be no more medium discontinuity with cave medium.

I have not changed the volume names, so I assume backward compatibility should not be affected. I have run the ShipAna and ShipReco scripts and they worked without errors.

Please have a look and tell me if there is something I need to adjust or to improve.

Best regards,
Antonio Iuliano

### DIFF
--- a/charmdet/Box.cxx
+++ b/charmdet/Box.cxx
@@ -204,10 +204,7 @@ void Box::AddEmulsionFilm(Double_t zposition, Int_t nreplica, TGeoVolume * volTa
 void Box::ConstructGeometry()
 {
     InitMedium("tantalum");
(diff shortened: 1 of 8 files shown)


ThomasRuf:
after merging your previous commit, this one has now a conflict with nutaudet/NuTauMudet.cxx

PR #261, opened 2019-08-08
Veto sim

ghkorol:


ThomasRuf:
Hi Ievgen,
Didn't looked in detail into the code, but let me remark that replacing nav.LocalToMaster(origin,master)
with
    tr = node.GetMatrix().GetTranslation();
    rot = node.GetMatrix().GetRotationMatrix()
is maybe not safe. node.GetMatrix() will "only" give you the transformation from local to the next level up. It will not give you the transformation to the global coordinate system. Is this what you want?

ghkorol:
Yes, thank you!   In my case LocalToMaster did do nothing. And i did not
get why. I would prefer switch back to
nav.LocalToMaster(origin,master) , but i don know how to get it work, and i
need help at this point. Actually this could be the reason why I have shift
in Z.

ThomasRuf:
i think you are missing a nav.cd, see for example macro/getGeoInformation. 
...
node = digi.GetNode()
path = "cave/DecayVolume_1/T2_1/VetoLiSc_0/"
nav.cd(path+node.GetName())

ghkorol:
Hello Thomas,

Thank you.
Your suggestion solved my problem.

Cheers, Ievgen

пт, 9 авг. 2019 г. в 15:48, Thomas Ruf <notifications@github.com>:

PR #266, opened 2019-08-13
add python import after moving away from ROOT Sign

siscia:
Error catch by CI: https://gitlab.cern.ch/smosciat/FairShip/-/jobs/5129934


siscia:
Next CI clean: https://gitlab.cern.ch/smosciat/FairShip/-/jobs/5145800

PR #270, opened 2019-08-30
Transform code in order to work with python3, the steps:

siscia:
1. autopep8 --in-place --select E101, fix py3 error in mixing space and tabs
2. futurize -wn -1 **/*.py, make code py3 compatible
3. futurize -wn -2 **/*.py, add imports for py3

Then we hack the custom pickler in order to work with python3.
Reference: https://github.com/rootpy/rootpy/issues/808

### DIFF
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -22,7 +22,7 @@ vtop = ROOT.TVector3()
 rnr       = ROOT.TRandom()
 #-----prepare python exit-----------------------------------------------
 def pyExit():
- ut.errorSummary()
+    ut.errorSummary()
 #-----list of arguments--------------------------------------------------
 parser = ArgumentParser()
 parser.add_argument("-f", "--files", dest="listOfFiles", help="list of files comma separated", required=True)
@@ -45,34 +45,34 @@ nevent = int(options.nevent)
 
 fnames = []
 if options.catalog:
- tmp = open(options.listOfFiles)
- for x in tmp.readlines():
-  fname = x.replace('\n','')
-  if fname.find("root")<0:continue
-  f=ROOT.TFile.Open(fname)
-  sTree = f.cbmsim
-  if not sTree.GetBranch("FitTracks"): 
-   print("does not contain FitTracks: ",fname)
-   f.Close()
-   continue
-  fnames.append(fname)
-  fnames.append(x.replace('\n',''))
- tmp.close()
+    tmp = open(options.listOfFiles)
+    for x in tmp.readlines():
+        fname = x.replace('\n','')
+        if fname.find("root")<0:continue
+        f=ROOT.TFile.Open(fname)
+        sTree = f.cbmsim
+        if not sTree.GetBranch("FitTracks"): 
+            print("does not contain FitTracks: ",fname)
+            f.Close()
+            continue
+        fnames.append(fname)
+        fnames.append(x.replace('\n',''))
+    tmp.close()
 else:
- fnames = options.listOfFiles.split(',')
+    fnames = options.listOfFiles.split(',')
 fname = fnames[0]
 if options.updateFile:
- f=ROOT.TFile(fname,'update')
- sTree=f.Get('cbmsim')
- if not sTree: 
-   print("Problem with updateFile: ",f)
-   exit(-1)
+    f=ROOT.TFile(fname,'update')
+    sTree=f.Get('cbmsim')
+    if not sTree: 
+        print("Problem with updateFile: ",f)
+        exit(-1)
 else:
- sTree = ROOT.TChain('cbmsim')
- for f in fnames: 
-  print("add ",f)
-  if options.onEOS: sTree.Add(os.environ['EOSSHIP']+f)
-  else:             sTree.Add(f)
+    sTree = ROOT.TChain('cbmsim')
+    for f in fnames: 
+        print("add ",f)
+        if options.onEOS: sTree.Add(os.environ['EOSSHIP']+f)
+        else:             sTree.Add(f)
 
 #-------------------------------geometry initialization
 from ShipGeoConfig import ConfigRegistry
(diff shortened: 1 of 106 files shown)


JLTastet:
A lot (most?) of `old_div()` are unneeded. Is there any place in FairShip where we really need integer division?

ThomasRuf:
It is a common way to use integer division in the decoding of detector IDs. And it was very handy in python2. Don't understand why for python3 this had been dropped.

JLTastet:
Having a different behaviour for `/` depending on whether the operands are integers or floats is very risky in a dynamically typed language. In Python 3, the floor division `//` achieves the same result as the former integer division.

I will try to review the files I am familiar with (mostly physics models) to choose the appropriate division operator.

JLTastet:
I now get a segfault when exiting the Python shell after running the event display. This is deterministic and was not present with the Python 2 version.
```
$ python3 -i "$FAIRSHIP"/macro/eventDisplay.py -f ship.conical.Pythia8-TGeant4_rec.root -g geofile_full.conical.Pythia8-TGeant4.root
```
Output (incl. backtrace) attached: [py3_evtdisp_segfault.txt](https://github.com/ShipSoft/FairShip/files/3559835/py3_evtdisp_segfault.txt)

Relevant line: ```*** Error in `python3': free(): invalid pointer: 0x00007f78fdac5800 ***```

ThomasRuf:
until somebody knows how to fix it properly, you can try to activate again the evExit() method. To bypass a similar issue in older ROOT versions, there exists in eventDisplay.py   if float(ROOT.gROOT.GetVersion().split('/')[0])>6.07: atexit.register(evExit) 
change 6.07 to 6.19

JLTastet:
When sourcing `/cvmfs/ship.cern.ch/develop/setUp.sh` instead of `/cvmfs/ship.cern.ch/SHiP-2020/latest/setUp.sh`, I get the following error when trying to start Python 2:
```bash
[FairShip/latest-master-fairship] ~ $> python2
ImportError: No module named site
```
Is it expected? Python 3 works fine.

ThomasRuf:
why are you using python2 and python3 and not python?

under fairship python3 environment:
which python
/cvmfs/ship.cern.ch/develop/sw/slc7_x86-64/Python/v3.6.8-1/bin/python
 
i suspect your python3 or python2 points to somewhere outside ship.

thomas

JLTastet:
You are right. It seems that only Python 2 can be used with `SHiP-2020/latest`, and only Python 3 with `develop`.

Sourcing `/cvmfs/ship.cern.ch/SHiP-2020/latest/setUp.sh`:
```bash
[FairShip/latest-master-fairship] ~/Workdir2 $> which python
/cvmfs/ship.cern.ch/SHiP-2020/2019/August/12/sw/slc7_x86-64/Python/v2.7.10-1/bin/python
[FairShip/latest-master-fairship] ~/Workdir2 $> which python2
/cvmfs/ship.cern.ch/SHiP-2020/2019/August/12/sw/slc7_x86-64/Python/v2.7.10-1/bin/python2
[FairShip/latest-python3-fairship] ~ $> which python3
/usr/bin/python3
```
Sourcing `/cvmfs/ship.cern.ch/develop/setUp.sh`:
```bash
[FairShip/latest-python3-fairship] ~ $> which python
/cvmfs/ship.cern.ch/develop/sw/slc7_x86-64/Python/v3.6.8-1/bin/python
[FairShip/latest-python3-fairship] ~ $> which python2
/usr/bin/python2
[FairShip/latest-python3-fairship] ~ $> which python3
/cvmfs/ship.cern.ch/develop/sw/slc7_x86-64/Python/v3.6.8-1/bin/python3
```

siscia:
> A lot (most?) of `old_div()` are unneeded. Is there any place in FairShip where we really need integer division?

The point of the code transformation is to keep exactly the same semantic in the code. Unfortunately the semantic of the `/` operator changed from Py2 to Py3, hence we cannot use anymore the `/` operator but it was necessary to introduce this new function.

> Having a different behaviour for `/` depending on whether the operands are integers or floats is very risky in a dynamically typed language. In Python 3, the floor division `//` achieves the same result as the former integer division.
> 
> I will try to review the files I am familiar with (mostly physics models) to choose the appropriate division operator.

Yes, you are correct, is risky. Unfortunately Py2 works like that.

The python3 floor `//` operator is not quite the same as the python2 division operator.

```
[root@cvm-perf02 ~]# python2
Python 2.7.5 (default, Jun 20 2019, 20:27:34) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 4.0 / 3
1.3333333333333333
>>> 4 / 3
1
>>> exit()
[root@cvm-perf02 ~]# python3
Python 3.6.8 (default, Apr 25 2019, 21:02:35) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 4.0 / 3
1.3333333333333333
>>> 4.0 // 3
1.0
>>> 4 // 3
1
>>> 4 / 3
1.3333333333333333
```

JLTastet:
> The python3 floor `//` operator is not quite the same as the python2 division operator.

@siscia You are right indeed.

But at least for the files which I know sufficiently well, I will try to resolve the ambiguity and choose the correct operator. I can do that later, once this PR is merged, or directly send a PR to your branch.

siscia:
Is the runtime crash solved?

JLTastet:
> Is the runtime crash solved?

No, I haven't tried Thomas' suggestion yet. But the crash only happens when the program exits (and does not crash the SSH session), so it does not affect its function.

JLTastet:
I just compared the numerical values from the Pythia decay table for HNLs before / after, and they match perfectly (up to a slightly different rounding).

However, I see some large numerical differences for the RPV SUSY model. I will investigate them further, and make sure that they do not come from my testing procedure.

JLTastet:
I will try to run more systematic regression tests using [my old script](https://gist.github.com/JLTastet/f3b828b0295ea7337fb4d9712d28e965).

olantwin:
I'm reviewing the files I'm familiar with as well.

Concerning the division, I am only aware of the integer division being necessary for detector ID conversion. All other cases (e.g. for digitisation or raw data conversion), are in C++ so should not be affected.

JLTastet:
Ok, after fixing the floating-point formatting issues, all tests for the HNL model pass without regressions.

However, the lifetime and decay branching ratios can be vastly different between Python 2 and 3 for the RPV SUSY model. I will try to review it.

I have attached the [unordered diff](https://github.com/ShipSoft/FairShip/files/3563558/diff.txt).

JLTastet:
Regarding `rpvsusy.py`, there seems to be no difference between Python 2 and 3 on this branch (when varying this variable only).
However there is a clear difference between:
* Before: Python 2 + SHiP-2020/latest
* After: Python 3 + develop + this commit

Here is a minimal example showing the problem:
```python
from rpvsusy import RPVSUSY
b = RPVSUSY(1.,[1, 1],1e3,1,True)
b.computeNLifetime()
```
Before, the result was `1.3844222124690955e-08`. Now it is `1.8307090617172296e-25`.
That's a 17 orders of magnitude difference :scream:

JLTastet:
I found it! It seems that `futurize` messed up operator precedence when replacing `/` with `old_div`:
```diff
@@ -128 +134 @@ self.MW   = 80.385*u.GeV
-        self.gW2  = self.GF/math.sqrt(2)*8*self.MW*self.MW # SU(2)L gauge coupling squared
+        self.gW2  = old_div(self.GF,math.sqrt(2)*8*self.MW*self.MW) # SU(2)L gauge coupling squared
```
The second line should instead be:
```python
self.gW2  = old_div(self.GF,math.sqrt(2))*8*self.MW*self.MW
```

JLTastet:
The division issue is a [known bug](https://github.com/PythonCharmers/python-future/issues/434) in `futurize`. There is currently no fix for it, but it seems to be planned for the 0.18 release.

@siscia Could you re-run `futurize`, but separately commit E101, stage 1, stage 2 without division, and finally just the divisions? This will ease the process of reviewing the changes.

1. `autopep8 --in-place --select E101`
2. `futurize -wn -1 **/*.py`
3. `futurize -wn -2 --nofix=division_safe **/*.py`
4. `futurize -wn --fix=division_safe **/*.py`

siscia:
Nice catch!

I am a little busy at this point, do you think you can take this over? If it is a problem for you I can do it, but it will be a little slower...

JLTastet:
I will see what I can do. Stage 1 should be safe (by design), but I fear that stage 2 will require each file to be reviewed by someone familiar with it. Not sure whether E101 can be breaking. We have a lot of crazy indentation in FairShip.

At the very least, I should be able to convert the physics models to Python 3 and make sure they are free of regressions. However, I can't do much for the rest of the code, since I have currently no way of running regression tests for the entire simulation (my tests only compare the Pythia configuration).

This will take some time too, since I am working on multiple projects in parallel (including a new physics model, which fortunately is already compatible with Python 3 and is well covered by tests).

siscia:
Honestly I did not run any test, except making sure that the plot before and after where similar. But I am not a physicist myself.

E101 should be safe/required.

ThomasRuf:
outdated

Issue #271, opened 2019-08-31
Missing space in Pythia string for RPV SUSY


JLTastet:
I just came across this by chance. The following two lines:

https://github.com/ShipSoft/FairShip/blob/4911ea5eed6c9dd445ed0d31728b7a7bcd8e0f2e/python/pythia8_conf.py#L103-L104

result in an invalid Pythia string:
```
521:new  B+               B-    1   3   0    5.279250.00000    0.00000    0.00000  4.91100e-01   0   1   0   1   0
                                             ^^^^^^^^^^^^^^
```
Not sure if it affects any results. I am not familiar with this model, and I haven't checked yet if it contains similar issues.

PR #272, opened 2019-08-31
Explicitly specify the format of floats in Pythia strings for HNLs

JLTastet:
The default format used for printing floating point values can vary across Python and Numpy versions. This tends to break string-based regression tests.

This PR explicitly formats all floating point values for the HNL model, while trying to keep the same number of significant digits as before (±1).


JLTastet:
Don't merge it yet. I will push a few more changes.

JLTastet:
I ran a few regression tests (not the whole suite), and they all passed.

This PR also fixes #271.

I have these changes both on the master and Python 3 branches, so we can merge them either before or after #270 or #274 is merged.

JLTastet:
Added a few dark photon fixes.

This should be ready for merging now. If possible, please do not squash the commits, in order to preserve their messages.

JLTastet:
Closes #271.

PR #274, opened 2019-09-02
Python3 safe division

siscia:
PR as discussed in #270

### DIFF
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -22,7 +22,7 @@ vtop = ROOT.TVector3()
 rnr       = ROOT.TRandom()
 #-----prepare python exit-----------------------------------------------
 def pyExit():
- ut.errorSummary()
+    ut.errorSummary()
 #-----list of arguments--------------------------------------------------
 parser = ArgumentParser()
 parser.add_argument("-f", "--files", dest="listOfFiles", help="list of files comma separated", required=True)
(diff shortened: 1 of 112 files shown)


siscia:
Honestly I didn't even run the test on this PR.

JLTastet:
Nice! Note that `division_safe` is the same fixture as before, which uses `old_div`. So the division bug should still be present, but the code will be easier to review / fix now that each stage has its own commit.

JLTastet:
I will try to run the tests a bit later on Stage 2 without division. If they pass, then we can add `from __future__ import division` just for physics models, hence defaulting to real division in both Python 2 and 3.

JLTastet:
I ran the tests again without the `old_div`, and they all pass now. I will extend them to include the Dark Photon model, and if they still pass I will explicitly import the real division in all the relevant files and open a PR.

@siscia I think it would be a good idea to get all your commits *except* the Stage 2 ones merged as soon as possible (*). Then we can "futurize" and validate individual files on a case-by-case basis, until they all support Python 3. But if we futurize everything at once, I fear we will introduce more silent breakage.

_(*) We also need to keep in mind that indentations changes etc. will cause a lot of conflicts when rebasing, so if anyone has a long-lived feature branch lying around, it might be a good idea to get it merged ASAP. Maybe we could make an announcement on the mailing list?_

JLTastet:
After a few fixes (cf. #272), dark photon tests pass too.
I will modify the relevant files tomorrow to use the real division where it makes sense.

JLTastet:
#275 defaults to using real division in physics models.

ThomasRuf:
outdated

PR #275, opened 2019-09-03
Use real division in physics models and for units

JLTastet:
These files never use integer division, so I have added `from __future__ import division` to them in order to default to real division.

I checked it either manually or by comparing the output of automated tests between Python 2 and 3.


JLTastet:
Closes siscia/FairShip#4.

PR #276, opened 2019-09-06
Don't write empty event for SoS when processing single partition

olantwin:
@anabarbararc @antonioiuliano2 @leoredi


ThomasRuf:
Hallo Oliver,
could you explain what this change means for the existing muflux data? Are there any errors which would be fixed when rerunning the eventbuilder machinery? 
thomas

olantwin:
Hi Thomas,
For existing data there should be absolutely no difference (I did some checks with fabricated raw data to confirm this):
Essentially the conversion of the pre-EvB data was updated to have all the same features and checks of the post-EvB data, so for the post-EvB data nothing changes.
Oliver

PR #277, opened 2019-09-09
set center of RPCs at x=0

evanherwijnen:
Set x=0 at center of RPCs. Add aluminium panels.


ThomasRuf:
Hi Eric,
Not sure what to do. I am afraid that this change could cause potential problems if it is not consistently used for MC and Data. Currently, I am using an alignment correction to align DT tracks with RPC. With this modification, I would need to remove this correction in case the data is being re-reconstructed. I can merge it on the master branch, but continue for the muflux branch without, until all the Data is being again reconstructed.

evanherwijnen:
OK that is fine with me, but it would be nice if this fix corrects the problem.

PR #284, opened 2019-09-13
Dictionary fixes in order to support Python 3.

JLTastet:
Run `futurize -wn --fix=dict **/*.py`, then manually discard some changes where
wrapping the iterator in a `list()` is unnecessary.
- When iterating over a temporary object, keep the iterator.
- When checking membership (`in`) with a temporary object, keep the iterator.

### DIFF
--- a/charmdet/drifttubeMonitoring.py
+++ b/charmdet/drifttubeMonitoring.py
@@ -56,7 +56,7 @@ viewsI = {1:[0,1],2:[0,2],3:[0],4:[0]}
 viewC = {0:"_x",1:"_u",2:"_v"}
 
 muSources = {'eta':221,'omega':223,'phi':333,'rho0':113,'eta_prime':331}
-muSourcesIDs = muSources.values()
+muSourcesIDs = list(muSources.values())
 rnr       = ROOT.TRandom()
 #-----prepare python exit-----------------------------------------------
 def pyExit():
@@ -780,7 +780,7 @@ DT={}
 def compareAlignment():
  ut.bookHist(h,'alignCompare','compare Alignments',100,-120.,120.,100,-120.,120.)
  ut.bookHist(h,'alignCompareDiffs','compare Alignments',100,-1.,1.)
- keys = xpos.keys()
+ keys = list(xpos.keys())
  keys.sort()
  for d in keys:
    test = ROOT.MufluxSpectrometerHit(d,0.)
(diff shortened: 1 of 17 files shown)


JLTastet:
Let's get started with stage 2 changes!

I will make some tests on my side.

Please go through the changes and comment / report issues here.

JLTastet:
`run_simScript.py` / `ShipReco.py` / `ShipAna.py` / `eventDisplay.py` work for me. Go ahead and test the scripts you are familiar with to increase coverage.

ThomasRuf:
looks ok for me, except that I again don't understand the Python developpers, why more work is put on the users. Instead of providing a method which returns the list of keys as before, the user has to wrap dict.keys() around list(dict.keys()) to get previous result. Maybe one should add another method to type 'dict', lkeys(). 
There is one issue with list(dict.keys()), it is not thread safe, see:
https://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3

JLTastet:
`d.keys()` is lazy in Python 3: it does not need to allocate a new array to store the keys (like Python 2 did). Wrapping it in `list()` restores the old / less efficient behaviour.

JLTastet:
This is why I do not wrap `d.keys()` in `list()` if the keys are only iterated over once. By default `futurize` is too conservative and always wrap it in `list()`.

ThomasRuf:
what do you think about the objection that the list() solution is not thread safe?

JLTastet:
I'll have a look at it. Not sure it matters to us.

JLTastet:
Do we really use multiple threads anywhere in FairShip?

ThomasRuf:
don't think so. In any case, looking at some of the list() constructs, I see two cases:
1) header_keys = list(include_dict.keys())
    for header in header_keys:
    This can be easily replaced with  for header in include_dict: ...
2)  keys = list(pdg.keys())
     keys.sort()
     for n in range(len(keys)):
     maybe there exists a better way to make a sorting without detour through a list.

JLTastet:
> what do you think about the objection that the list() solution is not thread safe?

My (limited) understanding of the issue is that `dict.keys()` being atomic was more of an implementation detail in Python 2, and should not be relied upon.

If a `dict` is to be shared between multiple threads, then it should be locked whenever it is being mutated, in order to prevent simultaneous reads. Doing otherwise is fundamentally thread-unsafe (not just when calling `.keys()`) and having the code throw an exception is a lesser evil.

Regarding your suggestions, I think they would make the code cleaner, so I will try to implement them. But if we ever need concurrent access to a `dict`, we should use a mutex.

JLTastet:
> maybe there exists a better way to make a sorting without detour through a list.

`sorted(dict.keys())` should do the trick, but in any case it has to allocate at least one list for the output.

PR #285, opened 2019-09-13
Use Python 3 `range`.

JLTastet:
Run `futurize -wn --fix=xrange_with_import **/*.py`, then discard useless `list()` where the lazy Python 3 behaviour is desired.

Discard the imports in files where `range`/`xrange` is not used.

There are very few places where we want to keep the Python 2 behaviour. In these rare cases, I just kept `range` wrapped in a `list()`.


JLTastet:
The basic simulation workflow works fine. Please test and review these changes when you have time.

PR #286, opened 2019-09-18
use ArgumentParser instead of getopt

ThomasRuf:
I started to replace getopt with ArgumentParser, which is much more user friendly. I started with run_simScript.py. There shouldn't be any change in the way of executing run_simScript.py. Please check and let me know if something does not work, which worked before. 

run_simScript.py --DarkPhoton does not work, but also didn't worked before the change.

### DIFF
--- a/macro/ShipAna.py
+++ b/macro/ShipAna.py
@@ -7,57 +7,48 @@ from ShipGeoConfig import ConfigRegistry
 from rootpyPickler import Unpickler
 from decorators import *
 import shipRoot_conf
-shipRoot_conf.configure()
+from argparse import ArgumentParser
 
-debug = False
-chi2CutOff  = 4.
+shipRoot_conf.configure()
 PDG = ROOT.TDatabasePDG.Instance()
-inputFile  = None
-geoFile    = None
-dy         = None
-nEvents    = 9999999
+
+chi2CutOff  = 4.
 fiducialCut = False
 measCutFK = 25
 measCutPR = 22
 docaCut = 2.
-try:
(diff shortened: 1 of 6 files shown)


olantwin:
I am really happy about this change, as argumentparser is much easier to work with (and easier to debug etc.).

Two suggestions for improvement:

* In most cases, `dest=` does not need to be specified, as automatically the values will be stored in e.g. `--longoption`  → `options.longoption`.
* Instead of having dozens of default variables defined in the beginning, we can use the `default=` keyword argument to define the defaults, if they are not needed elsewhere in the file.

Happy to help with the transition of the other scripts and will have a more careful look if I notice anything in this script later.

ThomasRuf:
for first point, if dest is specified, it cannot hurt. 
the second point, I don't understand.

olantwin:
Some more comments:

* `required=` defaults to `False`. While it doesn't hurt to explicitly specify it, as we use many arguments, it distracts and reduces readability in my opinion.
* We could use the [`type`](https://docs.python.org/2/library/argparse.html#type) and [`choices`](https://docs.python.org/2/library/argparse.html#choices) keywords to perform simple validation of inputs, as we have several cases, where we expect e.g. a number or one of a certain list of choices.

olantwin:
> for first point, if dest is specified, it cannot hurt.

While it can't hurt in the sense that the resulting behaviour is the same (or equivalent), I does think it reduces readability (but that's of course subjective)

> the second point, I don't understand.

Sorry, I noticed that in most cases where it does make sense (e.g. `nEvents`, `tankDesign`) the `default=` keyword is already in use.

PR #288, opened 2019-09-18
Veto sim

ghkorol:
Update of naming scheme. Place vacuum inside decay volume.


ThomasRuf:
There is still an issue with checking the fiducial volume. When extrapolating to the centre of the magnet, z=3068cm, and checking for the volume, cave_1 is found instead of DecayVacuum_XXX. While there is vacuum in front of the tracking stations, I think there is no vacuum between the tracking stations, and probably after the tracking stations.

ghkorol:
Hello Thomas,

Yes, you are right !

I think I know the reason why it's heaped.

I will discuss it with you tomorrow by vidyo.

Cheers, Ievgen

PR #291, opened 2019-09-26
Most remaining stage 2 changes

JLTastet:
Everything except changes related to division or standard library.

~**Do not merge yet:** needs testing & review.~

There might be a better way to implement https://github.com/ShipSoft/FairShip/commit/e0d740b7461c7d12a76e32e53b0dd12ebd51492d, and the variable affected by https://github.com/ShipSoft/FairShip/commit/95292a8d0443e35066ea895a2e80d2a3addefc30 seems unused. Please have a look at it.


olantwin:
Concerning the changes to cpplint, I suggest we should just remove it altogether. It's an external package that is not used by FairShip.

* They are probably porting it to python3 upstream, so we don't need to do it ourselves.
* If anyone actually wants to use cpplint, it's a much better idea to install it using the normal methods (system package manager).

@ThomasRuf what do you think?

olantwin:
@JLTastet are you sure we need to add the inheritance to `object` in 14edb09475ca72c82b71003299e9c28ec6c1e4a1 ? See e.g. https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x

ThomasRuf:
I never used anything from FairShip/housekeeping. Is our code of cpplint a FairRoot/FairShip specific adaptation of an external package? In principle, I agree with Oliver to remove it and rather add cpplint to the list of python modules to install in shipdist.

olantwin:
You'd need to ask Fons to make sure, but as far as I remember it was just a straight up copy without specific modifications.

JLTastet:
> @JLTastet are you sure we need to add the inheritance to `object` in [14edb09](https://github.com/ShipSoft/FairShip/commit/14edb09475ca72c82b71003299e9c28ec6c1e4a1) ? See e.g. https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x

You are right. I never realized that explicit inheritance from `object` was there for forward-compatibility (because `future` probably redefines it in Python 2). I'll remove it.

JLTastet:
Regarding `cpplint.py`, I will remove the changes from this PR. It can be handled later if needed.

JLTastet:
The basic workflow runs without crashing, and the results look meaningful.
I'll try to do more tests after lunch.

JLTastet:
Two things still need to be sorted out before this can be merged:
- I have no idea how to run `charmdet/drifttubemonitoring.py`. Can someone quickly test it?
- `shipStrawTracking.py` currently segfaults on master, so I cannot test it.

The rest works fine.

ThomasRuf:
forget about charmdet/drifttubemonitoring.py. This is maintained on the muflux branch, no need to convert to python3.

JLTastet:
> forget about charmdet/drifttubemonitoring.py.

Ok, I'll remove its commit

olantwin:
Should we then remove `drifttubeMonitoring.py` from `master`? Anyone who needs to use it, can use the `muflux` branch instead of this outdated version on `master`.

JLTastet:
Has anyone had a look at `shipStrawTracking.py`?

ThomasRuf:
How should we procceed with this request? Is it ready to be merged?

JLTastet:
@ThomasRuf @olantwin I have fixed the conflicts and added two more Python 3 fixes. The segfault in `shipStrawTracking.py` is still there, but since it has not been fixed in half a year, it is probably not very important :)

So I think this PR is ready to be merged, and essentially completes the Python 3 transition (at least as far as the code is concerned).

PR #292, opened 2019-09-26
`Tkinter` -> `tkinter` & add aliases for Python 2.

JLTastet:
`futurize -wn --fix=future_standard_library **/*.py` and select Tkinter fixes.

~**Do not merge yet:** needs testing.~

I think it is just a rename, so it should be unproblematic.


JLTastet:
The event display seems to work just fine. I am not very familiar with it though, so I might not be testing it fully.

This is probably safe to merge.

ThomasRuf:
If any problem comes up later, we will just fix it.

PR #293, opened 2019-09-26
`cPickle` -> `pickle` & add aliases for Python 2.

JLTastet:
`futurize -wn --fix=future_standard_library **/*.py` and select relevant fixes.

~**NEEDS REVIEW** (and testing):~ `pickle` is currently listed as unsupported in the `futurize`
  documentation. Someone who is familiar with pickling should have a closer look.

If it does not work, I guess we can just test for `sys.version_info[0]` like we did for the ROOT pickler.

### DIFF
--- a/python/ShipGeoConfig.py
+++ b/python/ShipGeoConfig.py
@@ -1,6 +1,8 @@
+from future import standard_library
+standard_library.install_aliases()
 import os
 import re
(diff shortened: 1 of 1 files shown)


JLTastet:
This one seems to work just fine, too. It can probably be merged.

PR #294, opened 2019-09-26
WIP: `__builtin__` -> `builtins` & add aliases for Python 2.

JLTastet:
`futurize -wn --fix=future_standard_library **/*.py` and select relevant fixes.

It seems like this should be a simple rename, so I would expect it to be safe.

**NEEDS REVIEW** (and testing): Not sure why we use `__builtin__` in the first place. Someone who is familiar with this usage should have a look to make sure that `builtins` is indeed equivalent.

### DIFF
--- a/charmdet/drifttubeMonitoring.py
+++ b/charmdet/drifttubeMonitoring.py
@@ -1,10 +1,12 @@
 from __future__ import print_function
 #import yep
+from future import standard_library
+standard_library.install_aliases()
 import ROOT,os,time,sys,operator,atexit
 ROOT.gROOT.ProcessLine('typedef std::unordered_map<int, std::unordered_map<int, std::unordered_map<int, std::vector<MufluxSpectrometerHit*>>>> nestedList;')
 
 from decorators import *
-import __builtin__ as builtin
+import builtins as builtin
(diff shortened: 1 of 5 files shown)


olantwin:
As far as I can tell, we use `builtin` exclusively to create global variables. There must be a better way to do this...

JLTastet:
> As far as I can tell, we use `builtin` exclusively to create global variables. There must be a better way to do this...

Apparently, the "canonical" way to create globals is to do exactly this, except with a custom module of ours (e.g. `globals.py`) instead of using `builtin`.

Ref: https://docs.python.org/3/faq/programming.html#how-do-i-share-global-variables-across-modules

olantwin:
How about creating a `ship.globals` module then? That would also avoid possible collisions with actual builtin values or functions.

JLTastet:
> How about creating a `ship.globals` module then?

This seems to be the best solution indeed.
Do you want to do it? Finding the globals should be straightforward (at least for those using `builtin`).
In the meantime I'll test the other changes.

olantwin:
I'll look into it.

ThomasRuf:
Can I close this request?

JLTastet:
Yes, this is now superseded by #295.

PR #295, opened 2019-09-27
Use global_variables module instead of polluting __builtins__

olantwin:
As discussed in #294, I replace the use of `__builtins__` to create global variables with a dedicated dummy module.

While this should be quite a trivial change, **please test!**

While making the replacement, I noticed that while several script initialise global variables, I did not see them used anywhere. In that case, it might make sense to remove them if they are really unused, or use them, if this is by accident.

### DIFF
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -6,7 +6,7 @@ import ROOT,os
 #ROOT.gROOT.ProcessLine('typedef std::unordered_map<int, std::unordered_map<int, std::unordered_map<int, std::vector<MufluxSpectrometerHit*>>>> nestedList;')
 import numpy
 from decorators import *
-import config
+import global_variables
 ROOT.gStyle.SetPalette(ROOT.kGreenPink)
 PDG = ROOT.TDatabasePDG.Instance()
 # -----Timer--------------------------------------------------------
@@ -77,7 +77,7 @@ else:
 #-------------------------------geometry initialization
 from ShipGeoConfig import ConfigRegistry
 ShipGeo = ConfigRegistry.loadpy("$FAIRSHIP/geometry/charm-geometry_config.py", Setup = 1, cTarget = 1)
-config.ShipGeo = ShipGeo
+global_variables.ShipGeo = ShipGeo
 import charmDet_conf
 run = ROOT.FairRunSim()
 run.SetName("TGeant4")  # Transport engine
--- a/charmdet/CharmdetHitPositions.py
+++ b/charmdet/CharmdetHitPositions.py
@@ -6,7 +6,7 @@ import ROOT,os
 #ROOT.gROOT.ProcessLine('typedef std::unordered_map<int, std::unordered_map<int, std::unordered_map<int, std::vector<MufluxSpectrometerHit*>>>> nestedList;')
 import numpy
 from decorators import *
-import globals
+import config
 ROOT.gStyle.SetPalette(ROOT.kGreenPink)
 PDG = ROOT.TDatabasePDG.Instance()
 # -----Timer--------------------------------------------------------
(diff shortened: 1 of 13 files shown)


ThomasRuf:
ShipReco.py does not run:
NameError: name 'realPR' is not defined
  File "/media/truf/disk1/SHiPTest/sw/ubuntu1804_x86-64/FairShip/master-1/python/shipDigiReco.py", line 4, in <module>
    if realPR == "Prev": import shipPatRec_prev as shipPatRec # The previous version of the pattern recognition

ThomasRuf:
What confuses me, globals is a python <built-in function globals>, which is overwritten by import globals to become <module 'globals' from '/media/truf/disk1/SHiPTest/sw/ubuntu1804_x86-64/FairShip/master-1/python/globals.pyc'> .

Is this the desired behaviour?

olantwin:
Thanks for letting me know. I'll look into it.

I suspect the import order could be an issue.

`globals` should always be a python module, which different variables are added to.

olantwin:
Now I understand why the `__builtins__` module was used: By using it, the variables become available in the global namespace automatically without having to import `__builtins__` or having to specify e.g. `__builtins__.realPR`, which means we now have to find every place this behaviour was exploited...
It will probably take a few commits to find and replace all of them. At least this explains why the global variables seemed unused.

olantwin:
Also, I'll need to rename the module, as `globals()` is a builtin function.

olantwin:
@ThomasRuf: Is `python/shipPatRec_config.py` used at all? Apart from your initial commit, it hasn't been modified, and it doesn't seem to be imported anywhere.

ThomasRuf:
looks like it was an attempt to move booking of histograms from shipPatRec_prev.py to somewhere else. I will delete it.

ThomasRuf:
instead of the name config, could we use something which reflects more the purpose? commonBlock, globalSpace, globalVariables, globalParameters?

olantwin:
I like `globalVariables`, but following the [python naming convention](https://www.python.org/dev/peps/pep-0008/#package-and-module-names), I'd probably make it `global_variables`.

olantwin:
@ThomasRuf , could you please test the scripts defining globals, i.e.:

- [ ] `charmdet/CharmdetHitPositions.py`
- [ ] `charmdet/drifttubeMonitoring.py`
- [ ] `macro/MufluxReco.py`
- [ ] `macro/runMufluxDigi.py`
- [ ] `macro/ShipReco.py`

?

`ShipReco.py` runs without crashing now.

There might still be some uses of the global variables I missed.
Unfortunately, they are quite hard to search for, and shadowing rules etc. sometimes lead to non-obvious behaviour.

I also used the opportunity to reformat the changed lines a bit.

Once everything looks OK I'll rename the module e.g. `global_variables` and then this is ready to merge.

ThomasRuf:
Not much testing for me:
charmdet/CharmdetHitPositions.py is for Antonio to test, I guess.
charmdet/drifttubeMonitoring.py is out of date, only maintained in muflux branch.
macro/MufluxReco.py  used for muflux simulation prior to testbeam,  never worked really for real data. I am not using it.
macro/runMufluxDigi.py maintened in muflux branch.

olantwin:
Thanks for letting me know! For the scripts only maintained in the `muflux` branch, I would suggest removing them from `master` (alongside the scripts in `python/` only used by them), as keeping around unmaintained files there doesn't serve any purpose I can think of.

@antonioiuliano2: Which of these scripts are used for the charm measurement? Could you please test whether they work?

antonioiuliano2:
Hi Oliver,

of these scripts, the only script I use is charmdet/CharmdetHitPositions.py. However, as soon as I have tested your changes (2020/master build), it complains that I have a missing 'config' module. So we probabily need to add it to the build.

Best regards,
Antonio

olantwin:
Then I will proceed with the final rename.

I will open a separate pull-request to remove the unmaintained and unused scripts.

olantwin:
The rename is complete and this pull request is now ready for merging.

PR #303, opened 2019-10-25
FairLogger python interface (version 1)

olantwin:
Up to now it was not possible to use FairLogger for logging in python, resulting in us using `print` statements for logging in python and FairLogger in C++.

This pull request adds an interface to FairLogger for python (2 or 3), which can be used as follows:
```python
from logger import info, warn, error, debug
warn("test") # > [WARN] test
error(5) # > [ERROR] 5
info("test") # > [INFO] test
debug(42) # is not printed by default
```
The name of the module could be easily changed to something else (e.g. `shiplog`).

Current limitations of this first version (to be fixed in future):
* Currently if several arguments are passed, they are printed on separate lines.
* Types not understood by PyROOT or not printable in C++ might not work (in that case, we recommend logging a pre-formatted string.
* There are some opportunities for optimisation and clean-up in python and the C++ code.
* The C++ header is not integrated with CMake yet.
* When experimental PyROOT becomes stable, this should become much simpler and more powerful!

For most use-cases the interface should however be already sufficient.

This pull request is one of the results of the Starter Kit hackathon (hopefully the first of many to come!), and is mainly the work of @JLTastet .

We will probably port e.g. `run_simScript.py` to use this interface as an example, including to show how the log-level can be modified.

### DIFF
--- a/python/logger.py
+++ b/python/logger.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+
+import os
+import ROOT
+# TODO: add header to CMake
+ROOT.gROOT.ProcessLine('#include "' + os.path.join(os.path.expandvars('$FAIRSHIP'), 'utils', 'logger.hxx') + '"')
+
+for func in ('fatal', 'error', 'warn', 'info', 'debug'):
+    # The double lambda is used in order to capture 'func' by value, not by reference
+    # (which would always point to the last value taken in the loop)
+    globals()[func] = (lambda f:
+        lambda *args: ROOT.ship.__dict__[f](*map(type, args))(*args)
+    )(func)
+    globals()[func].__doc__ = \
+        'Python wrapper around the C++ function template ship::{}(Types... args).'.format(func)
--- a/utils/logger.hxx
+++ b/utils/logger.hxx
@@ -1,0 +1,53 @@
+#include "fairlogger/Logger.h"
+
+namespace ship {
+
+template <fair::Severity severity, class T>
+void log(T arg)
+{
+   for (bool fairLOggerunLikelyvariable = false; fair::Logger::Logging(severity) && !fairLOggerunLikelyvariable;
+        fairLOggerunLikelyvariable = true)
+      fair::Logger(severity, __FILE__, CONVERTTOSTRING(__LINE__), __FUNCTION__).Log() << arg;
+}
+
+template <fair::Severity severity, class T, class... Types>
+void log(T arg, Types... args)
+{
+   log<severity>(arg);
+   log<severity>(args...);
+}
+
+template <class T>
+void log(T arg)
+{
+   constexpr fair::Severity severity = fair::Severity::INFO;
+   for (bool fairLOggerunLikelyvariable = false; fair::Logger::Logging(severity) && !fairLOggerunLikelyvariable;
+        fairLOggerunLikelyvariable = true)
+      fair::Logger(severity, __FILE__, CONVERTTOSTRING(__LINE__), __FUNCTION__).Log() << arg;
+}
+
+template <class... Types>
+void warn(Types &&... args)
+{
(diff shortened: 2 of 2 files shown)


JLTastet:
BTW, I have a working prototype (C++ and Python) which can print multiple arguments on the same line.

We still need to decide how to separate the different arguments (spaces?)

olantwin:
Oh, great! Maybe optional argument defaulting to spaces? Feel free to make a pull request against my original repo to update this pull request :)

JLTastet:
I am trying to figure out what is the default behaviour when using `LOG(severity) << msg << msg2 << ... ;` but it seems that ROOT does not know about FairLogger.
```
fair::Logger()
ROOT_prompt_4:1:1: error: 'fair' is not a class, namespace, or enumeration
fair::Logger()
^
ROOT_prompt_4:1:1: note: 'fair' declared here
```
Any idea how to tell ROOT about FairLogger? I am using the `SHiP-2020/latest` environment.

JLTastet:
Ok, the default C++ behaviour seems to concatenate the string representations of the passed objects:
```c++
root [11] LOG(info) << "Msg1" << "Msg2" << "Msg3" << 4 << 5.6;
[INFO] Msg1Msg2Msg345.6
```
So I will try to keep the same default behaviour for C++, but add an optional `sep` argument which will be set to space by the python wrapper, in order to be consistent with the behaviour of `print()`:
```python
>>> print('a', 'b', 'c')
a b c
```

olantwin:
I think it might make sense to follow the behaviour of `print()` in python, similar to how `LOG(info)` follows the behaviour of `std::cout`.

That way, the interface should be as familiar as possible to new users coming to FairShip.

JLTastet:
Unfortunately, I realized that implementing default arguments with variadic templates, although possible, would add a lot of complexity to the code.

So I think we can either have a clean C++ interface but require the user to include the spaces in Python, or a more complicated C++ interface (which always requires specifying the separator string) with the correct `print()` behaviour in Python, or both at the same time at the cost of a significantly more complex C++ code (and more complex error messages...).

I would tend to choose the first solution, i.e. simple a C++ interface and implementation, and ask the user to explicitly put spaces / separators. For the first version at least.

olantwin:
I think the C++ version would never need to be called by the user: for that we can directly use the macros provided by `FairLogger` (or am I missing something?).

JLTastet:
So your idea would be to directly use the FairLogger macros in the C++ code, and use our new interface for the Python bindings only?

Thinking about it again, it might also be possible (and easier!) to add the delimiters on the Python side. I will try to do that. If it works, it might be the best option.

olantwin:
Yes, that would be my suggestion. The C++ interfaces provided by FairLogger are pretty flexible (and they even added more in the most recent release), so I think we can use them directly in C++, as we already partially do.

Thanks!

JLTastet:
I have added a `delim` keyword argument to the Python wrappers. It did not require modifying the C++ code at all, and its behaviour now matches that of the `print()` function.

**Typical usage:**
```python
>>> from logger import *
>>> error(1, 2, "Test", 3.14159)
# [ERROR] 1 2 Test 3.14159
>>> error(1, 2, "Test", 3.14159, delim='')
# [ERROR] 12Test3.14159
>>> error(1, 2, "Test", 3.14159, delim=' | ')
# [ERROR] 1 | 2 | Test | 3.14159
```
```
>>> help(error)
Help on function function in module logger:

function(*args, **kwargs)
    Python wrapper around the C++ function template ship::error(Types... args).
    
    Keyword arguments:
    delim -- delimiter to be inserted between each argument (default ' ')
```

olantwin:
Since this way merged before I had a chance to merge @JLTastet's latest improvements, I'll rebase the original branch and open a new pull request with them.

PR #305, opened 2019-10-27
FairLogger python interface (version 2)

olantwin:
Quote @JLTastet:

I have added a `delim` keyword argument to the Python wrappers. It did not require modifying the C++ code at all, and its behaviour now matches that of the `print()` function.

**Typical usage:**
```python
>>> from logger import *
>>> error(1, 2, "Test", 3.14159)
# [ERROR] 1 2 Test 3.14159
>>> error(1, 2, "Test", 3.14159, delim='')
# [ERROR] 12Test3.14159
>>> error(1, 2, "Test", 3.14159, delim=' | ')
# [ERROR] 1 | 2 | Test | 3.14159
```
```
>>> help(error)
Help on function function in module logger:

function(*args, **kwargs)
    Python wrapper around the C++ function template ship::error(Types... args).
    
    Keyword arguments:
    delim -- delimiter to be inserted between each argument (default ' ')
```

_Originally posted by @JLTastet in https://github.com/ShipSoft/FairShip/pull/303#issuecomment-546626450_

### DIFF
--- a/python/logger.py
+++ b/python/logger.py
@@ -5,11 +5,25 @@ import ROOT
 # TODO: add header to CMake
 ROOT.gROOT.ProcessLine('#include "' + os.path.join(os.path.expandvars('$FAIRSHIP'), 'utils', 'logger.hxx') + '"')
 
-for func in ('fatal', 'error', 'warn', 'info', 'debug'):
-    # The double lambda is used in order to capture 'func' by value, not by reference
-    # (which would always point to the last value taken in the loop)
-    globals()[func] = (lambda f:
-        lambda *args: ROOT.ship.__dict__[f](*map(type, args))(*args)
-    )(func)
-    globals()[func].__doc__ = \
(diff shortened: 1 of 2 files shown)


olantwin:
The only thing I noticed while testing is that we don't get a second `[SEVERITY]` when using `delim='\n'`. Maybe we can add a special case for that at some point ;)

JLTastet:
> The only thing I noticed while testing is that we don't get a second `[SEVERITY]` when using `delim='\n'`. Maybe we can add a special case for that at some point ;)

This seems to be the FairLogger behaviour. What you propose might be added by breaking argument lists on newline characters and calling the C++ function several times (i.e. instantiate a different logger for each line).

It could be tricky though, since some non-string objects might have a representation containing newlines. So unless we do all the string processing on the Python side, we will miss those.

I would rather have the user do the work of calling the logging function for each line.

olantwin:
Completely agree. Maybe a future improvement, but I think this is good enough for all reasonable use cases as is :).

PR #306, opened 2019-10-28
WIP: makeGenieEvents.py

dcentanni:
@antonioiuliano2 and I updated the splines, I then added two new arguments: -n and -e. The first allows you to set the number of events to request to Genie, the latter allows you to set the event type of the interacting neutrinos. I set a comment that reminds you to export the environment variable GXMLPATH in order to disable Genie decays for charm particles and tau. And finally I added two last lines that call the functions to start the simulation in an autonomous manner. 

We're waiting to produce new splines before updating the input paths to the SHiP data folder.

### DIFF
--- a/macro/makeGenieEvents.py
+++ b/macro/makeGenieEvents.py
@@ -7,20 +7,20 @@ import shipRoot_conf
 import argparse
 shipRoot_conf.configure()
 
-os.system("export GXMLPATH=/eos/user/d/dcentann")
+os.system("export GXMLPATH=/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines")
 os.system("echo exporting executed")
 
 # 
-xsec = "nuSHiP.xml"
+xsec = "gxspl-FNAL-nuSHiP-minimal.xml"# new adapted splines from Genie site
 hfile = "pythia8_Geant4_1.0_withCharm_nu.root" #2018 background generation
 #xsec = "Nu_splines.xml"
 #hfile = "pythia8_Geant4-withCharm_onlyNeutrinos.root"
 #put the above files in folder linked by the environment variables GENIEXSECPATH and NUFLUXPATH
 
-splines   = '/eos/user/a/aiuliano/public/genie_input/DefaultPlusValenciaMEC/'+'/'+xsec #path of splines
-neutrinos = '/eos/user/a/aiuliano/public/genie_input'+'/'+hfile #path of flux
+splines   = '/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines'+'/'+xsec #path of splines
+neutrinos = '/eos/experiment/ship/data/Mbias/background-prod-2018'+'/'+hfile #path of flux
 
-#Possbile evtypes: CC, CCDIS, CCQE,CharmCCDIS,RES,CCRES
+#Possbile evtypes: CC, CCDIS, CCQE, CharmCCDIS, RES, CCRES
 #See other evtypes in $GENIE/config/EventGeneratorListAssembler.xml
 evtype = 'CCDIS'
 
--- a/macro/makeGenieEvents.py
+++ b/macro/makeGenieEvents.py
@@ -12,9 +12,8 @@ shipRoot_conf.configure()
 # Before runnig this script please run this command in FairShip bash:
 # export GXMLPATH='/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines'
 # this will disable Genie decays for charm particles and tau
-if os.environ.has_key('GXMLPATH') is False:
+if 'GXMLPATH' not in os.environ:
 	logging.warn('GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour')
-	logging.debug('GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour')
 else: logging.debug('GXMLPATH is set: Genie will not decay charm and tau particles')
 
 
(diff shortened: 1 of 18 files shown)


olantwin:
Maybe it would make sense to print a warning if `GXMLPATH` is not set?

ThomasRuf:
Will this command work for people not working on lxplus? Does it use xrootd or does it require access to a path /eos/experiment/ship/... ?

dcentanni:
> Maybe it would make sense to print a warning if `GXMLPATH` is not set?

Yes, I'll write it as soon as possible

dcentanni:
> Will this command work for people not working on lxplus? Does it use xrootd or does it require access to a path /eos/experiment/ship/... ?

I'm afraid it won't work out of lxplus.. It requires the access to the path you mentioned. I'll figure out how to make it possible.

ThomasRuf:
then i would propose another way. If the file does not exist in a to be defined local directory, then you copy it from eos using the xroot protocol.

olantwin:
If you have access to the file via the xroot protocol, you should be able to run this without having to copy it, as long as the filename is correctly specified in the script, i.e. `root://eospublic.cern.ch//eos/experiment/ship/...`.

No need to copy potentially large files to have them locally.

olantwin:
Even on lxplus, it's better to use xroot to deal with large files than the EOS FUSE mount.

ThomasRuf:
it depends who is opening the file. Since it is a xml file, I doubt it is opened via ROOT. I guess, it is GENIE opening the file.

olantwin:
ROOT can be used to open any file using xroot. One just needs to add a suffix to the filepath (I think `?raw`), to tell it that it's not a ROOT file.

Then it can be used as a normal file.

As GENIE is built on ROOT, it might have support for xroot built in as well.

olantwin:
Ah, I see. The question is essentially whether gevgen supports xroot uris instead of local filenames. Should be easy to test!

macro/makeGenieEvents.py
```diff
@@ -11,7 +11,8 @@
 # Before runnig this script please run this command in FairShip bash:
 # export GXMLPATH='/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines'
 # this will disable Genie decays for charm particles and tau
-
+if os.environ.has_key('GXMLPATH'): print('exporting GXMLPATH executed')
```
olantwin: We should probably not print anything if this is correct (or print a debug message, which would be silent by default unless the verbosity is increased).

macro/makeGenieEvents.py
```diff
@@ -11,7 +11,8 @@
...
+else: print('WARNING: exporting GXMLPATH not executed')
```
olantwin: This would be a good opportunity to use the new FairLogger interface (you'll need to import `logger.warn`:
`warn("GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour")

dcentanni: Do you mean just adding the line you wrote in the if statement  ? Anyway I'll rewrite those line in the way you suggested.

olantwin: Yes, but replacing the print statement.

For the debug one you can use `logger.debug` in a similar way.

`[WARN]` or `[DEBUG]` will be automatically printed in front of the message and the output can be centrally controlled similar to the log messages we have in C++ (e.g DEBUG messages are only printed when the verbosity is changed).

dcentanni: Oliver, I changed the file, please warn me if I made a mistake using logging functions.

macro/makeGenieEvents.py
```diff
@@ -5,14 +5,17 @@
...
 # IMPORTANT
 # Before runnig this script please run this command in FairShip bash:
 # export GXMLPATH='/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines'
 # this will disable Genie decays for charm particles and tau
-if os.environ.has_key('GXMLPATH'): print('exporting GXMLPATH executed')
-else: print('WARNING: exporting GXMLPATH not executed')
+if os.environ.has_key('GXMLPATH') is False:
+	logging.warn('GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour')
+	logging.debug('GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour')
+else: logging.debug('GXMLPATH is set: Genie will not decay charm and tau particles')
```
olantwin: In the case where it's not set, you only need the warning, not the debug message as well.

May I suggest using the slightly more pythonic condition:
```python3
if 'GXMLPATH' not in os.environ:
    logger.warn('...')
else:
    logger.debug('...')
```

ThomasRuf:
If I am not mistaken, this script is also used to generate neutrino background for HNL. For this use case, "Genie will decay charm and tau particles" is the desired behaviour. Therefore find warning message confusing.

instead of hardcoding the values for splines and neutrinos, it maybe better to have these as arguments with recent default values. 

Is GENIE now able to open files on EOS using xrootD?

olantwin:
@antonioiuliano2 Could you elaborate on the purpose of the script?
As far as I'm aware it's used both for neutrino physics and background studies, but I might be mistaken.

Would agree that it's better to not hard-code the splines and neutrinos but using command line arguments with sensible defaults.

ThomasRuf:
for the use case of background studies, I think "Genie will decay charm and tau particles" is the desired behaviour. 
for the use case of tau neutrino physics, it is not the desired behaviour. Therefore, the message logging.warn('GXMLPATH is not set: Genie will decay charm and tau particles, which is usually not the desired behaviour') is misleading for the first use case.

antonioiuliano2:
Dear all, it is as Thomas said, so we can maybe add the env variable check
for a --nudet option, which sets the splines and does this check.

I do not believe that the current makeGenie version is compatible with
rootrc. However, we can check that.

Best regards
Antonio

Il mar 29 ott 2019, 11:00 Thomas Ruf <notifications@github.com> ha scritto:

ThomasRuf:
any news? should I merge it?

dcentanni:
> any news? should I merge it?

Hi Thomas, there are no news. If there are no necessary updates to the file, you can merge it.

PR #307, opened 2019-10-29
Fix python2/3 scripts problems in python folder

shir994:


ThomasRuf:
i had no problems running with python3. I am therefore surprised about these changes. I was using `source /cvmfs/ship.cern.ch/SHiP-2020/2019/September/12/setUp.sh`

shir994:
Ok, but if you look at this script, in set PATH to python2.7 interpreter. So that's may be the reason you did not have problems.

I have rebuild everything from scratch on Ubuntu 16.04 LTS docker image. I have nothing installed upfront. So it was totally controlled environment. I believed, if you try to explicitly remove any python2.7 libs from PATH/PYTHONPATH, you will have my issue. Also, you can check unmerged PR from Simone @siscia , you will see he has fixed it as well.

PR #311, opened 2019-10-31
WIP: Adding instructions for local build

antonioiuliano2:
Dear all,

I have noticed that the current version of README does not contains information about how to build locally the FairShip software.

I have added the essential commands, but I would require your assistance on two points:

1) It is usually recommended to use only 1 core for installation, because compiling ROOT often leads to crash for lack of memory. Shall we modify aliBuild.sh accordingly, or leave it to the user?

2) To help with the first phase, shall we insert a list of pre-requisites? I know that aliDoctor tells the user which packages he needs to install, but it might be helpful to have an initial list of packages to install before launching aliBuild, similarly as it is done by the ALICE group:

https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-ubuntu.html

Best regards,
Antonio

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,21 @@ However, this won't work if you are using HTCondor. In such case you can do:
 eval alienv load FairShip/latest
 ```
 
+## Local build, without access to CVMFS
+Commands are similar to the previous case, but without access to CVMFS you need to build the required packages.
+1. Download the FairShip software
+    ```bash
+    git clone https://github.com/ShipSoft/FairShip.git
+    ```
+2. Build the software using aliBuild
+    ```bash
+    FairShip/aliBuild.sh
+    ```
+3. Load the environment
+    ```bash
+    alibuild/alienv enter FairShip/latest
+    ```
+    
 ## Docker Instructions
 1. Build an docker image from a Dockerfile:
     ```bash
--- a/system/install_on_ubuntu.sh
+++ b/system/install_on_ubuntu.sh
@@ -1,15 +1,15 @@
-# installation on blank Ubuntu 17.10: 
-# FROM ubuntu:17.10
+# installation on blank Ubuntu 18.04 LTS:
+# FROM ubuntu:18.04
 sudo apt-get update 
-sudo apt-get install python-dev python-pip make patch sed git-all gedit environment-modules \
+sudo apt-get install python3-dev python3-pip make patch sed git-all gedit environment-modules \
    libx11-dev libxpm-dev libxext-dev ncurses-dev libxml2-dev libxft-dev \
    libxmu-dev  ncurses-dev  curl bzip2  libbz2-dev gzip unzip tar gfortran libkrb5-dev \
     wget automake autoconf libtool bison  flex byacc libgif-dev libjpeg-dev libtiff5-dev\
    libexpat1-dev libcurl4-openssl-dev libssl-dev libbz2-dev libbz2-dev libglu1-mesa libglu1-mesa-dev\
    autopoint texinfo gettext libtool libtool-bin pkg-config python-tk cmake linuxbrew-wrapper 
 
-sudo pip install --upgrade pip
-sudo pip install matplotlib numpy scipy certifi ipython ipywidgets ipykernel notebook metakernel pyyaml
+sudo pip3 install --upgrade pip
+sudo pip3 install matplotlib numpy scipy certifi ipython ipywidgets ipykernel notebook metakernel pyyaml sklearn
 
 # sudo apt-get krb5-user
 # for kinit: cp from lxplus /etc/krb5.conf


owtscharenko:
I use to build FairSHiP on my Ubuntu lab pc, so maybe I can give some insight to this:

I think the memory issue for ROOT building on several cores is not really a problem, for me it used never more than approx. 4gb of memory

For the second point I completely agree, aliDoctor was not helpful at all for me, and I had to find all the missing packages by hand during the fist installation.

olantwin:
> I have added the essential commands, but I would require your assistance on two points:
> 
> 1. It is usually recommended to use only 1 core for installation, because compiling ROOT often leads to crash for lack of memory. Shall we modify aliBuild.sh accordingly, or leave it to the user?

I would actually suggest that we retire aliBuild.sh and recommend using aliBuild directly.
The commands are quite simple, and this way it's more transparent what happens (and how to fix things if something goes wrong!). And now that we're using aliBuild directly on lxplus, it's also easier to remember for users.

We can add a remark about the ROOT issue (it's an issue during linking).
@owtscharenko does your PC have swap? That could explain why you don't have a crash.

> 2. To help with the first phase, shall we insert a list of pre-requisites? I know that aliDoctor tells the user which packages he needs to install, but it might be helpful to have an initial list of packages to install before launching aliBuild, similarly as it is done by the ALICE group:
> 
> 
> https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-ubuntu.html

We used to have the depenencies in the readme, I think. Maybe check in its history?

We should probably update them though and list them for the most common distributions (CC7, Ubuntu,, Ubuntu LTS, Fedora).

owtscharenko:
> > I have added the essential commands, but I would require your assistance on two points:
> > 
> > 1. It is usually recommended to use only 1 core for installation, because compiling ROOT often leads to crash for lack of memory. Shall we modify aliBuild.sh accordingly, or leave it to the user?
> 
> I would actually suggest that we retire aliBuild.sh and recommend using aliBuild directly.
> The commands are quite simple, and this way it's more transparent what happens (and how to fix things if something goes wrong!). And now that we're using aliBuild directly on lxplus, it's also easier to remember for users.
> 
> We can add a remark about the ROOT issue (it's an issue during linking).
> @owtscharenko does your PC have swap? That could explain why you don't have a crash.
> 
Yes I have a lot of memory (32gb) and the same size of swap partition. But I do not remember that this was maxed at any point of the installation procedure.
> > 1. To help with the first phase, shall we insert a list of pre-requisites? I know that aliDoctor tells the user which packages he needs to install, but it might be helpful to have an initial list of packages to install before launching aliBuild, similarly as it is done by the ALICE group:
> > 
> > https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-ubuntu.html
> 
> We used to have the depenencies in the readme, I think. Maybe check in its history?
> 
> We should probably update them though and list them for the most common distributions (CC7, Ubuntu,, Ubuntu LTS, Fedora).

antonioiuliano2:
> > I have added the essential commands, but I would require your assistance on two points:
> > 
> > 1. It is usually recommended to use only 1 core for installation, because compiling ROOT often leads to crash for lack of memory. Shall we modify aliBuild.sh accordingly, or leave it to the user?
> 
> I would actually suggest that we retire aliBuild.sh and recommend using aliBuild directly.
> The commands are quite simple, and this way it's more transparent what happens (and how to fix things if something goes wrong!). And now that we're using aliBuild directly on lxplus, it's also easier to remember for users.
Personally I would like to keep aliBuild.sh, because it automatically clones all the required repositories (alibuild, shipdist...) and sets the environment. This helps a lot in the first time installation. This does not forbid to use aliBuild directly, or changing the aliBuild command in aliBuild.sh, if it is needed. But having the commands in a script makes the installation procedure clearer to launch and to understand.
> 
> We can add a remark about the ROOT issue (it's an issue during linking).
> @owtscharenko does your PC have swap? That could explain why you don't have a crash.
> 
> > 1. To help with the first phase, shall we insert a list of pre-requisites? I know that aliDoctor tells the user which packages he needs to install, but it might be helpful to have an initial list of packages to install before launching aliBuild, similarly as it is done by the ALICE group:
> > 
> > https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-ubuntu.html
> 
> We used to have the depenencies in the readme, I think. Maybe check in its history?
Found them. They are still in the master version, in the system folder: 
https://github.com/ShipSoft/FairShip/blob/master/system/install_on_ubuntu.sh
Commit is dated 12 December 2017, however, so they need to be checked and updated.
> 
> We should probably update them though and list them for the most common distributions (CC7, Ubuntu,, Ubuntu LTS, Fedora).

ThomasRuf:
I second Antonio. I think for many non-expert users it would be good to have just one simple command. I myself, if I have not worked for some time with alibuild, I forget the syntax, specially the one needed for lxplus, and have to find it somewhere. So, I place everything in a private script.

For the dependencies, isn’t this checked by Alibuild? Maybe some dependencies are missing and need to be added.

Cheers,
Thomas

Von: Antonio <notifications@github.com>
Gesendet: Donnerstag, 31. Oktober 2019 17:21
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [ShipSoft/FairShip] WIP: Adding instructions for local build (#311)


I have added the essential commands, but I would require your assistance on two points:

  1.  It is usually recommended to use only 1 core for installation, because compiling ROOT often leads to crash for lack of memory. Shall we modify aliBuild.sh accordingly, or leave it to the user?

I would actually suggest that we retire aliBuild.sh and recommend using aliBuild directly.
The commands are quite simple, and this way it's more transparent what happens (and how to fix things if something goes wrong!). And now that we're using aliBuild directly on lxplus, it's also easier to remember for users.
Personally I would like to keep aliBuild.sh, because it automatically clones all the required repositories (alibuild, shipdist...) and sets the environment. This helps a lot in the first time installation. This does not forbid to use aliBuild directly, or changing the aliBuild command in aliBuild.sh, if it is needed. But having the commands in a script makes the installation procedure clearer to launch and to understand.

We can add a remark about the ROOT issue (it's an issue during linking).
@owtscharenko<https://github.com/owtscharenko> does your PC have swap? That could explain why you don't have a crash.

  1.  To help with the first phase, shall we insert a list of pre-requisites? I know that aliDoctor tells the user which packages he needs to install, but it might be helpful to have an initial list of packages to install before launching aliBuild, similarly as it is done by the ALICE group:

https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-ubuntu.html

We used to have the depenencies in the readme, I think. Maybe check in its history?
Found them. They are still in the master version, in the system folder:
https://github.com/ShipSoft/FairShip/blob/master/system/install_on_ubuntu.sh
Commit is dated 12 December 2017, however, so they need to be checked and updated.

We should probably update them though and list them for the most common distributions (CC7, Ubuntu,, Ubuntu LTS, Fedora).

olantwin:
Local installations without CVMFS is not something I would recommend to the average user.

Even with aliBuild.sh, it will most likely not work out of the box (maybe on CC7, but even there I regularly run into issues). So the expectation should be that some troubleshooting and basic understanding of aliBuild will be needed.

I find it very useful to have a local installation, so we should definitely document as well as possible how to do it. But my fear is, that offering "just run this script", users will get the impression that this will just magically work.

ThomasRuf:
I don't see why it would not be possible to have one script which does the installation from scratch. It is the standard procedure of all comercial products, download and run install script.

olantwin:
If we want it to reliably work from scratch, we'd have to bundle or download all the pre-built dependencies and just install them (aliBuild has a feature that can be used for that, but I haven't had time to play around with it). That's what happens for most commercial software on Linux (and Windows).

If we want to build it, we have such a script: aliBuild. We could fork it (e.g. to ShipBuild) to look for shipdist by default (and we should rename/symlink `defaults_fairship.sh`→`defaults_release.sh`), such that it becomes just a single invocation by default:
```
ShipBuild build FairShip; ShipEnv enter FairShip/latest
```

But I don't think a wrapper adds much in addition and adds a layer of opacity.

olantwin:
BTW, the ALICE prerequisites should also be valid for us:

<https://alice-doc.github.io/alice-analysis-tutorial/building/custom.html#prerequisites>

We should definitely add a link to them to this section for now, to be replaced with our own, once we have a list of our own.
Maybe only to the CC7 and Ubuntu pages, as local builds on MacOS and Fedora are a bit broken right now...

ThomasRuf:
apparently, there is now a conflict. Probably should have merged this before (#314)

olantwin:
I think the conflict  is due to #312. Trivial to fix though.

antonioiuliano2:
Hi all,

sorry for my late reply, but I was going back to my home in Naples. I have fixed the conflict

antonioiuliano2:
Dear all,

I have just installed a fresh Ubuntu 18.04.3 LTS on my laptop, then I have started testing the prerequisites. They seem mostly fine until  now, except that they refer to pip, instead after the transition to python3 shipdist packages refer to pip3 (look for example at python-modules prefer system check). 
Therefore, I simply changed from python-dev and python-pip to python3-dev and python3-pip. I have also added sklearn, which is required by python-modules prefer system check.

antonioiuliano2:
Excuse me, I have another question. Since it is connected to the test build I am doing I will write it here:
Ubuntu 18.04 LTS still uses python2 as default python, instead of python3 (I do not know the reason).
Would you recommend that I use python2 from the system, or build python3 from alibuild package?

olantwin:
> Excuse me, I have another question. Since it is connected to the test build I am doing I will write it here:
> Ubuntu 18.04 LTS still uses python2 as default python, instead of python3 (I do not know the reason).
> Would you recommend that I use python2 from the system, or build python3 from alibuild package?

We'd need to check which packages and scripts use `python` unqualified and make sure that they use `python2` and `python3` where appropriate, I think.
We can override the system-checks based on the `defaults_*.sh` file, so we can ensure there, that the right python versions are used.

antonioiuliano2:
Both ROOT and LHAPDF5 use /usr/bin/python, so Python2 by default.
Moreover, LHAPDF5 cannot be built with Python3, since it (ironically) says python version is too old due to this check in the configure script (line 21470 of https://github.com/ShipSoft/LHAPDF/blob/alice/v5.9.1/configure):

minver = map(int, string.split('2.3', '.')) + [0, 0, 0]

which is not python3 compatible.

ThomasRuf:
On my Ubuntu 18.04, I do have

python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
I would always recommend to use the system version.

For ROOT, you need a special ROOT python3 version. The normal ROOT version will not work.

Concerning issues with LHAPDF, how did Simone managed to use python3 on lxplus? I tried his latest python3 release, /cvmfs/ship.cern.ch/SHiP-2020/2019/September/12/ some time ago, worked without problems.

olantwin:
For ROOT we will need to modify the compile options. I think it can pick the python to use based on an environment variable...

If not, we'll need to override the recipe in the defaults.

As to which python3 version to use, I would use the system version iff it's ≥3.6.

PR #314, opened 2019-11-01
Add a remark on the branches and python versions

olantwin:
This should hopefully clarify which python version should be used with which branch and aliBuild default right now.

Maybe we can also add a remark on which setup script should work with each on `lxplus`.


ThomasRuf:
i wouldn't call muflux a frozen branch, i am almost changing it daily.

olantwin:
Good point.

PR #315, opened 2019-11-05
Add remark that `fairship` default needed for master

olantwin:


olantwin:
In reaction to the maiing list discussion.

Issue #316, opened 2019-11-14
Exceptions thrown in branch muflux, class MufluxReco


StBies:
# Description of bug
In the class MufluxReco (package charmdet), the method MufluxReco::extrapolateToPlane(genfit::Track* fT,Float_t z, TVector3& pos, TVector3& mom) has a few lines of code where unhandled exceptions can be thrown. For me this happened at least once in RUN_8000_2200, and resulted in a crash.

## Lines that can throw exceptions

https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/MufluxReco.cxx#L485

https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/MufluxReco.cxx#L492

https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/MufluxReco.cxx#L506

https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/MufluxReco.cxx#L514

## Type of exception?
The exceptions thrown are of type genfit::Exception (https://github.com/ShipSoft/FairShip/blob/muflux/genfit/core/include/Exception.h)

# Where method is called
The method throwing the unhandled exception is called in drifttubeMonitoring.py, line 3094 
https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/drifttubeMonitoring.py#L3094
 and its result is saved in a variable called tracklength (that is never used from there on).

# Possible way to handle this
In drifttubeMonitoring.py, the same method from c++ is called and there, the exception is handled:
https://github.com/ShipSoft/FairShip/blob/126ff278056319178e12f30b084ef1ea991fd605/charmdet/drifttubeMonitoring.py#L1772

ThomasRuf:
I am also suffering from this problem. But what is the solution? I am not a C++ expert, no idea how to handle exceptions.

StBies:
The way in C++ is to put the code in a block like
```c++
try
{
        auto st  = fT->getFittedState(m);
        auto Pos = st.getPos();
}
catch(const genfit::Exception& e)
{
        //do whatever should happen when getFittedState fails
}


If you can tell me, what should happen, when getFittedState fails I can implement that for you.

[...]

PR #318, opened 2019-11-15
Catching a genfit::Exception in track extrapolation

StBies:
# Related issue
This PR is related to issue #316.

# Applied changes
In this PR three tings are changed: 

1. In all cases where an unhandled genfit::Exception can be thrown it is now handled. This could in general happen in four cases. The first three are very unlikely (in fact should only happen if the fit didn't converge which should be realized before extrapolating a passed track) and should not happen. The last case is only handled differently where the cardinal track representation is extrapolated in case of an exception.
2. Two variables (both TVector3) were declared as global heap veriables which might be a security risk and might lead to unforeseeable behavior of the program. Made them private member variables of the class MufluxReco. This way, they don't need to be re-allocated everytime the method ```extrapolateToPlane(Double_t MufluxReco::extrapolateToPlane(genfit::Track* fT, Float_t z, TVector3& pos, TVector3& mom)``` is called and they are now encapsulated in the ```MufluxReco namespace```.
3. Probably most important change: Removed an ```if``` check that was labelled to be there because it is more stable against crashes. Removing this leads to residuals changing in the sub micron range but has a performance increase (for my tests between 5% and 20%). Need to discuss if we want to keep this. For more info see the comment in the issue: https://github.com/ShipSoft/FairShip/issues/316#issuecomment-554358758

### DIFF
--- a/charmdet/MufluxReco.cxx
+++ b/charmdet/MufluxReco.cxx
@@ -20,15 +20,14 @@
 #include <algorithm>
 #include <vector>
 
-TVector3* parallelToZ = new TVector3(0., 0., 1.);
-TVector3* NewPosition = new TVector3(0., 0., 0.);
 std::vector<int> charmExtern = {4332,4232,4132,4232,4122,431,411,421};
 std::vector<int> beautyExtern = {5332,5232,5132,5232,5122,531,511,521};
 std::vector<int> muSources   = {221,223,333,113,331};
 // -----   Standard constructor   ------------------------------------------ 
-MufluxReco::MufluxReco() {}
+MufluxReco::MufluxReco(): m_new_position(0., 0., 0.), m_parallelToZ(0., 0., 1.) {}
 // -----   Standard constructor   ------------------------------------------ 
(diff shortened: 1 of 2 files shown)


ThomasRuf:
looks good ! Only one point:
auto point = fT->getPointWithMeasurementAndFitterInfo(1, CRep); 
i think this should be
auto point = fT->getPointWithMeasurementAndFitterInfo(last_hit_id, CRep);

could you please test?

thanks Thomas

StBies:
> looks good ! Only one point:
> auto point = fT->getPointWithMeasurementAndFitterInfo(1, CRep);
> i think this should be
> auto point = fT->getPointWithMeasurementAndFitterInfo(last_hit_id, CRep);
> 
> could you please test?
> 
> thanks Thomas

Right, that would make more sense. I added this change in https://github.com/ShipSoft/FairShip/pull/318/commits/01c051025fc75a2310d6bf9fa59b831360e4c636 and tested with one spill both for magnet off and magnet on. Compared to before there is no difference in the residuals (since this should (if ever) happen very rarely but at least now the code makes more sense.

PR #321, opened 2019-12-18
Addition of an Upstream Veto Tagger based on MRPC technology

CelsoFCF:


antonioiuliano2:
Dear Celso,

thank you very much for all the work and additions you have made to the geometry.

It seems fine by me. However I cannot address how to preserve backward compatibility, since the passage of methods from NuTauMuDet class to the new UpstreamTagger class inevitably breaks it (at least according to my tests).

Thomas, can I please ask you how  would you suggest to address it? Can I ask you how do you treat cases like this, now that you have introduced official FairShip branches?

Best regards,
Antonio

ThomasRuf:
If it is decided to go ahead with Upstream Tagger as the baseline without any backup option, there is no need to preserve backward compatibility, in my mind. However, it should be checked that any reconstruction/analysis code using NuTauMuDet is updated or the authors informed how to update it. Siutation becomes more complicated if two detector options should be pursued.

antonioiuliano2:
Dear Thomas,

I agree with your point. However, even in this version, the NuTauMudet class is not abandoned as a detector, since it is still used for the RPCs of the Muon Filter. It simply lost the last volumes, which became part of the new Upstream Tagger class.

Therefore, both classes store hit points (ShipRpcPoints and UpstreamTaggerPoints), which then should be included in the analysis. They are not actually two independent detector options.

Best regards,
Antonio

ThomasRuf:
OK. But then I might have a problem of understanding your concern about backward compatibility. Is it more an issue of co-existence? The digitized hits of the NuTauMudet and the new Upstream Tagger will be used in a to-be-updated veto-tagging algorithm, correct?

thomas

antonioiuliano2:
You are right, 
I should have explained it before, I am sorry. 
I am referring to the checks you requested in the past before closing a pull request: doing a simulation with current master branch, than try to launch ShipReco and ShipAna with the new one. In this case it fails with the following error:

****
antonio@antonio-X555LJ:~/Simulations/sim_master/backwardcompatibilitycheck$ python -i ~/SHiPBuild/FairShip/macro/ShipAna.py -f ../ship.conical.PG_13-TGeant4_rec.root -g ../geofile_full.conical.PG_13-TGeant4.root 
[INFO] Media file used: /home/antonio/SHiPBuild/sw/ubuntu1804_x86-64/FairShip/master-1/geometry/media.geo
 Mag Spectro Design 3
Traceback (most recent call last):
  File "/home/antonio/SHiPBuild/FairShip/macro/ShipAna.py", line 68, in <module>
    modules = shipDet_conf.configure(run,ShipGeo)
  File "/home/antonio/SHiPBuild/sw/ubuntu1804_x86-64/FairShip/master-1/python/shipDet_conf.py", line 220, in configure
    taumuondetector.SetRpcDimDifferences(ship_geo.tauMudet.deltax, ship_geo.tauMudet.deltay)
AttributeError: 'NuTauMudet' object has no attribute 'SetRpcDimDifferences'
>>> exit()
make suicid, until better solution found to ROOT/genfit interference
Terminato
****

This happens because the method SetRpcDimDifference is not present anymore in NuTauMuDet class, since it was used to set the size difference between the RPCs used for filtering and for veto taggings (the latter ones are naturally larger).

I also confirm what you said: our goal is to use digitized hit for veto tagging and reconstruction.

Best regards,
Antonio

ThomasRuf:
Not sure I understand correctly. For the master branch, without backward compatibility, so no requirement reading old data with new software. The only requirement, with the changes applied, simulation, reconstruction and analysis step should not produce an error.

antonioiuliano2:
Now I understand,

There is no problem then, I was just referring to the old configuration (before the introduction of official branches).

Thank you very much for your explanation, I am sorry for the misunderstanding.

Best regards,
Antonio

ThomasRuf:
Can we go ahead and merge this pull request?

antonioiuliano2:
Dear Thomas,

as far as I remember, there was no other issue to be solved, so you should be able to merge it. 
Let us just wait for Celso's confirmation, if possible.

Thank you for your assistance and 
Best regards,
Antonio

CelsoFCF:
Dear all,

I confirm that the pull request can be merged as it is.

Best regards,

Celso

Em 2020-03-27 07:31, Antonio escreveu:

olantwin:
This seems to have broken master (see issue #334)

CelsoFCF:
Dear Oliver,

I should be able to solve this issue on Monday. I guess that some 
detectors positions were slightly changed in last couple of months.

Best regards,

Celso

stEm 2020-03-28 18:01, Oliver Lantwin escreveu:

ThomasRuf:
The only issue I see when running the reconstruction over newly simulated data:

Error in <TStreamerInfo::Build>: UpstreamTaggerHit, discarding: const double* mom, no [dimension]

CelsoFCF:
Dear Oliver,

I submitted a new pull request to solve the issues that you reported. 
The class UpstreamTaggerHit.cxx was not completely up to date (I just 
updated this function). Concerning the error reported by Thomas I must 
say that it has no impact in the reconstruction (still, if one of you 
knows how to make it disapear I would appreciate it).

The code should be fully functional now. Nevertheless, I would like to 
report that I have the z position of UpstreamTagger hardcoded in the 
class UpstreamTaggerHit.cxx (line 179). This may be a problem in case of 
change in the position of the muon detector. Therefore, it would be 
desirable to obtain the position of the UpstreamTagger detector from 
GeoManager (in line 179). I was not able to do it. Can someone of you 
help in this?

Best regards,

Celso

Em 2020-03-28 18:01, Oliver Lantwin escreveu:

ThomasRuf:
you cannot get the position reliably from the geomanager, since it is not for sure that the muon detector is already being created when your code is called. Therefore, the positioning along z-axis is done in shipDet_conf.py during the configuration phase using the variables provided by each subdetector, some start z and length of element.

PR #322, opened 2020-01-07
pixel modules slice option and passive material added

matclim:
An option for slicing the pixel modules has been added, it's implemented by setting c.PixelModules.nSlice= 10. The change requires rebuilding FairShip for some reason however.

Passive material for the pixel modules has also been added.


ThomasRuf:
Please, only commit the files which you really want to commit. I hope this are not the 92 in list above.

PR #323, opened 2020-01-07
pixel modules slice option and passive material added

matclim:
An option for slicing the pixel modules has been added, it's implemented by setting c.PixelModules.nSlice= 10. The change requires rebuilding FairShip for some reason however.

Passive material for the pixel modules has also been added.


olantwin:
@matclim What do you mean by it has to be rebuilt?

It should normally not be necessary to rebuild to make changes in the geometry config take effect.

matclim:
This is a mystery for me as well but if it isn't rebuilt, the change in the slicing doesn't impact anything.

ThomasRuf:
If you don't rebuild, the modifed file will stay in the source directory, while the program expects it from $FAIRSHIP, which is the build directory.

PR #327, opened 2020-02-17
Fix pixel positions

owtscharenko:
This pull request corrects for wrongly placed pixel detector volumes in the charm simulation.
It also updates the slicing option to be able to choose an arbitrary number of slices per sensor volume. 
The simulation runs and the output files seem fine.

However the errors which triggered my investigation were not fixed. After simulation following output is created:

```
 --------  End PYTHIA Event Listing  -----------------------------------------------------------------------------------------------
Error in <TStreamerInfo::Build>: PixelModules, discarding: int* PixelIDlist, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* xs, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* ys, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* zs, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* xangle, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* yangle, no [dimension]

Error in <TStreamerInfo::Build>: PixelModules, discarding: double* zangle, no [dimension]

[INFO] ***  FairBaseParSet written to ROOT file   version: 1
[INFO] ***  FairGeoParSet written to ROOT file   version: 1
```

I believe this is thrown when Geant4 tries to store these variables, however they are only used for placement of the volumes, and seem not to be important afterwards ? If anyone has a clue how to prevent this I would be happy to learn.

### DIFF
--- a/charmdet/PixelModules.cxx
+++ b/charmdet/PixelModules.cxx
@@ -124,42 +124,25 @@ void PixelModules::SetSiliconStationAngles(Int_t nstation, Double_t anglex, Doub
  zangle[nstation] = anglez;
 }
 
-void PixelModules::SetSiliconSlicesNumber(Int_t nSl){
-	nSlices=nSl;
-        if(nSlices==1){
-		nSi=nSi1;
-	}
-	else if(nSlices=10){
-	nSi=nSi10;
-	}
+void PixelModules::SetSiliconSlicesNumber(Int_t nSl)
+{
+   nSlices = nSl;
+   nSi = nSlices * 12;
 }
 
-Double_t *PixelModules::GetPosSize1(){
-	return new Double_t[nSi1];
+Double_t *PixelModules::GetPosSize(Int_t n)
+{
+   return new Double_t[n];
 }
 
-Double_t *PixelModules::GetPosSize10(){
-	return new Double_t[nSi10];
-}
-
-void PixelModules::SetPositionSize(){
-	if(nSlices==1){
-		xs=GetPosSize1();
-		ys=GetPosSize1();
-		zs=GetPosSize1();
-		xangle=GetPosSize1();
-		yangle=GetPosSize1();
-		zangle=GetPosSize1();
-	}
-	else if(nSlices==10){
-		xs=GetPosSize10();
-		ys=GetPosSize10();
-		zs=GetPosSize10();
-		xangle=GetPosSize10();
-		yangle=GetPosSize10();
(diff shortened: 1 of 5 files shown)


olantwin:
I think ROOT has trouble serialising those arrays. Could you try replacing them with vectors?

Alternatively, you can tell ROOT to ignore some objects while serialising. (I think there's a magic comment (`//!`) to instruct ROOT to ignore an object.

olantwin:
https://root.cern.ch/root/html/guides/users-guide/AddingaClass.html#adding-a-class-with-a-shared-library for more info concerning `//!`

owtscharenko:
Thank you for the info. I did as you recommended and replaced the arrays by vectors.

olantwin:
Changing to `std::vector` worked?

owtscharenko:
yes, the error is not thrown any more

owtscharenko:
I used autopep8 to automatically reformat the python code according to pep8 style guidelines. This of course introduced a lot of white space changes, but in return makes the code more readable.
All code style changes are in the first commit of this pull request. I did not touch the lines by hand, so there were no programmatical changes introduced.
For comparison it helps to activate the "hide whitespace changes" option while in the "files changed" view on github.
The following commits changed actual code.

ThomasRuf:
It would be good to have one commit for changes regarding the charm detector, and another commit for cosmetics. I have no idea how to split your request.

olantwin:
@owtscharenko I would suggest you do a `git rebase -i origin/master` and remove the cosmetic commit for now. Then you can create a new pull request with cosmetic changes, if you like.

In general I'm strongly in favour of automatic formatting to ensure the code is nice and readable, but agree with Thomas that it's not a good idea to include those changes in feature pull requests.

owtscharenko:
Yes I agree that it makes sense to have 2 pull requests for this.

olantwin:
Maybe we should discuss some strategy for automatic code formatting (at least of new code) at a software meeting soon? @ThomasRuf 

This would help a lot with consistency and in my experience also can catch subtle errors due to e.g. wrong indentation wrt if/for statements.

owtscharenko:
I removed the first commit but there remain a couple of white space changes in charmDet_conf.py
I will create a pull request for formatting after this is merged.

PR #331, opened 2020-03-17
Updates on DP

tugberk92:
This update includes;
* Application of VDM Form Factor,
* Correction of meson production BR of DP,
* DP production by separately from each meson mothers,
* Correction of quark-pair ratios of hadronic decay modes,
* Addition of new meson production mode.


tugberk92:
I was trying to change the following files only;
* python/dpProductionRates.py
* python/readDecayTable.py
* python/proton_bremsstrahlung.py
* python/pythia8darkphoton_conf.py
* macro/run_simScript.py

olantwin:
It seems you based your work on the muflux branch or accidentally merged the muflux branch into it at some point...

I would suggest using an interactive rebase against the current master and removing all commits that aren't by you (seems like this should leave ~10 commits by you). You might have to fix some merge conflicts in places, but that's going to be unavoidable any way you do it.

olantwin:
Oh, and definitely undo that last merge!

PR #332, opened 2020-03-17
DP Updates + ALPACA Implementation

tugberk92:
This commit includes;

1.DP updates;
  * Application of VDM Form Factor,
  * Correction of meson production BR of DP,
  * DP production by separately from each meson mothers,
  * Correction of quark-pair ratios of hadronic decay modes,
  * Addition of new meson production mode.

2. ALPACA Generator implementation;
  * ALPACA generator is implemented via shipdist,
  * a python script is used for creating and adjusting the events,
  * a ntuple generator is used to process the events into FairShip.


tugberk92:
both(DP changes+ALPACA) are tested, and they are working fine so far..

macro/run_simScript.py
```diff
@@ -115,6 +118,10 @@
 if options.genie:    simEngine = "Genie"
 if options.nuradio:  simEngine = "nuRadiography"
 if options.ntuple:   simEngine = "Ntuple"
+if options.alpaca:
+    options.inputFile= False
+    CaloDesign = "2"
```
olantwin: This behaviour might be very surprising to the user!

Would suggest removing it. For studies ALP→digamma one would of course have to enable split-cal (although I expect it would probably  become default sooner or later anyway?

Also, are you sure CaloDesign takes a string and not an int?

macro/run_simScript.py
```diff
@@ -291,6 +300,28 @@
 # P8gen.SetMom(500.*u.GeV)
 # P8gen.SetId(-211)
  primGen.AddGenerator(P8gen)
+ 
+if simEngine == "Alpaca":#Rmin and Rmax of Decay Volume is needed...
+  #primGen.SetTarget(ship_geo.target.z0, 0.)
```
olantwin: Would suggest removing unnecessary commented lines of code. They are at best confusing.

macro/run_simScript.py
```diff
@@ -291,6 +300,28 @@
...
+  SmearBeam  = 1*u.cm # finite beam size
+  print(SmearBeam,"run")
+  Lmin       = ((ship_geo.Chamber1.z - ship_geo.chambers.Tub1length) - ship_geo.target.z0)/100.
+  Lmax       = (ship_geo.TrackStation1.z - ship_geo.target.z0)/100.
+  inputFile  = makeAlpacaEvents.runEvents(options.theMass,options.theDPepsilon,options.nEvents,Lmin,Lmax,startZ,endZ,SmearBeam)
+  ut.checkFileExists(inputFile)
+  #run.CreateGeometryFile("%s/geofile_full.%s.root" % (options.outputDir, tag))
+  # save ShipGeo dictionary in geofile
+  #import saveBasicParameters
+  #saveBasicParameters.execute("%s/geofile_full.%s.root" % (options.outputDir, tag),ship_geo)
```
olantwin: Again, these should not be included.

python/dpProductionRates.py
```diff
@@ -54,47 +62,64 @@ def brMesonToMesonDP(mass,epsilon,mumPdg,dauPdg,doprint=False):
...
-    fac2 = ROOT.TMath.Sqrt((mMeson**2-mass**2+mDaughterMeson**2)**2 - 4*mMeson**2*mDaughterMeson**2)
-    fac3 = pow(mMeson**2-mass**2,3)
-    massfactor = fac1*fac2/fac3
-    if (mass<(mMeson-mDaughterMeson)): br = epsilon**2*mesonBRtoPhoton(mumPdg,doprint)*massfactor
+    #BR is fixed after the review
+    #print(mMeson, mDaughterMeson)
+    #fac1 = (mMeson**2-mass**2-mDaughterMeson**2)**2
+    #fac2 = ROOT.TMath.Sqrt((mMeson**2-mass**2+mDaughterMeson**2)**2 - 4*mMeson**2*mDaughterMeson**2)
+    #fac3 = pow(mMeson**2-mass**2,3)
+    #massfactor = fac1*fac2/fac3
```
olantwin: Remove the commented lines.

python/makeAlpacaEvents.py
```diff
@@ -0,0 +1,146 @@
...
+
+def AlpacaFormatting(s):
+    s=str(s)
+    if s.find('d')==-1:
+        s = numpy.format_float_scientific(float(s))
+        s= s.replace('e','d')
+        s= s.replace('+','')
+    return s
+
+#determine the path for online usage it can be /eos/../ship/data/.. BUT don't know what we shall do in local usage??
```
olantwin: Maybe use <root://eospublic.cern.ch//eos/....>, which works from anywhere and give an option to use a local file via `-f localfilename.root`?

python/makeAlpacaEvents.py
```diff
@@ -0,0 +1,146 @@
...
+def Decayweight(Lmin,Lmax,Decaylength,LS):
+    return math.exp(-LS/Decaylength)*((Lmax-Lmin)/Decaylength)
+
+def inputWrite(mres,gax,nev,Lmin,Lmax):#need to apply Rmin & Rmax as well..
+    mres    = AlpacaFormatting(mres)
+    gax     = AlpacaFormatting(gax)
+    L   = AlpacaFormatting(float(Lmax)-float(Lmin))
+    Lmin    = AlpacaFormatting(Lmin)
+    print Lmin,Lmax,L
+    f=open("input.DAT","w")
```
olantwin: Would suggest using a contextmanager instead:
```python
with open("...", "w") as f:
    f.write(...)
    ...
```
Closing the file is then automatic (even if there is a crash or something else), so does not have to be done manually.

python/makeAlpacaEvents.py
```diff
@@ -0,0 +1,146 @@
...
+
+def runEvents(mres,gax,nev,Lmin,Lmax,startZ,endZ,SmearBeam):
+    pid=9900015
+    m = mres
+    g= gax
+    print m, g
+    pdg = ROOT.TDatabasePDG.Instance()
+    pdg.AddParticle('a','ALP', m, False, g, 0., 'a', pid)
+    wdir = os.environ['ALPACABIN']
+    #wdir="/home/atakan/ShipSoft/sw/ubuntu1804_x86-64/alpaca/v1.1-2/bin"
```
olantwin: remove

shipgen/AlpacaGenerator.cxx
```diff
@@ -0,0 +1,89 @@
...
+{
+  while (fn<fNevents) {
+  fTree->GetEntry(fn);
+  if (track==0&&parent==-1) {cpg->AddTrack(pdg,px,py,pz,x,y,z,-1.,false,e,tof,w);} //Be sure the order, units, and possible errors(i.e pdg/track_id)
+  fTree->GetEntry(fn+1);
+  if (track==1&&parent==0) {cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w);}
+  fTree->GetEntry(fn+2);
+  if (track==2&&parent==0) {cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w);}
+  fn=fn+3;
+  cout << "eee?"<<fn <<endl;
```
olantwin: What's this?

shipgen/AlpacaGenerator.cxx
```diff
@@ -0,0 +1,89 @@
...
+  break;
+  }
+  //if (fn==fNevents) {
+     //cout << "No more input events"<<endl;
+     //return kFALSE; }
```
olantwin: Why is this commented?

olantwin:
Looks good apart from the questions/comments above :)

macro/run_simScript.py
tugberk92: apparently it takes integer because it does not work.
So, deleting it would be better.

macro/run_simScript.py
tugberk92: OK, let me do it.

macro/run_simScript.py
tugberk92: OK, thanks.

macro/run_simScript.py
olantwin: If you want to keep the comments for your local work, you can leave them in your local copy but remove them in the committed version (e.g. via `git add -p`).

python/makeAlpacaEvents.py
tugberk92: does it work even for ```/afs/cern.ch/...```?
Then, nice!

python/makeAlpacaEvents.py
tugberk92: OK.

shipgen/AlpacaGenerator.cxx
tugberk92: it was used for debugging (in my language; soo?) which I forgot to delete...

shipgen/AlpacaGenerator.cxx
tugberk92: I replicated it from ```shipgen/NTupleGenerator.cxx```. However, it did not work properly in my case...

python/makeAlpacaEvents.py
olantwin: Unfortunately not (but data files should be on EOS anyway).

Note, this only works when opening files with ROOT (however, you can use ROOT to open any file by using a `?raw` suffix to the filename and then it can be used similarly to the usual python `open` function.

shipgen/AlpacaGenerator.cxx
olantwin: Please try to debug. What happens in the ALPACA case if you run out of events in the input file?

tugberk92:
> Looks good apart from the questions/comments above :)

Many thanks!! 
Your questions/comments are appreciated :)

olantwin:
Happy to help :)

macro/run_simScript.py
tugberk92: Perfect, thanks for the tip!

shipgen/AlpacaGenerator.cxx
tugberk92: OK

macro/run_simScript.py
olantwin: `git add -p` is also great to selectively add some changes to make it easier to figure out which logically belong together (e.g. is you rename a variable just add the related changes). Maybe of interest to you: https://jeremykun.com/2020/01/14/the-communicative-value-of-using-git-well/

python/makeAlpacaEvents.py
tugberk92: > Unfortunately not (but data files should be on EOS anyway).
Yes, indeed. The reason of that comment was this.

macro/run_simScript.py
tugberk92: Let me have a look.
Thank you for the tips again!

olantwin:
Could you squash the last two commits into one? Then git should be able to understand that it's a rename and not a deletion of the file.

If you make a mistake in a commit, you can use `git commit --amend` to fix it. Just pushing a second commit can be very confusing (see the discussion with Thomas in the other PR)

Why are you deleting the comment explaining the beam smearing?

shipgen/ALPACAGenerator.cxx
```diff
@@ -0,0 +1,88 @@
...
+  fTree->GetEntry(fn+1);
+  if (track==1&&parent==0) {cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w);}
+  fTree->GetEntry(fn+2);
+  if (track==2&&parent==0) {cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w);}
+  fn=fn+3;
+  break;
+  }
+  //if (fn==fNevents) {
+     //cout << "No more input events"<<endl;
+     //return kFALSE; }
```
olantwin: This would still be good to fix, as it will currently crash in a bad way if the file is shorter than the number of events requested...

python/dpProductionRates.py
```diff
@@ -54,47 +62,56 @@ def brMesonToMesonDP(mass,epsilon,mumPdg,dauPdg,doprint=False):
...
 #from interpolation of Pythia XS, normalised to epsilon^2
 def qcdprodRate(mass,epsilon,doprint=False):
     xs = 0
     if (mass > 3):
-        xs = math.exp(-5.673-0.8869*mass)
+        xs = math.exp(-5.928-0.8669*mass)
     elif (mass > 1.4):
-        xs = math.exp(-3.802-1.532*mass)
+        xs = math.exp(-4.1477-1.4745*mass)
     else:
```
olantwin: This branch seems superfluous, as xs is alsready initialised to 0 in line 95.

python/dpProductionRates.py
```diff
@@ -54,47 +62,56 @@ def brMesonToMesonDP(mass,epsilon,mumPdg,dauPdg,doprint=False):
...
+        avgMeson1 = getAverageMesonRate(mumPdg)*brM2DP[1]
+        return avgMeson*0.6, avgMeson1*0.6
+        #return avgMeson + avgMeson1
+    if not mumPdg==331:
+        avgMeson = getAverageMesonRate(mumPdg)*brM2DP
+        return avgMeson*0.6
 
 #from interpolation of Pythia XS, normalised to epsilon^2
 def qcdprodRate(mass,epsilon,doprint=False):
     xs = 0
```
olantwin: Maybe make this 0., just to ensure floats are used for the multiplication.

macro/run_simScript.py
```diff
@@ -291,6 +299,23 @@
...
+  target     = ship_geo.target
+  z_middle   = target.z0
+  DZ         = target.length/2
+  startZ     =   z_middle - DZ
+  endZ       =   z_middle + DZ
+  SmearBeam  = 1*u.cm
+  print(SmearBeam,"run")
+  Lmin       = ((ship_geo.Chamber1.z - ship_geo.chambers.Tub1length) - ship_geo.target.z0)/100.
+  Lmax       = (ship_geo.TrackStation1.z - ship_geo.target.z0)/100.
+  inputFile  = makeALPACAEvents.runEvents(options.theMass,options.theDPepsilon,options.nEvents,Lmin,Lmax,startZ,endZ,SmearBeam)
```
olantwin: Maybe you could print a log message here to tell the user that ALPACA is generating the input file?

macro/run_simScript.py
```diff
@@ -115,6 +118,9 @@
 if options.genie:    simEngine = "Genie"
 if options.nuradio:  simEngine = "nuRadiography"
 if options.ntuple:   simEngine = "Ntuple"
+if options.alpaca:
+    options.inputFile= False
```
olantwin: There's a way to add mutually exclusive arguments in `argparse`. Maybe have a look in the `argparse` documentation to disable the `inputFile` option when `--ALPACA` is used?

ThomasRuf:
Hi,
How do we proceed? Haven't seen any answers to the last comments.

macro/run_simScript.py
tugberk92: thanks it is done

macro/run_simScript.py
tugberk92: thank you for the advice.
It is also resolved.
However, it could be revised for every situation that inputFile is using.

tugberk92:
Hi,
Sorry for the delay.
Please have a look at the changes.

python/makeALPACAEvents.py
```diff
@@ -0,0 +1,140 @@
...
+    endZ        = float(endZ)
+    SmearBeam   = float(SmearBeam)
+    wdir = os.environ['ALPACABIN']
+    with open(wdir+"/outputs/output_"+str(mres)+"_"+str(gax)+".dat","r") as fp:
+        for line in fp:
+            if "Cross section" in line:
+                line=line.split()
+                xs=line[3]+"_"+line[5]
+    fn=fn.readlines()
+    inputFile="/eos/experiment/ship/data/ALP/ntuples/"+"alp_m"+str(mres)+"_g"+str(gax)+"_xs"+str(xs)+".root"
```
olantwin: I would add $EOSSHIP here, either hardcoded or taking it from the environment. The /eos/ mountpoint is only available on EOS.

python/makeAlpacaEvents.py
tugberk92: Now, ntuples are saving into eos.

macro/run_simScript.py
olantwin: From the message in 303 I don't think it's clear that an input file is generated. I would add a message before and after `makeALPACAEvents.runEvents`. Ideally also one after opening the file, so that in future it's easier to debug if e.g. the file can't be opened, or something went wrong while generating the file.

python/makeALPACAEvents.py
tugberk92: Okey.

shipgen/ALPACAGenerator.cxx
```diff
@@ -0,0 +1,89 @@
...
+ fInputFile->Close();
+ fInputFile->Delete();
+ delete fInputFile;
+}
+// -------------------------------------------------------------------------
+
+// -----   Passing the event   ---------------------------------------------
+Bool_t ALPACAGenerator::ReadEvent(FairPrimaryGenerator* cpg)
+{
+  while (fn<fNevents) {
```
olantwin: The formatting in this while loop is very confusing! I had to read it several times to see what's in each conditional. Could you please put each statement on it's own line?
What happens when the if conditions evaluate to `false`? Are we sure the events can just be skipped?

shipgen/ALPACAGenerator.cxx
olantwin: have you had a look at why this doesn't work anymore? I guess `fNevents` could be skipped over if it's not a multiple of 3. So maybe this can be changed to `fn>=fNevents`?

python/makeAlpacaEvents.py
olantwin: Ntuples should probably not be written to EOS directly (imagine use in HTCondor). Reading from EOS is great, but writing should ideally happen to some local scratch space and then the job can take care of copying files to EOS.

Sorry for the confusion!

python/makeALPACAEvents.py
olantwin: Although see the other comment: it's a bad idea to *write* to EOS directly.
I would suggest creating this file in the current working directory of the script.

macro/run_simScript.py
tugberk92: [this line](https://github.com/ShipSoft/FairShip/blob/f4baf2fbe05c9ad9f5d87f6df7906f1562957893/shipgen/ALPACAGenerator.cxx#L25) tells it is generated and opening.
[This one](https://github.com/ShipSoft/FairShip/blob/f4baf2fbe05c9ad9f5d87f6df7906f1562957893/shipgen/ALPACAGenerator.cxx#L28) gives an error message if the input is empty.
[In here](https://github.com/ShipSoft/FairShip/blob/f4baf2fbe05c9ad9f5d87f6df7906f1562957893/shipgen/ALPACAGenerator.cxx#L72), it tells us the nth event is reading.
For the end message, I will fix [the condition](https://github.com/ShipSoft/FairShip/pull/332#discussion_r397204266).
Also for the ALPACA generation, It has it's own message e.g. as Pythia.

olantwin: I saw the messages in the generator (and they are very informative, but did not see any in the makeALPACAEvents script.

If ALPACA has its own messages, i would suggest having a message before entering it and after leaving it. My main worry is that it might be hard to debug right now if e.g. the ALPACA file isn't written properly (maybe due to some AFS issue), but ALPACA claims it was successful. Then you might not have any indication of an error until the generator tries to open the file.

python/makeAlpacaEvents.py
tugberk92: Okay, I am going to fix it as soon as my jobs are done in HTCondor.

macro/run_simScript.py
tugberk92: I added printouts into [run_simScripy](https://github.com/ShipSoft/FairShip/blob/6dddb88070d7416db8ebbe02081dd940a3a75d9a/macro/run_simScript.py), [makeALPACAEvents](https://github.com/ShipSoft/FairShip/blob/6dddb88070d7416db8ebbe02081dd940a3a75d9a/python/makeALPACAEvents.py) and [ALPACAGenerator](https://github.com/ShipSoft/FairShip/blob/6dddb88070d7416db8ebbe02081dd940a3a75d9a/shipgen/ALPACAGenerator.cxx).
Can you have a look at the content of messages?

olantwin: Looks good to me! Thanks!

shipgen/ALPACAGenerator.cxx
tugberk92: >The formatting in this while loop is very confusing! I had to read it several times to see what's in each conditional.

I am sorry... :/
I put some comments. I hope that it is better now.


> Could you please put each statement on it's own line?

I couldn't understand...
However, new comments would be helpful. Can you have a look at [them](https://github.com/ShipSoft/FairShip/blob/bedcc7759239cc34d391759dd076daf7fc383be8/shipgen/ALPACAGenerator.cxx)?

> What happens when the if conditions evaluate to ```false```?

In ALPACA, events are produced with an ALP and its two daughters 
OR 
Events are not produced at all due to given options -> no input file -> generator would exit.


> Are we sure the events can just be skipped?

It would be better to have a conditional just in case.
I will have a look into this.


Thanks!

macro/run_simScript.py
tugberk92: Thanks!

python/makeALPACAEvents.py
tugberk92: Same in here, waiting for my jobs are done in HTCondor.

shipgen/ALPACAGenerator.cxx
```diff
@@ -0,0 +1,89 @@
...
+  fTree->GetEntry(fn); // ALP is getting,
+  if (track==0&&parent==-1) {cpg->AddTrack(pdg,px,py,pz,x,y,z,-1.,false,e,tof,w);} // ALP track is adding.
```
olantwin:
```suggestion
  if (track==0 && parent==-1) {
      // Add ALP track
      cpg->AddTrack(pdg,px,py,pz,x,y,z,-1.,false,e,tof,w);
  } 
```

As an example of how I would find it more readable.
Otherwise when quickly skimming the code it can look as if the `fTree->GetEntry(...)` calls are guarded by the conditional, which is what confused me initially.

shipgen/ALPACAGenerator.cxx
olantwin: > [...]
> I put some comments. I hope that it is better now.

See my example below as an explanation of what I mean. The comments help (but it's "is add*ed*", not "is adding"). Would remove the comment on line 65.

> > Could you please put each statement on it's own line?
> 
> I couldn't understand...
> However, new comments would be helpful. Can you have a look at [them](https://github.com/ShipSoft/FairShip/blob/bedcc7759239cc34d391759dd076daf7fc383be8/shipgen/ALPACAGenerator.cxx)?
> 
> > What happens when the if conditions evaluate to `false`?
> 
> In ALPACA, events are produced with an ALP and its two daughters
> OR
> Events are not produced at all due to given options -> no input file -> generator would exit.

So if any of these if statements is not `true`, the code should exit?

Currently my understanding is that the ntuple always has groups of three entries with ALP, daughter1, daughter2. Is that correct?

> > Are we sure the events can just be skipped?
> 
> It would be better to have a conditional just in case.
> I will have a look into this.

I think you misunderstood what I meant: if these conditionals are not `true`, the track is just skipped. Is this the intended behaviour? Should there be an error if the conditional is `false`?

shipgen/ALPACAGenerator.cxx
```diff
@@ -0,0 +1,98 @@
...
+      cpg->AddTrack(pdg,px,py,pz,x,y,z,-1.,false,e,tof,w); // ALP track is added.
+  }
+  fTree->GetEntry(fn+1); // First daughter/photon is getting,
+  if (track==1&&parent==0) {
+      cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w); // First daughter/photon track is added.
+  }
+  fTree->GetEntry(fn+2); // Second daughter/photon is getting,
+  if (track==2&&parent==0) {
+      cpg->AddTrack(pdg,px,py,pz,x,y,z,0,true,e,tof,w); // Second daughter/photon track is added.
+  }
```
olantwin: Much better.
Now the only remaining issue I see is that if any of these three conditions fail, none of the tracks should be saved.
I would suggest reading all three events into some buffer, check whether they are OK and only then save any of the tracks.

Alternatively, as you control the format of the NTuple, it might be easier to change its format (e.g. to be a TTree with a branch of MCTrack for each particle in the event)?

ThomasRuf:
Independent of the ongoing review of the code, it is important to have an internal note which describes the purpose of this new generator, what it is suppossed to do, what is used, how it works.

tugberk92:
> Independent of the ongoing review of the code, it is important to have an internal note which describes the purpose of this new generator, what it is suppossed to do, what is used, how it works.

Okey,
I have started to write. I will keep you updated.

ThomasRuf:
I downloaded this pull request together with shipdist pull request. No side effect on run_simScript.py with default option for HNL. Also tried:  $FAIRSHIP/macro/run_simScript.py --ALPACA 
Some minor comments:
Observation 1: alpaca produces "lot" of output. is there an option to suppress it? 
Observation 2: do we need a printout for each event? 
Event Number:1
Event Number:2
....

I would propose to merge both requests and continue with improving code with future pull requests.

olantwin:
Hi Thomas,
I agree that this can probably be merged, with the caveat that I think the input file format between `makeALPAKAEvents` and `run_simScript.py`/`ALPACAGenerator` should probably change, and I don't think that can be done in a backwards-compatible manner.

The `shipdist` PR can definitely be merged.

ThomasRuf:
I missed where the input file format is being discussed. However it looks to me as an intermediate storage, in principle the input file could be removed after run_simScript.py has finished. Like this, any format change is transparent. Or do I miss something?

olantwin:
> I missed where the input file format is being discussed. However it looks to me as an intermediate storage, in principle the input file could be removed after run_simScript.py has finished. Like this, any format change is transparent. Or do I miss something?

You are correct, currently it is created on demand, so changing the format would not be an issue. I wonder whether it would make sense to introduce the possibility to generate the file separately, which would mostly depend on how long ALPACA takes compared to the time it takes to transport the events through GEANT.

ThomasRuf:
ALPACA cpu time is completely negligible compared to the rest.

olantwin:
OK, so then let's merge this?

tugberk92:
Hi,
Thanks for the comments and the merge.
I will improve on the discussed points, and create future pull requests.

PR #335, opened 2020-03-28
Minor updates on DP and ALPACA

tugberk92:
- MesonMother option is modified,
- _rtdb error_ is avoided on the new merge for pbrem (we have the same error more than a year...),
- run makeALPACAEvents standalone via interactive python and parsers,
- printout of run_simScript for DP+ALPACA is revised, 
- some other minor changes related to ALPACA and pbrem,
- outputs of ALPACAGenerator is minimized.


macro/run_simScript.py
```diff
@@ -27,10 +26,7 @@
 
 # Default dark photon parameters
 theDPmass    = 0.2*u.GeV
-
-# Alpaca
-motherMode = True
-
```
olantwin: Please rebase your branch against master: this undoes a change by @ThomasRuf .

You should remove the commits included in the old pull request from this rebase to ensure everything works well (e.g. drop them in the interactive rebase).

macro/run_simScript.py
```diff
@@ -150,6 +145,7 @@
 if options.DarkPhoton: HNL = False
 if not options.theMass:
   if options.DarkPhoton: options.theMass  = theDPmass
+  if options.ALPACA:     options.theMass  = theALPmass
```
olantwin: What's the point of `--mass` if it's overwritten with the default here?

Also, please avoid writing conditionals and following statements on one line. It's much more readable if you write
```python
if condition:
    expression
```

macro/run_simScript.py
```diff
@@ -77,13 +73,14 @@
...
+parser.add_argument("-m", "--mass", dest="theMass",  help="Mass of hidden particle, default "+str(theHNLMass)+"GeV for HNL, "+str(theDPmass)+"GeV for DP, "+str(theALPmass)+"GeV for ALP", required=False, default=None, type=float)
 parser.add_argument("-c", "--couplings", "--coupling", dest="thecouplings",  help="couplings \'U2e,U2mu,U2tau\' or -c \'U2e,U2mu,U2tau\' to set list of HNL couplings.\
  TP default for HNL, ctau=53.3km", required=False,default="0.447e-9,7.15e-9,1.88e-9")
 parser.add_argument("-cp", "--production-couplings", dest="theprodcouplings",  help="production couplings \'U2e,U2mu,U2tau\' to set the couplings for HNL production only"\
                                             ,required=False,default=None)
 parser.add_argument("-cd", "--decay-couplings", dest="thedeccouplings",  help="decay couplings  \'U2e,U2mu,U2tau\' to set the couplings for HNL decay only", required=False,default=None)
 parser.add_argument("-e", "--epsilon", dest="theDPepsilon",  help="to set mixing parameter epsilon", required=False,default=0.00000008, type=float)
+parser.add_argument("-gax", dest="gax",  help="to set photon coupling coefficient", required=False,default=1.0e-5, type=float)
```
olantwin: this should be `--gax`; it also automatically ends up in `options.gax`, so that `dest=` does not need to be specified.

macro/run_simScript.py
```diff
@@ -165,7 +161,9 @@
 if (HNL and options.RPVSUSY) or (HNL and options.DarkPhoton) or (options.DarkPhoton and options.RPVSUSY): 
  print("cannot have HNL and SUSY or DP at the same time, abort")
  sys.exit(2)
-
+if inclusive=='meson' and options.MM not in ['pi0','eta','omega','eta1','eta11']:
+ print("Please choose a DP sources from one of these (pi0, eta, omega, eta1, eta11) for the Meson Production")
+ sys.exit()
```
olantwin: You can use [argparse choices](https://docs.python.org/3/library/argparse.html#choices) to enforce this automatically, so that you don't need to do it manually.

python/makeALPACAEvents.py
```diff
@@ -1,12 +1,11 @@
...
 import os,sys,getopt
 import numpy,math
 import ROOT
-
-hGeV = 6.58211928*pow(10.,-16)* pow(10.,-9) # no units or it messes up!!
-c_light = 2.99792458e+10
-mres="1.0d0"
-gax="1d-7"
-nev="100"
+hGeV      = 6.58211928*pow(10.,-16)* pow(10.,-9) # no units or it messes up!!
```
olantwin: Why not `6.58211928e-27`? Does that give the same results? If not, can you use `pow(10.,-27)`?

olantwin: Also, please don't change the formatting like this. There should be exactly one space before and after `=`. This and the following line are changed for no reason.

python/makeALPACAEvents.py
```diff
@@ -34,7 +32,7 @@ def inputWrite(mres,gax,nev,Lmin,Lmax):#need to apply Rmin & Rmax as well..
     L   = ALPACAFormatting(float(Lmax)-float(Lmin))
     Lmin    = ALPACAFormatting(Lmin)
     Lmax    = ALPACAFormatting(Lmax)
-    with open("input.DAT","w") as f:
+    with open("input_"+str(mres)+"_"+str(gax)+".DAT","w") as f:
```
olantwin: I would evaluate this name once and then reuse it everywhere in this file. I see at least 3 instances of this evaluation.
If you ever want to change it in future, it's easy to make mistakes by forgetting to change it somewhere (or introducing a typo in one of the 3 cases).

Same applies for the output names.

This is something you can have in the global scope of the file (or maybe in the module namespace).

python/makeALPACAEvents.py
```diff
@@ -83,6 +81,12 @@ def inputWrite(mres,gax,nev,Lmin,Lmax):#need to apply Rmin & Rmax as well..
         f.write("****************************************************************************************\n")
         f.write("****************************************************************************************\n")
         f.write("****************************************************************************************\n")
+def getXS(wdir,mres,gax):
```
olantwin: The convention usually is to have blank lines between function definitions.

python/makeALPACAEvents.py
```diff
@@ -120,17 +120,22 @@ def ntupleWrite(ctau,fn,Lmin,Lmax,startZ,endZ,SmearBeam,mres,gax,old):
         ntup.Fill(int(i),int(2),int(dau2[1]),float(dau2[7]),float(dau2[8]),float(dau2[9]),daux,dauy,dauz,int(0),float(1),float(dau2[10]),float(dau2[15])/10./c_light,w)
     ntup.Write()
     f.Close()
+    print("ALP -> Gamma + Gamma events are produced with cross-section of {} pb at mass of {} GeV and photon coupling of {} GeV^-1".format(xs,mres,gax))
     return inputFile
+def addALPtoPDG(mres,gax):
```
olantwin: Blank line between functions.

python/makeALPACAEvents.py
```diff
@@ -139,6 +144,59 @@ def runEvents(mres,gax,nev,Lmin,Lmax,startZ,endZ,SmearBeam):
     print('Events are recording into a ntuple.')
     inputFile = ntupleWrite(ctau,fn,Lmin,Lmax,startZ,endZ,SmearBeam,mres,gax,old)
     os.chdir(old)
+    os.system("rm "+wdir+"/input_"+str(mres)+"_"+str(gax)+".DAT")
+    os.system("rm "+wdir+"/evrecs/evrec_"+str(mres)+"_"+str(gax)+".dat")
+    os.system("rm "+wdir+"/outputs/output_"+str(mres)+"_"+str(gax)+".dat")
```
olantwin: It's better to use `os.remove`.

python/makeALPACAEvents.py
```diff
@@ -139,6 +144,59 @@ def runEvents(mres,gax,nev,Lmin,Lmax,startZ,endZ,SmearBeam):
...
+    if options.endZ:
+        endZ=options.endZ
+        print("endZ is set to {} cm".format(endZ))
+    else: print("default endZ is {} cm".format(endZ))
+    if options.nev:
+        nev=options.nev  
+        print("nEvents is set to {}".format(nev))
+    else: print("default nEvents is {}".format(nev))
+    runEvents(mres,gax,nev,Lmin,Lmax,startZ,endZ,SmearBeam)   
+#main()
```
olantwin: Usually it's a good idea to do
```python
if __name__='__main__':
    main()
```
Which makes sure that `main()` is run when the script is called directly, but **not** if it is imported in e.g. `run_simScript.py`.

python/proton_bremsstrahlung.py
```diff
@@ -188,7 +188,7 @@ def hProdPDF(mDarkPhoton, epsilon, norm, binsp, binstheta, tmin = -0.5 * math.pi
             hPDF.Fill(p,theta,w)
             hPDFtheta.Fill(theta,w)
             hPDFp.Fill(p,w)
-    hPdfFilename = sys.modules['__main__'].outputDir+"/ParaPhoton_eps%s_m%s%s.root"%(epsilon,mDarkPhoton,suffix)
```
olantwin: This isn't particularly elegant, but appreciate the attempt to not hard-code the location.

tugberk92:
Hi Oliver,
Thank you for your advice.
I will take care of them.

macro/run_simScript.py
```diff
@@ -557,13 +556,14 @@
...
 run.Run(options.nEvents)
 # -----Runtime database---------------------------------------------
-kParameterMerged = ROOT.kTRUE
-parOut = ROOT.FairParRootFileIo(kParameterMerged)
-parOut.open(parFile)
-rtdb.setOutput(parOut)
-rtdb.saveOutput()
-rtdb.printParamContexts()
-getattr(rtdb,"print")()
+if not options.DarkPhoton:
```
tugberk92: So far, I saw this error in the pbrem case. However, I thought it would be better to apply it to all DP productions.
How should we proceed with this problem?

olantwin: @ThomasRuf what do you think? As a workaround it's fine, but I think we should try to understand why the `rtdb` causes a crash/error. Do we actually use it for any purpose currently?

macro/run_simScript.py
ThomasRuf: "Choose DP production meson source(pi0,eta,omega,eta1,eta11" together with pythia8darkphoton_conf.configure, see below.

I don't think the help text explains the real meaning of the variable. It is not only selecting the mother particle but also the production mode. Particle called eta1 or eta11 is confusing, please use names from ROOT.TDatabasePDG.Instance().



    if motherMode=='pi0' and pi0mass-mass>=0.0000001:
        # use pi0 -> gamma A'
        selectedMum = 111
    elif motherMode == 'eta' and etamass-mass>=0.000001:
        # use eta -> gamma A'
        #print "eta"
        selectedMum = 221
    elif motherMode=="omega" and omegamass-mass>=0.00001:
        # use omega -> pi0 A'
        #print "omega"
        selectedMum = 223
        P8gen.SetParameters("223:oneChannel = 1 1 0 111 9900015")
        
    elif motherMode=='eta1' and eta1mass-mass>=0.00001:
        # use eta' -> gamma A'
        selectedMum = 331
        P8gen.SetParameters("331:oneChannel = 1 1 0 22 9900015")
        #should be considered also for mass < 0.188 GeV....
        #P8gen.SetParameters("331:oneChannel = 1 1 0 223 9900015")29%BR
        #P8gen.SetParameters("331:oneChannel = 1 1 0 113 9900015")2.75%BR
        
    elif motherMode=='eta11' and eta1mass-mass>=0.00001:
        # use eta' -> gamma A'
        selectedMum = 331

macro/run_simScript.py
ThomasRuf: i am lost. what is the pbrem case error?

python/makeALPACAEvents.py
ThomasRuf: what about using >>> ROOT.TMath.C()
299792458.0
>>> ROOT.TMath.H()
6.62606876e-34

if everybody hardcodes its own values, will be a mess.

python/makeALPACAEvents.py
ThomasRuf: is there a way to have this intermediate file in memory? How big are these files usually in MB?

python/makeALPACAEvents.py
olantwin: I agree: Would be much nicer to not hard-code these!

Are they already in `shipunit`?

python/makeALPACAEvents.py
olantwin: Don't know about the file size, but there are ways to have ROOT and non-ROOT files in memory only (or as temporary files).

macro/run_simScript.py
tugberk92: Hi Thomas,
You investigated this error last year(27 Feb, 2019) and title was " Error in Dark Photon Production (in Proton Bremsstrahlung Process)";


```FairShip/macro/run_simScript.py --DarkPhoton -A pbrem --mass 0.88 --epsilon 1.5848932e-07 -n 2```

...

```run.Run(2)```
...
>>>Debug: decay product of A: 2 -2
>>>[INFO] ship event 0 / pythia event-nr 0
>>>Debug: decay product of A: 13 -13

```kParameterMerged = ROOT.kTRUE```
```parOut = ROOT.FairParRootFileIo(kParameterMerged) >>> parOut.open(parFile)```

>>>True

```rtdb.setOutput(parOut)```

>>>True

``` rtdb.saveOutput()```

>>>*** Break *** segmentation violation

Comment (ThomasRuf) [reply macro/run_simScript.py]: I agree with Oliver, would be better to understand what is the reason for the problem, rather than to hide it.

File: python/makeALPACAEvents.py (hunk 5 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: okay, agreed.
Then, should I  change this [line](https://github.com/ShipSoft/FairShip/blob/master/python/shipunit.py#L271)?

File: macro/run_simScript.py (hunk 2 above)
Comment (tugberk92) [reply macro/run_simScript.py]: Okay, I will rewrite the text and put the names from PDG database.
I need to think about the name of eta1 and eta11 since eta' has two different production modes.
I am open to suggestions :D

File: python/makeALPACAEvents.py (hunk 5 above)
Comment (olantwin) [reply python/makeALPACAEvents.py]: > okay, agreed.
> Then, should I change this [line](https://github.com/ShipSoft/FairShip/blob/master/python/shipunit.py#L271)?

I would say yes (and `c_light` and `pi` as well).

But we might want to consider using <https://github.com/scikit-hep/hepunits> rather than implementing all of these ourselves. The package is actively maintained by the HEP python community (including but not limited to several LHCb authors and is in my experience very reliable).

File: macro/run_simScript.py (hunk 2 above)
Comment (olantwin) [reply macro/run_simScript.py]: Which production modes are these? that might give me some inspiration for naming.

Comment (tugberk92) [reply macro/run_simScript.py]: Dark Photon Production via mesons

Comment (olantwin) [reply macro/run_simScript.py]: You said there are two different production modes of the eta'. What are they?

Comment (olantwin) [reply macro/run_simScript.py]: Different thing:
Is there a reason the epsilons are not all the same?
It might be more understandable what you're doing if you define `epsilon = 0.00001` once and then test for `mass >= <meson>mass + epsilon`. Does `>` not suffice instead of the `>= small number` construction?

Comment (tugberk92) [reply macro/run_simScript.py]: having an epsilon variable would be a nice idea but they have different significant figures. So, epsilon should differ, right?

Comment (tugberk92) [reply macro/run_simScript.py]: > Different thing:
> Is there a reason the epsilons are not all the same?
> It might be more understandable what you're doing if you define `epsilon = 0.00001` once and then test for `mass >= <meson>mass + epsilon`. Does `>` not suffice instead of the `>= small number` construction?

I modified it.

Comment (tugberk92) [reply macro/run_simScript.py]: > You said there are two different production modes of the eta'. What are they?

they are;
pi0 -> photon + darkphoton
eta -> photon + darkphoton
omega -> pi0 + darkphoton
eta' -> photon + darkphoton
eta' -> rho + darkphoton

Comment (tugberk92) [reply macro/run_simScript.py]: still eta' remains as eta1 and eta11.
help message will change un run_simScript.
 I will think about them.

File: python/makeALPACAEvents.py (hunk 5 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: I am using the defn from shipunit now.

File: python/makeALPACAEvents.py (hunk 6 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: I made a  defn for this situation.

File: python/makeALPACAEvents.py (hunk 8 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: Thank you for your attention.
Done.

File: python/makeALPACAEvents.py (hunk 10 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: Thank you for your hint.
Done.

File: python/makeALPACAEvents.py (hunk 7 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: Thank you for your attention.
Done.

File: python/makeALPACAEvents.py (hunk 9 above)
Comment (tugberk92) [reply python/makeALPACAEvents.py]: I used os.remove.
Thank you for your tip.
Done.

Comment (tugberk92) [reply python/makeALPACAEvents.py]: > is there a way to have this intermediate file in memory? How big are these files usually in MB?

Honestly, I did not know. I will have a look and report here.
Since ALPACA needs these files inside bin dir, I did not want to mess the sw dir and made this deletion.

**Comment by tugberk92** [comment]
I think I revised the makeALPACAEvents reviews.
They were so useful.
I will have a look for the others.
Many thanks!

File: python/proton_bremsstrahlung.py (hunk 11 above)
Comment (tugberk92) [reply python/proton_bremsstrahlung.py]: I have seen elegant ones in somewhere at the FairShip.. Maybe it was in the run_fixedTarget.py of Thomas.
I will have a look at it.

File: macro/run_simScript.py (hunk 2 above)
Comment (olantwin) [reply macro/run_simScript.py]: Maybe they could be etaprime and etaprime_rho?

**Comment by olantwin** [comment]
Before you do anything else, please `rebase -i` against master, dropping the commits already included in the old pull request.

I'll have a look at your last commit now.

**Comment by olantwin** [comment]
> Before you do anything else, please `rebase -i` against master, dropping the commits already included in the old pull request.
> 
> I'll have a look at your last commit now.

Could you please push updates as separate commits? It's hard to review changes if you keep amending the same commit...

File: python/makeALPACAEvents.py [hunk 13]
```diff
@@ -1,12 +1,16 @@
...
 import numpy,math
 import ROOT
+import shipunit as u
 
-hGeV = 6.58211928*pow(10.,-16)* pow(10.,-9) # no units or it messes up!!
-c_light = 2.99792458e+10
-mres="1.0d0"
-gax="1d-7"
-nev="100"
+#natural units were used!
```
olantwin:
```suggestion
# Using natural units
```

python/makeALPACAEvents.py
```diff
@@ -120,25 +130,82 @@ def ntupleWrite(ctau,fn,Lmin,Lmax,startZ,endZ,SmearBeam,mres,gax,old):
...
     old=os.getcwd()
     os.chdir(wdir)
     inputWrite(mres,gax,nev,Lmin,Lmax)
-    rn="./alpaca < input.DAT"
+    rn="./alpaca < "+"input"+naming+".DAT"
     os.system(rn)
     print('ALPACA generated the events.')
-    pa="./evrecs/evrec_"+str(mres)+"_"+str(gax)+".dat"
+    pa="./evrecs/evrec"+naming+".dat"
     fn=open(pa,"r")
```
olantwin: 1. fn is never closed. I would suggest using a context manager again (`with open(pa, "r") as fn:`.

2. Also, please avoid names such as `fn`, `pa`, `rn`. They are meaningless to anyone but you (and even you might forget what they mean), and make the code harder to follow because of that.
Exceptions to this rule of thumb would be names conventionally used (e.g. because of the mathematical meaning).

python/makeALPACAEvents.py
```diff
@@ -120,25 +130,82 @@ def ntupleWrite(ctau,fn,Lmin,Lmax,startZ,endZ,SmearBeam,mres,gax,old):
...
-    pdg = ROOT.TDatabasePDG.Instance()
-    pdg.AddParticle('a','ALP', mres, False, gax, 0., 'a', 9900015)
+    addALPtoPDG(float(mres),float(gax))
     wdir = os.environ['ALPACABIN']
     old=os.getcwd()
     os.chdir(wdir)
     inputWrite(mres,gax,nev,Lmin,Lmax)
-    rn="./alpaca < input.DAT"
+    rn="./alpaca < "+"input"+naming+".DAT"
     os.system(rn)
```
olantwin: Please use `subpocess.run` instead of `os.system`.

python/makeALPACAEvents.py
```diff
@@ -120,25 +130,82 @@ def ntupleWrite(ctau,fn,Lmin,Lmax,startZ,endZ,SmearBeam,mres,gax,old):
...
+    print("ALP -> Gamma + Gamma events are produced with cross-section of {} pb at mass of {} GeV and photon coupling of {} GeV^-1".format(xs,mres,gax))
     return inputFile
 
 def runEvents(mres,gax,nev,Lmin,Lmax,startZ,endZ,SmearBeam):
+    mres = float(mres)
+    gax  = float(gax)
     print('ALPACA is starting for mass of {} GeV with photon coupling coeffiecient {} GeV^-1.'.format(mres,gax))
-    pdg = ROOT.TDatabasePDG.Instance()
-    pdg.AddParticle('a','ALP', mres, False, gax, 0., 'a', 9900015)
+    addALPtoPDG(float(mres),float(gax))
```
olantwin: Casting again here seems superfluous.

I would suggest to rename the function arguments or local variables to make this less confusing.

ThomasRuf:
Sorry, I am completely lost with all these hundreds of changes. Could we make a reset, and start with one issue after the other?

Von: Oliver Lantwin <notifications@github.com>
Gesendet: Montag, 30. März 2020 14:52
Cc: Thomas Ruf <Thomas.Ruf@cern.ch>; Mention <mention@noreply.github.com>
Betreff: Re: [ShipSoft/FairShip] Minor updates on DP and ALPACA (#335)


@olantwin commented on this pull request.

In python/makeALPACAEvents.py<https://github.com/ShipSoft/FairShip/pull/335#discussion_r400165363>:

>      print('ALPACA is starting for mass of {} GeV with photon coupling coeffiecient {} GeV^-1.'.format(mres,gax))

-    pdg = ROOT.TDatabasePDG.Instance()

-    pdg.AddParticle('a','ALP', mres, False, gax, 0., 'a', 9900015)

+    addALPtoPDG(float(mres),float(gax))

Casting again here seems superfluous.

I would suggest to rename the function arguments or local variables to make this less confusing.

olantwin:
What would you suggest in practice?

Please note that the pull request is not mergeable until the Atakan performs the rebase to remove changes already in master (and in one instance undo a recent change in master).

ThomasRuf:
My problem, I cannot follow so many issues in parallel. While I understand one issue, you have sent already an answer, and Atakan sends out a new PR. Maybe if you have time, you follow it up with Atakan, and I wait until the dust has settled.
Cheers,
thomas

Von: Oliver Lantwin <notifications@github.com>
Gesendet: Montag, 30. März 2020 15:53
Cc: Thomas Ruf <Thomas.Ruf@cern.ch>; Mention <mention@noreply.github.com>
Betreff: Re: [ShipSoft/FairShip] Minor updates on DP and ALPACA (#335)


What would you suggest in practice?

Please note that the pull request is not mergeable until the Atakan performs the rebase to remove changes already in master (and in one instance undo a recent change in master).

olantwin:
How about the following: we close this pull request and @tugberk92 can make a pull request against my repository, where we deal with the different nitpicks and tips I have been giving here during the review.

Once that's done we can then create a new pull request here.

tugberk92:
Okay, @olantwin.
Sounds good to me.

ThomasRuf:
Sounds good.

Von: Oliver Lantwin <notifications@github.com>
Gesendet: Montag, 30. März 2020 16:25
Cc: Thomas Ruf <Thomas.Ruf@cern.ch>; Mention <mention@noreply.github.com>
Betreff: Re: [ShipSoft/FairShip] Minor updates on DP and ALPACA (#335)


How about the following: we close this pull request and @tugberk92<https://github.com/tugberk92> can make a pull request against my repository, where we deal with the different nitpicks and tips I have been giving here during the review.

Once that's done we can then create a new pull request here.

PR #351, opened 2020-05-30
TT digi

aleksanp:
Added a TT digitization script that simulates the signal like from real SciFi modules. It reads the data file (selected manually), analyzes and shows some graphs from TT.


aleksanp:
This is the first stable version, but it should be upgraded in the future to take into account the effect on the signal depending on the angle of passage of particles

aleksanp:
Hi! 
1a) I used the PG only for troubleshooting because in a run usually TT histogram is empty. But the script doesn't run the PG, it takes the file after runsimscript (`ship.conical.Pythia8-TGeant4_rec.root` or `ship.conical.PG_13-TGeant4.root` for my case...)
1b) Ok, I could see the ShipDigiReco.py and try to integrate.

2.) I tried to use global values as much as possible. 
But for example, `cm_to_chan` includes the SiPMs array size and design.
`ly_loss_mean` and cluster_width_mean includes formulas from experimental studies.
`ly_to_CDF`, `CDF_to_ly`, `ly_to_CDF_landau`, `edep_to_ly` includes approximations from experimental data.
Where does this relate to the global size of the detector I used 
`ship_geo.NuTauTT.<...>`

shir994:
In the current implementation, it is unusable. I agree with Thomas that you need to redefine hardcoded numbers. Even if you do not use ship_geo values, you need to properly define all the numbers as variables, and do not let them be standalone unnamed magic numbers. Also, ideally you can logically split your code into classes/scripts, to make it easier to maintain in the future. Also, as Thomas mentioned it should not be one huge macro that does everything and indeed should be integrated with `DigiReco`.

PR #352, opened 2020-06-04
Update on the UpstreamTaggerHit.cxx class

CelsoFCF:
Attempt to remove warnings while debugging the code.

### DIFF
--- a/UpstreamTagger/UpstreamTaggerHit.cxx
+++ b/UpstreamTagger/UpstreamTaggerHit.cxx
@@ -148,71 +148,69 @@ TGeoNode* UpstreamTaggerHit::GetNode(Double_t &hit_final, Int_t &mod)
    TGeoNavigator* nav = gGeoManager->GetCurrentNavigator();
    mod = 0;
 
-   Double_t XHit = point_final[0];
-   Double_t YHit = point_final[1];
-   Double_t ZHit = point_final[2];
+   Double_t XHit1 = point_final[0];
+   Double_t YHit1 = point_final[1];
+   Double_t ZHit1 = point_final[2];
 
-   const Double_t PxHit = *mom[0];
-   const Double_t PyHit = *mom[1];
-   const Double_t PzHit = *mom[2];
+   const Double_t PxHit1 = *mom[0];
+   const Double_t PyHit1 = *mom[1];
+   const Double_t PzHit1 = *mom[2];
 
    
-   Double_t module[11][3];
-   Double_t det_zGlassPos = 0.2;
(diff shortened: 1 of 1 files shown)


ThomasRuf:
I don’t think this is the right solution to your problem. 

If I understand correctly, the problem originates from the fact that you define variables in UpstreamTagger.h and the same variables in UpstreamTaggerHit.cxx:

UpstreamTagger/UpstreamTagger.h:    Double_t     det_zGlassPos;     //!  z-size of Active Glass plates
UpstreamTagger/UpstreamTaggerHit.cxx:   Double_t det_zGlassPos = 0.2;

and UpstreamTaggerHit.cxx includes UpstreamTagger.h.

your solution, change name in UpstreamTaggerHit.cxx: Double_t det_zGlassPos1 = 0.2;

This is the wrong approach. det_zGlassPos is the property of an UpstreamTagger object. It is created by the constructor with a default value of 0, det_zGlassBorderPos(0), and redefined via void SetSizeZ_Glass(Double_t zg) {det_zGlassPos = zg;} 

In ShipReco.py,
# -----Create geometry----------------------------------------------
modules = shipDet_conf.configure(run,ShipGeo) 

This will create all detector modules with the same geometry as was used when running the simulation. By the way, this is the other big problem of your approach, how to synchronize the geometry used in the simulation and digitization? 

My proposal, assuming I understood what is happening, in shipDigiReco.py, when you convert point to hit

aHit = ROOT.UpstreamTaggerHit(aMCPoint,self.sTree.t0)

you could pass in addition the pointer to the UpstreamTagger detector, global_variables.modules["UpstreamTagger"], and extract all the information you need from there. 

HOWEVER, also realize that doing this via a constructor from TimeDetPoint from TimeDetHit is maybe not the best way. Would be better to have one object UpstreamTaggerHit  which is created and initialized once with the UpstreamTagger detector pointer, and you call a method digitize of this object to make the transformation. 

Best regards,
Thomas


PS: in addition, your file has a git problem: "This branch has conflicts that must be resolved"

olantwin:
1. We should also consider the size on disk of the UpstreamTaggerHit, since this would also be the data format for actual data. 
What we could do is make properties of the geometry static members of the class, such that they are shared between all instances of UpstreamTaggerHit (and not persist them to disk).
Alternatively, what e.g. the strawtubes do is to load the information from the geometry and using its methods directly. This might be a better approach. I suggest looking at the implementation there.

2. Please use git rather than the github file upload. Not only does it *consistently* cause conflicts, but the default "Add files via upload" adds no useful information about the changes, which makes reviewing them unnecessarily difficult.

PR #363, opened 2020-09-11
fieldMap generation + tools

PlaZMaD:
2 new scripts to add noise and convert fieldmap
1 new method for ShipFieldMaker to generate the fieldmap for the given region.

ShipFieldMaker::generateFieldMap("output_file_name") can be called in the main simulation script to generate the csv fieldMap for the MuonShield region. Can be configured for different step size and positions.

Generated csv should be processed by add_noise_to_field.py.
Options to use:
--input #input file name
--output #output file name
--sidesOnly #not a random noise cores, but just an edge distortions
--sigma #sigma value for Gaussian filter
--nCores #number of  gauss noise cores over the field volume (ignored if sidesOnly is used)
--peak #max peak values for the gaussian cores (ignored if sidesOnly is used)
--fraction #contols the overall noise effect strength (ignored if sidesOnly is used)

New csv should be converted to root by converNoisyMap.py   (modification of the common convert.py. Pandas lib is needed),
no options, hardcoded values only.

Default values (sigma=30, nCores=1000, peak=500, fraction=0.4)    -> Generaly give ~same efficiencies of our standard fieldMap
Bzy
(sigma=8, nCores=1000, peak=500, fraction=0.4)
Bzy
(sigma=6, sidesOnly)
Bzy

### DIFF
--- a/field/ShipFieldMaker.cxx
+++ b/field/ShipFieldMaker.cxx
@@ -1380,6 +1380,46 @@ void ShipFieldMaker::plotField(Int_t type, const TVector3& xAxis, const TVector3
 
 }
 
+void ShipFieldMaker::generateFieldMap(TString fileName, const float step, const float xRange, const float yRange, const float zRange, const float zShift){
+        std::ofstream myfile;
+        myfile.open (fileName);
+        int xSteps = ceil(xRange/step) + 1;   //field map has X range from 0 to xMax
+        int ySteps = ceil(yRange/step) + 1;   //from 0 up to yMax
+        int zSteps = ceil(zRange*2./step) + 1;//from -zMax up to zMax
+        Double_t position[3] = {0.0, 0.0, 0.0};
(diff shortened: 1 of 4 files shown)


ThomasRuf:
I have two questions:
1) What is the pupose of adding random noise to a magnetic field?
2)When generating a fieldmap on the fly with this new method ShipFieldMaker::generateFieldMap, how does it fit with the procedure to addVMCFields of geomGeant4.py? Global, local, local+global?

PlaZMaD:
> I have two questions:
> 
> 1. What is the pupose of adding random noise to a magnetic field?
>    2)When generating a fieldmap on the fly with this new method ShipFieldMaker::generateFieldMap, how does it fit with the procedure to addVMCFields of geomGeant4.py? Global, local, local+global?

1) This is a new optimization feature. We are expecting new solutions to be more robust to the ideal->realistic field transformation.
2)Same as plotField method. Get volume->get local volume field and return its value. No local field->check global one. No global -> return 0. If we expect something different, should fix the plot method as well.

PR #367, opened 2020-11-17
2021 edition

ThomasRuf:
changes required for new FairRoot


olantwin:
It's a shame that this was squashed, as it's now impossible to tell what changes were due to the line endings and which are for the new FairRoot (without looking at ~7000 lines in detail).

But glad we finally moved to the new FairRoot. I will update the docker base image so that the docker build works again.

ThomasRuf:
I have no idea what squash does, just followed the usual procedure. Maybe it helps for your purpose if you look at https://github.com/ThomasRuf/FairShip ?

passive/ShipMuonShield.h
```diff
@@ -34,7 +34,7 @@ class ShipMuonShield : public FairModule
 
   void SetSupports(Bool_t supports) { 
     fSupport = supports;
-    FairLogger::GetLogger()->Warning(MESSAGE_ORIGIN, "Setting supports to %s. This will not have any effect if called after the geometry has been constructed.", fSupport ? "true" : "false");
+    LOG(INFO) <<"Setting supports to %s. This will not have any effect if called after the geometry has been constructed.", fSupport ? "true" : "false";
```
olantwin: Shouldn't this be `LOG(WARN)`? We should not change the severity of these, unless the change is intentional.

python/hnl.py
```diff
@@ -274,8 +274,9 @@ def I(self,x1,x2,x3):
         """
         Numerical integral needed for 3-body decays trough W boson.
         xi = mi/MN
-        """
-        func = ROOT.TF1('func',self.Integrand,0,1,3) # Setting function
+        """ 
+        theFunction = self.Integrand
+        func = ROOT.TF1('func',theFunction,0,1,3) # Setting function
```
olantwin, line 279: I don't understand why this change would make any difference.

ThomasRuf:
yes, good question. This didn't worked anymore with the new ROOT. I think it is related to the deletion, garbage collection, ... of temporary objects. By making it explicit, not as argument, it worked again. There was another place with a similar change.

olantwin:
> I have no idea what squash does, just followed the usual procedure. Maybe it helps for your purpose if you look at https://github.com/ThomasRuf/FairShip ?

When merging pull requests there are several options. By default you squash all commits into a single one, but you can also rebase or merge commits one-by-one to conserve the history. I think you just need to press a different button or uncheck a checkbox when merging.

In this case I can reconstruct it (and am looking through the commits one-by-one right now), but generally it's worth considering which option might be best for a given set of changed.

passive/ShipMuonShield.h
ThomasRuf: yes, you are right. Will correct it.

shipdata/ShipStack.cxx
```diff
@@ -418,8 +411,7 @@ Int_t ShipStack::GetCurrentParentTrackNumber() const
 TParticle* ShipStack::GetParticle(Int_t trackID) const
 {
   if (trackID < 0 || trackID >= fNParticles) {
-    fLogger->Info(MESSAGE_ORIGIN, "ShipStack: Particle index %i out of range. Max=%i",trackID,fNParticles);
-    Fatal("ShipStack::GetParticle", "Index out of range ");
+    LOG(INFO) << "ShipStack: Particle index %i out of range. Max=%i",trackID,fNParticles;
```
olantwin, line 414: Is it intentional that the `Fatal` line was removed here?

ThomasRuf: yes. Looked to me strange to have an info message followed by fatal.

olantwin: Yes, it looks very strange...
Maybe we should make it an ERROR instead of INFO though? Not sure whether it makes sense to exit though...

ThomasRuf:
changed it to fatal. Don't remember that this error turned up.

PR #371, opened 2020-12-08
Fix LOG calls when printf-style syntax is used

olantwin:
The renaming left a lot of broken log statements which were using `printf`-style syntax (i.e. `%s`, `%d, `%i` etc.)

### DIFF
--- a/shipdata/ShipStack.cxx
+++ b/shipdata/ShipStack.cxx
@@ -175,7 +175,7 @@ TParticle* ShipStack::PopPrimaryForTracking(Int_t iPrim)
 
   // Test for index
   if (iPrim < 0 || iPrim >= fNPrimaries) {
-    LOG(FATAL) << "ShipStack: Primary index out of range! %i ", iPrim;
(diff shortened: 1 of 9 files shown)


shipdata/ShipStack.cxx
```diff
@@ -199,7 +199,7 @@ TParticle* ShipStack::GetCurrentTrack() const
 {
   TParticle* currentPart = GetParticle(fCurrentTrack);
   if ( ! currentPart) {
-    LOG(WARNING) << "ShipStack: Current track not found in stack!";
+    LOG(WARNING << "ShipStack: Current track not found in stack!";
```
antonioiuliano2, line 202: Does this line miss a parenthesis?

antonioiuliano2 (commented):
Moved my review here.

shipdata/ShipStack.cxx
olantwin: Yes. I guess the compiler didn't catch it because it's a macro...

olantwin: Addressed in #372

PR #378, opened 2020-12-22
Adding a generic interface to Genie functions

antonioiuliano2:
Dear all,

this pull request follows an issue discussed in sndsw:

https://github.com/SND-LHC/sndsw/issues/34,

where I was suggested to create an interface to GENIE, used both from the FairShip and sndsw scripts. This is the first (minimal) version.

The sndsw script will be added with a second pull request directly to the sndsw's macro folder, but you can see a preview in my personal macro repository:

https://github.com/antonioiuliano2/macros-snd/blob/master/makeSNDGenieEvents.py

Please let me know if you have any question or suggestion to improve the code.

Best regards,
Antonio

### DIFF
--- a/macro/makeGenieEvents.py
+++ b/macro/makeGenieEvents.py
@@ -105,7 +105,7 @@ os.chdir(work_dir)
 def makeSplines():
  '''first step, make cross section splines if not exist'''
  nupdglist = [16,-16,14,-14,12,-12]
- genie_interface.make_splines(nupdglist, targetcode, 400, nknots = 500, outputfile = work_dir + "/xsec_splines.xml")
+ genie_interface.make_splines(nupdglist, targetcode, 400, nknots = 500, outputfile = "xsec_splines.xml")
 
 def makeEvents(nevents = 100):
  run = 11
--- a/macro/makeGenieEvents.py
+++ b/macro/makeGenieEvents.py
@@ -6,6 +6,7 @@ import shipunit as u
 import shipRoot_conf
 import argparse
 import logging
+import FairShipGenieInterface as genieinterface
 shipRoot_conf.configure()
 
 
@@ -20,10 +21,10 @@ xsec = "gxspl-FNAL-nuSHiP-minimal.xml"# new adapted splines from Genie site
 hfile = "pythia8_Geant4_1.0_withCharm_nu.root" #2018 background generation
 #xsec = "Nu_splines.xml"
 #hfile = "pythia8_Geant4-withCharm_onlyNeutrinos.root"
-#put the above files in folder linked by the environment variables GENIEXSECPATH and NUFLUXPATH
 
-splines   = '/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines'+'/'+xsec #path of splines
-neutrinos = '/eos/experiment/ship/data/Mbias/background-prod-2018'+'/'+hfile #path of flux
+
+defaultsplinedir   = '/eos/experiment/ship/user/aiuliano/GENIE_FNAL_nu_splines' #path of splines
+defaultfiledir  = '/eos/experiment/ship/data/Mbias/background-prod-2018' #path of flux
 
 
 
@@ -36,9 +37,11 @@ def get_arguments(): #available options
       description='Run GENIE neutrino" simulation')
   ap.add_argument('-s', '--seed', type=int, dest='seed', default=65539) #default seed in $GENIE/src/Conventions/Controls.h
   ap.add_argument('-o','--output'    , type=str, help="output directory", dest='work_dir', default=None)
+  ap.add_argument('-f','--filedir', type=str, help="directory with neutrino fluxes", dest='filedir', default=defaultfiledir)
+  ap.add_argument('-c','--crosssectiondir', type=str, help="directory with neutrino splines crosssection", dest='splinedir', default=defaultsplinedir)
   ap.add_argument('-t', '--target', type=str, help="target material", dest='target', default='iron')
   ap.add_argument('-n', '--nevents', type=int, help="number of events", dest='nevents', default=100)
-  ap.add_argument('-e', '--event-generator-list', type=str, help="event generator list", dest='evtype', default='ALL') # Possbile evtypes: CC, CCDIS, CCQE, CharmCCDIS, RES, CCRES, see other evtypes in $GENIE/config/EventGeneratorListAssembler.xml
+  ap.add_argument('-e', '--event-generator-list', type=str, help="event generator list", dest='evtype', default=None) # Possbile evtypes: CC, CCDIS, CCQE, CharmCCDIS, RES, CCRES, see other evtypes in $GENIE/config/EventGeneratorListAssembler.xml
   ap.add_argument("--nudet", dest="nudet", help="option for neutrino detector", required=False, action="store_true")
   args = ap.parse_args()
   return args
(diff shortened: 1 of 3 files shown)


olantwin:
Looking through the commits in detail now, but one thing I noticed at first glance: Can we maybe give it a shorter name than `FairShipGenieInterface`?

How about `genie_interface`?

Also for the functions in the interface, I would stick to one naming convention. Usually `snake_case` is preferred in [python](https://namingconvention.org/python/)

olantwin:
* I think we should be very careful about the `GetXDFluxName` functions, as they take an integer and return a string. This will probably cause plenty of avoidable problems.

* Generally it is recommended to use the `subprocess` module instead of `os.system`. See e.g. [here](https://docs.python.org/3/library/subprocess.html#replacing-os-system)

olantwin:
Looks good to me otherwise!

I'd maybe run [black](https://github.com/psf/black) over the interface to make sure the formatting is consistent. While old python files will probably need to be reformatted incrementally, new files should follow a consistent style in my opinion, and black`s style seems to be becoming the de-facto standard.

antonioiuliano2:
Thank you very much for your comments. I appreciate the link to the black script, since I really need to use some automatic tool to check the formatting style of my codes.

The only comment I do not understand is this one:

> * I think we should be very careful about the `GetXDFluxName` functions, as they take an integer and return a string. This will probably cause plenty of avoidable problems.

This function refers to the naming convention in the 1D and 2D histograms in the flux files 1012, 1014, etc. These names are the same for  both sndsw and FairShip fluxes histograms.
How would you suggest to replace this function? Or would you suggest to remove it entirely and pass directly the names from the user macros?

olantwin:
> Thank you very much for your comments. I appreciate the link to the black script, since I really need to use some automatic tool to check the formatting style of my codes.

In fact, I will try to add the checking part of black to the sndsw repository soon. It's a very nice tool :)
 
> The only comment I do not understand is this one:
> 
> > * I think we should be very careful about the `GetXDFluxName` functions, as they take an integer and return a string. This will probably cause plenty of avoidable problems.
> 
> This function refers to the naming convention in the 1D and 2D histograms in the flux files "1012", "1014", etc. These names are the same for both sndsw and FairShip fluxes histograms.
> How would you suggest to replace this function? Or would you suggest to remove it entirely and pass directly the names from the user macros?

I'm not sure. i think it's probably fine as is, especially as it does mention "name" in the function name, so it's not that surprising that it returns a string.

antonioiuliano2:
Good morning Oliver,

do you have any further comment on this pull request? I should have implemented all the modifications you suggested (except the GetFluxName one).

Best regards,
Antonio

antonioiuliano2:
Re-checking the code, I just noticed that makeGenieEvents.py has a makeSplines() function, but there is not an option to call it, without changing the macro by hand. 
Since making splines is a necessary first step when producing Genie events in a new target configuration (tunsgten instead of lead, for example), I assume it is better to add a --MS option to make the splines, instead of the Genie events.

I added a final commit with --MS option, a pure tungsten target option, and the addition of tau neutrinos in the list of neutrinos.

Please let me know your opinions, according to your experience.

macro/makeGenieEvents.py
```diff
@@ -57,6 +58,7 @@ def get_arguments(): #available options
 nudet = args.nudet
 splines = args.splinedir+'/'+xsec #path of splines
 neutrinos = args.filedir+'/'+hfile #path of flux
+MakeSpline = args.MakeSpline
```
olantwin, line 61: Why not just use args.MakeSpline later?

macro/makeGenieEvents.py
```diff
@@ -44,6 +44,7 @@ def get_arguments(): #available options
   ap.add_argument('-n', '--nevents', type=int, help="number of events", dest='nevents', default=100)
   ap.add_argument('-e', '--event-generator-list', type=str, help="event generator list", dest='evtype', default=None) # Possbile evtypes: CC, CCDIS, CCQE, CharmCCDIS, RES, CCRES, see other evtypes in $GENIE/config/EventGeneratorListAssembler.xml
   ap.add_argument("--nudet", dest="nudet", help="option for neutrino detector", required=False, action="store_true")
+  ap.add_argument('--MS', help="make a new spline file", dest='MakeSpline', action="store_true")
```
olantwin, line 47: I would make make the option longer easier to understand, e.g. `--make_spline` or something like that. If you want a short option as well, that could be e.g. `-s`.

olantwin:
Good idea to add an option to make the splines. Do the other command line options make sense in that case? If not, it might make sense to define makespline exclusive to other options, or make it a standalone small script, if there is little overlap.

olantwin:
> Good morning Oliver,
> 
> do you have any further comment on this pull request? I should have implemented all the modifications you suggested (except the GetFluxName one).
> 
> Best regards,
> Antonio

Your previous changes all look good to me. I can't approve or merge in this repository, but would if I could.

antonioiuliano2:
> Good idea to add an option to make the splines. Do the other command line options make sense in that case? If not, it might make sense to define makespline exclusive to other options, or make it a standalone small script, if there is little overlap.

I would consider as needed only target and output folder options.

macro/makeGenieEvents.py
antonioiuliano2: Ok, I have renamed the option as suggested. Since -s is already used for the seed, as a short alternative I have set -ms.
Thank you for your comment.

ThomasRuf (commented):
Just a general remark. Instead of adding a for each command a new argument, another option would be to have one argument -c --command, which takes as input makeEvent, makeSplines, ...

olantwin:
Alternatively one could also use [subcommands](https://docs.python.org/3/library/argparse.html#sub-commands). I think this would also allow only allowing certain other flags that make sense and provide helpful info to `--help` automatically.

antonioiuliano2:
Dear all, I have separated the functions as suggested by Oliver (@olantwin) with subcommands. Since you also mentioned the help message, I report what I get by calling it:

$python FairShip/macro/makeGenieEvents.py -h
usage: makeGenieEvents.py [-h] {sim,spline} ...

Run GENIE neutrino" simulation

positional arguments:
  {sim,spline}
    sim         make genie simulation file
    spline      make a new cross section spline file

optional arguments:
  -h, --help    show this help message and exit

macro/makeGenieEvents.py
```diff
@@ -134,9 +107,35 @@ def addHists():
   genie_interface.add_hists(neutrinos, "genie-"+sDict[p]+".root", p)
   os.chdir('../')
 
-if (MakeSpline):
+if ("splinedir" not in args):
```
olantwin, line 110: I don't understand this: `splinedir` should always be in args, as it has a default value. Or is it undefined, if the subparser is not chosen?
Is there no better way to check whether we are in the spline or sim subparser?

olantwin:
> Dear all, I have separated the functions as suggested by Oliver (@olantwin) with subcommands. Since you also mentioned the help message, I report what I get by calling it:
> 
> $python FairShip/macro/makeGenieEvents.py -h
> usage: makeGenieEvents.py [-h] {sim,spline} ...
> 
> Run GENIE neutrino" simulation
> 
> positional arguments:
> {sim,spline}
> sim make genie simulation file
> spline make a new cross section spline file
> 
> optional arguments:
> -h, --help show this help message and exit

Thank you, this looks very nice!

macro/makeGenieEvents.py
antonioiuliano2: If the subparser is not chosen, all its args are not defined. I confirmed it by inspecting the args namespace when the two different options are called.

An alternative is to save the subparser in args itself, as discussed here:
https://stackoverflow.com/questions/8250010/argparse-identify-which-subparser-was-used/9286586#9286586

However, this would lead back to the previous case, with an if(args.comment=="spline") statement (equivalent to if(MakeSpline)).

olantwin: Thank you for the explanation!  I agree that your check is probably the best way to do this :)

PR #383, opened 2021-02-16
Import TTCluster code by A. Petrov

olantwin:
As this code should be useful to FairShip and SNDSW, I suggest we merge it here and make SND-specific adjustments in SNDSW.

I am currently studying the code and Alex's presentation to understand what is implemented and what is left to do.

### DIFF
--- a/python/TTCluster.py
+++ b/python/TTCluster.py
@@ -1,0 +1,784 @@
+#!/usr/bin/python
+
+import ROOT
+import numpy as np
+import math
+from math import exp, sqrt, log, ceil
+import random
+import matplotlib.pyplot as plt
+
+from ROOT.TMath import Landau, Gaus, Poisson
+from ShipGeoConfig import ConfigRegistry
+
+#####################################################
+##########  GLOBAL VALUES AND PARAMETERS ############
+#####################################################
+
+# Position of SiPMs in the detector (left/right and up/down)
+sipm_hor_pos = +1 # at X = +len_vert / 2 
+sipm_vert_pos = -1 # at Y = -len_hor /  2
+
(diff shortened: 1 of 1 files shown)


ThomasRuf:
should it be placed in the nutaudet folder rather than in the python folder until the logic can be implemented in shipdigi and shipreco?

olantwin:
Hm, I think generally it would be best to have all/most python modules in the python folder, so that imports etc. are straightforward, unless they are scripts which one would not want to import.

Also in light of SND@LHC, in which case this code would not be used for the nutaudet class.

ThomasRuf:
yes, but this file is not yet ready to be imported anywhere. Does it produce digits? Are digits == cluster? Or are clusters ensemble of digits? How are the clusters saved to disk?

olantwin:
> yes, but this file is not yet ready to be imported anywhere.

Maybe not yet, but I think there is little point in moving it back and forth.

> Does it produce digits? Are digits == cluster? Or are clusters ensemble of digits? 

The code generates clusters, but as part of this seems to simulate the light-yield per channel. I am still digging through the code, but I think it should be possible to also make these per-channel digitised hits available, without too much additional effort, as the information is internally available.

> How are the clusters saved to disk?

This is the responsibility of the code using the module. In the script there is a usage example, creating and plotting information from the

olantwin:
For some additional information, [here the slides Alex shared with me](https://github.com/ShipSoft/FairShip/files/5994913/TT.in.FairShip.copy.pptx).

One thing I notice on slide 10 is that the substructure of the light yield isn't reproduced. I wonder whether this is an artifact of the readout via the SiPMs.

PR #386, opened 2021-04-16
Fix HNL lifetime, which was uninitialised if debug==false

olantwin:
Bug was introduced in 20e9d10c548dec5430f914c9fc34e9cca9d84bdc


ThomasRuf:
Well done.

PR #388, opened 2021-06-22
Nu tau pythia8

ThomasRuf:
Can go in testing mode.


ThomasRuf:
I am hesitating to merge, because it tells me there are 79 files changed, while I was expecting to see only one file, macro/runPythia8PP.py

ThomasRuf:
OK, understand. For some reason I am in FairShip and not in sndsw

PR #394, opened 2022-01-24
no need to import XRootD when on lxplus 

siilieva:


siilieva:
Created pull request to wrong master.. Closing the request now.

PR #400, opened 2022-05-13
Updating package list for ubuntu 20

antonioiuliano2:
Dear FairShip/SNDSW users,

Together with Simona Ilieva, we checked the package requirements list in Ubuntu 20.04 LTS

Most of those from Ubuntu 18 still works, with two exceptions:
* linuxbrew-wrapper
* metakernel

Also added some packages:
* future (required by some python imports)
* python3-tk (requierd by the Event Display)

Best regards,
Antonio Iuliano


ThomasRuf (approved):
OK

PR #402, opened 2022-06-08
Fix deprecation warning

olantwin:


olantwin:
Updated to fix them in eventDisplay.py as well.

ThomasRuf (commented):
OK

PR #405, opened 2022-06-16
IgnoreCommandLIneOptions no longer needed

olantwin:
On my request the default value in PyROOT has changed to the value we were setting (already a while ago, so all ROOT versions we use should have this fix).

See https://root.cern/manual/python/#configuration-options

Fixes olantwin/FairShip#51


ThomasRuf (approved):
OK

PR #406, opened 2022-06-16
Don't import getopt when it's not used

olantwin:
To my surprise I saw that run_simScript still imported getopt, so I did
a quick search and remove for the other locations we import but don't
use it.

Fixes olantwin/FairShip#53


ThomasRuf (approved):
OK

PR #412, opened 2022-08-30
Fix ECN3 dimensions and raise beamline (→lower TCC8 and ECN3) by 50 cm

olantwin:


ThomasRuf:
Is the region between T1 and T4 still with air? It would limit the momentum reconstruction.

olantwin:
> Is the region between T1 and T4 still with air? It would limit the momentum reconstruction.

Yes, at least looking at the event display.

PR #419, opened 2023-01-19
Shield geo fix

PlaZMaD:
2 new methods added to macro/ShipReco.py:
**generate_shield_rootfile** - create analog of "combi.root" with parameters provided
**generate_shield_rootfile** - generate array with parameters from opened geofile with FULL geometry

macro/ShipReco.py was adapted to work with it:
-If muShieldDesign == 8, then tmp "combi.root"  (not exactly that name, but still you should understand the idea) is created from the full geometry file
-the "ShipGeo.muShieldGeo" option is overwritten by our tmp file path
-tmp file is deleted in the end

### DIFF
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -195,4 +195,6 @@ for global_variables.iEvent in range(options.firstEvent, options.nEvents):
  # memory monitoring
  # mem_monitor()
 # end loop over events
+if ShipGeo.muShieldGeo==8:
+  os.system("rm  tmp_muShieldGeo.root")
 SHiP.finish()
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -104,6 +104,8 @@ run.SetOutputFile(ROOT.TMemFile('output', 'recreate'))  # Output file
 run.SetUserConfig("g4Config_basic.C") # geant4 transport not used, only needed for creating VMC field
 rtdb = run.GetRuntimeDb()
 # -----Create geometry----------------------------------------------
+import numpy as np
+from array import array
 def generate_shield_rootfile(geofile, params):
   f = ROOT.TFile.Open(geofile, 'recreate')
   params =  np.array(params, dtype=float)
@@ -126,10 +128,10 @@ def generate_shield_params_from_geofile(lfGeo):
   for node in cave_node.GetNodes():
       for subnode in node.GetNodes():
           nodeName = subnode.GetName()
-          if "MagRetR" in nodeName and 'Absorb' not in nodeName: 
+          if "MagRetR" in nodeName and 'Absorb' not in nodeName and "7" not in nodeName: 
               lVol =  subnode.GetVolume().GetShape()
               dots = lVol.GetVertices()
-              dZ.append(lVol.GetDZ())
+              dZ.append(lVol.GetDZ() + 5)
               vetz = [dots[i] for i in range(16)]
               Y = vetz[1::2]
               X = np.absolute(vetz[::2])
@@ -138,8 +140,8 @@ def generate_shield_params_from_geofile(lfGeo):
               x_out.append(np.max(X[4:]) - np.min(X[4:]))
               gap_in.append(np.min(X[:4]) - x_in[-1])
               gap_out.append(np.min(X[4:]) - x_out[-1])
-              y_in.append(max(Y[:4]) - x_in[-1])
-              y_out.append(max(Y[4:]) - x_out[-1])
+              y_in.append(max(Y[:4]) - x_in[-1] + 0.1)
+              y_out.append(max(Y[4:]) - x_out[-1] + 0.1)
 
   out_params = dZ
   for i in range(8):
@@ -151,8 +153,8 @@ def generate_shield_params_from_geofile(lfGeo):
     out_params.append(gap_out[i])
   return out_params
 
-if ShipGeo.muShieldGeo==8:
-  shiedlParams = generate_shield_options_from_geo(fgeo)
+if ShipGeo.muShieldDesign==8:
+  shiedlParams = generate_shield_params_from_geofile(fgeo)
   generate_shield_rootfile('tmp_muShieldGeo.root', shiedlParams)
   ShipGeo.muShieldGeo = 'tmp_muShieldGeo.root'
 
@@ -195,6 +197,6 @@ for global_variables.iEvent in range(options.firstEvent, options.nEvents):
  # memory monitoring
  # mem_monitor()
 # end loop over events
-if ShipGeo.muShieldGeo==8:
+if ShipGeo.muShieldDesign==8:
   os.system("rm  tmp_muShieldGeo.root")
 SHiP.finish()
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -104,6 +104,58 @@ run.SetOutputFile(ROOT.TMemFile('output', 'recreate'))  # Output file
 run.SetUserConfig("g4Config_basic.C") # geant4 transport not used, only needed for creating VMC field
 rtdb = run.GetRuntimeDb()
 # -----Create geometry----------------------------------------------
+def generate_shield_rootfile(geofile, params):
+  f = ROOT.TFile.Open(geofile, 'recreate')
+  params =  np.array(params, dtype=float)
+  parray = ROOT.TVectorD(len(params), array('d',params))
+  parray.Write('params')
+  f.Close()
+  print('Geofile constructed at ' + geofile)
+  return geofile
+
+def generate_shield_params_from_geofile(lfGeo):
+  geo = lfGeo.Get("FAIRGeom")
+  cave_node = geo.GetTopNode()
+  x_in = [40, 80]
+  x_out = [40, 80]
+  y_in = [150, 150]
+  y_out = [150, 150]
+  gap_in = [2, 2]
+  gap_out = [2, 2]
+  dZ = [70, 170]
+  for node in cave_node.GetNodes():
+      for subnode in node.GetNodes():
+          nodeName = subnode.GetName()
+          if "MagRetR" in nodeName and 'Absorb' not in nodeName: 
+              lVol =  subnode.GetVolume().GetShape()
+              dots = lVol.GetVertices()
+              dZ.append(lVol.GetDZ())
+              vetz = [dots[i] for i in range(16)]
+              Y = vetz[1::2]
+              X = np.absolute(vetz[::2])
+
+              x_in.append(np.max(X[:4]) - np.min(X[:4]))
+              x_out.append(np.max(X[4:]) - np.min(X[4:]))
+              gap_in.append(np.min(X[:4]) - x_in[-1])
+              gap_out.append(np.min(X[4:]) - x_out[-1])
+              y_in.append(max(Y[:4]) - x_in[-1])
+              y_out.append(max(Y[4:]) - x_out[-1])
+
+  out_params = dZ
+  for i in range(8):
+    out_params.append(x_in[i])
+    out_params.append(x_out[i])
+    out_params.append(y_in[i])
+    out_params.append(y_out[i])
+    out_params.append(gap_in[i])
+    out_params.append(gap_out[i])
+  return out_params
+
+if ShipGeo.muShieldGeo==8:
+  shiedlParams = generate_shield_options_from_geo(fgeo)
+  generate_shield_rootfile('tmp_muShieldGeo.root', shiedlParams)
+  ShipGeo.muShieldGeo = 'tmp_muShieldGeo.root'
+
 modules = shipDet_conf.configure(run,ShipGeo)
 # run.Init()
 fgeo.FAIRGeom


olantwin (changes requested):
While not very elegant, this seems to be a reasonable fix to the issue at hand.

As long as "RECREATE" is replace with "NEW"/"CREATE", this looks OK to me. See [here](https://root.cern.ch/doc/master/classTFile.html#ad0377adf2f3d88da1a1f77256a140d60).

ThomasRuf (commented):
These lines need to be copied to all other programs which reconstruct / analyze data relying on the geometry, like ShipAna, but of course all the user code. If people insist of this change, I can approve it. I have the feeling that having this temporary root file is trying to solve a problem which is really addressed in sndsw, see https://indico.cern.ch/event/1151104/contributions/4869659/attachments/2448082/4195016/Ship-Soft-CollaborationMeetingMay2022.pdf, slide 3.

olantwin:
The root file is definitely a long-term solution, but as a fix to allow reconstructing simulations which were created with muon shield 8, this seems reasonable.

> These lines need to be copied to all other programs which reconstruct / analyze data relying on the geometry, like ShipAna, but > of course all the user code. If people insist of this change, I can approve it

Maybe we could deduplicate this code and only define this once to be imported in all files that need to construct the geometry?
We have a lot of identical code across our python macros, which always risks getting out of sync. If we only define this procedure once, it's also going to be easier to change it in the future, when we e.g. remove this root file altogether.

ThomasRuf:
What about a fix which does not require any change to ShipReco, ShipAna, EventDisplay, and the user code?
In shipdet_conf, we have 
elif ship_geo.muShieldDesign == 8:
  MuonShield = ROOT.ShipMuonShield(ship_geo.muShieldGeo,...

Requiring that the geofile and the ship_geo.muShieldGeo have to be in the same directory, EOS or local doesn't matter, we just prepend the directory of the geofile, MuonShield = ROOT.ShipMuonShield(pathToGeo+ship_geo.muShieldGeo,...

By the way, what is first, the combi.root file or run_simScript.py?

olantwin:
> What about a fix which does not require any change to ShipReco, ShipAna, EventDisplay, and the user code? In shipdet_conf, we have elif ship_geo.muShieldDesign == 8: MuonShield = ROOT.ShipMuonShield(ship_geo.muShieldGeo,...
> 
> Requiring that the geofile and the ship_geo.muShieldGeo have to be in the same directory, EOS or local doesn't matter, we just prepend the directory of the geofile, MuonShield = ROOT.ShipMuonShield(pathToGeo+ship_geo.muShieldGeo,...

@PlaZMaD what do you think?

In fact, this is why I never noticed this issue before this discussion, as I always had these input files locally in the working directory.
 
> By the way, what is first, the combi.root file or run_simScript.py?

combi.root is an input to run_simScript.py.

ThomasRuf:
and the code which makes combi.root, it exists in the FairShip distribution? If not, it definitely needs to be added.

olantwin:
A muon shield configuration right now is a ROOT file which contains a single TVectorD called "params" with ~50 parameters (@PlaZMaD could you please confirm the number of parameters).

This can be created any way one wants, and is automatically created in the optimisation. Combi.root is just one of these configurations with the numbers corresponding to the Combi shield.

olantwin:
A simple snippet to create a geofile:

```python
from array import array
import logging
import ROOT as r


def generate_geo(geofile, params):
    f = r.TFile.Open(geofile, 'recreate')
    parray = r.TVectorD(len(params), array('d', params))
    parray.Write('params')
    f.Close()
    logging.info('Geofile constructed at ' + geofile)
    return geofile
```

PlaZMaD:
> @PlaZMaD what do you think?

We don't have a good way to get the information on geofile path in shipdet_conf. So either we need to modify the user code, or have external  agreement on what the path should look  (like copy it to working directory and have a well-defined naming). Both options are so so (+ we need to open geofile while it's already opened in case of putting code to shipdet_conf). Can we use some kind of global ROOT variables to get list of opened files or smthg?

olantwin:
> We need (1 length + 2 x sizes + 2 y sizes + 2 gap sizes) x N_magnets = 8 x 7 = 56 parameters. Some of therm are not even used, but still.

Yes, the 14 parameters referring to the hadron absorber magnets are ignored. This is because they were taken out of the scope of the optimisation and hardcoded instead.

ThomasRuf:
For Oliver: this addition of params to the geofile, is this done by run_simScript?

For PlaZMaD: Yes, all these is available via ROOT.gROOT, GetListOfFiles, ... and many other thing. There is also a global space.

essentially, your problem is that you don't know how to insert a TVectorD into the muon shield configuration except of reading it via a persistent ROOT file?

ThomasRuf:
there is no problem of getting info about geofile in shipdet_conf

PlaZMaD:
> 



> essentially, your problem is that you don't know how to insert a TVectorD into the muon shield configuration except of reading it via a persistent ROOT file?

The root file is used both in python and c++ code (yes file, not just params). In some weird way. Just none want to deal with code rework.

olantwin:
> For Oliver: this addition of params to the geofile, is this done by run_simScript?

What do you mean by addition of parameters to the geofile?

> essentially, your problem is that you don't know how to insert a TVectorD into the muon shield configuration except of reading it via a persistent ROOT file?

The problem is that instead of adding the parameters to the geofile, we only add the path to the muon shield configuration to the geofile. We could use the default muon shield instead and overwrite the parameters accordingly when doing the simulation, so that the reconstruction and analysis can just use the geofile as usual.

olantwin:
I have quite a few utilities for creating, rendering, checking etc. of muon shield configurations in [this repo](https://github.com/olantwin/muon_shield_optimisation)

olantwin:
> The problem is that instead of adding the parameters to the geofile, we only add the path to the muon shield configuration to the geofile. We could use the default muon shield instead and overwrite the parameters accordingly when doing the simulation, so that the reconstruction and analysis can just use the geofile as usual.

The one hurdle to this solution is that currently only the lengths of the muon shield magnets are exposed in the geofile. All the transverse parameters are currently set in C++ either via the ROOT file or via the hardcoded configurations 7, 9, 10.

ThomasRuf:
I am getting again confused. 
1) The dictionary in the geofile shipped with the simulated data, does it contain the correct parameters to configure the muon shield geometry?
2) Or are these parameters superseded by the parameters in the combi file? Which cannot be since you claim you can recreate the combi file with the geofile.
3) Is then the problem, that there is no interface made to the muon shield class of retrieving the parameters from the geo dictionary? Only reading from an external file is supported?

olantwin:
> I am getting again confused.
> 
>     1. The dictionary in the geofile shipped with the simulated data, does it contain the correct parameters to configure the muon shield geometry?

No.
 
>     2. Or are these parameters superseded by the parameters in the combi file? Which cannot be since you claim you can recreate the combi file with the geofile.

The geofile does only define the lengths of the muon shield, which are overwritten by the parameters.
Evgeny recreates the geofile using the ROOT ShipMuonShield object saved in `FAIRGeom`, not from the dictionary.

>     3. Is then the problem, that there is no interface made to the muon shield class of retrieving the parameters from the geo dictionary? Only reading from an external file is supported?

Yes, that's correct, and what I meant in my last comment.

ThomasRuf:
this sounds very dangerous. The philosophy of geometry_config.py, which is the source of the dictionary in the geofile, was to build SHiP step by step, knowing the end position of the previous detector element, and then adding a new detector piece, only using relative positions. If the length of the muon shield only becomes known when reading the combi.root file, I don't see how this can work anymore.

PlaZMaD:
So how about:
in  shipDet_conf right before the line "latestShipGeo = ConfigRegistry.loadpy(...)":
-check if the muShieldDesign==8 and some geofile is opened in global ROOT space and there is some kind of shield node in it ->generate tmp "combi.root" in system tmp folder and replace ship_geo.muShieldGeo path in config with it?

ThomasRuf:
Not sure if system tmp folder is a good idea and will work everywhere, for example with HTCondor. Still looks more complicated to me as necessary, but apparently you don't trust that the combi file which is shipped together with the data and the geofile is the correct one. I think ROOT has also the possibility to have a virtual TFile in memory. Anyhow, you can loop through the list of open files using gROOT.GetListOfFiles(), check if the file contains FAIRgeom, if yes, run your lines of code of making the temporary combi file. This would be transparent to the users.

PlaZMaD:
Local macro folder? I can't just see a good way to delete the tmp file in a proper time from shipDet_conf code.

olantwin:
> this sounds very dangerous. The philosophy of geometry_config.py, which is the source of the dictionary in the geofile, was to build SHiP step by step, knowing the end position of the previous detector element, and then adding a new detector piece, only using relative positions. If the length of the muon shield only becomes known when reading the combi.root file, I don't see how this can work anymore.

This system was never meant to be used outside the optimisation. At the time this strategy was chosen in order to make the minimal set of changes necessary to perform the optimisation. Exposing all the information hardcoded in ShipMuonShield.cxx for the already existing configurations to python would have been a very significant amount of work.

If this had been designed for general use, it would have been designed very differently.

olantwin:
> I think ROOT has also the possibility to have a virtual TFile in memory. 

Yes, this exists and I have used it in the past for some of the files needed by FairRoot. There are some limitations, but this might be a way forward.

ThomasRuf:
What about https://root.cern/doc/master/classTMemFile.html ? 

create the file in memory, fill it with your code as a normal root file, set muShieldGeo 

tmp =ROOT.TMemFile('combi.root','recreate')
your code

 latestShipGeo = ConfigRegistry.loadpy("$FAIRSHIP/geometry/geometry_config.py", muShieldGeo = tmp)

in geometry_config.py, instead of opening directly f = r.TFile.Open(muShieldGeo), only do it if muShieldGeo is a string, otherwise f=muShieldGeo.

when you return to shipDet_conf.py, do tmp.Close().

PlaZMaD:
Ok from gROOT in shipDet_conf I can see geofile and geometry object:
TFile: name=/eos/experiment/ship/user/edursov/combi_ecn3_spill_snd_scor_planes_fixed_fields_12012023_tracking_try_1/1/geofile_full.conical.MuonBack-TGeant4.root, title=, option=READ

OBJ: TGeoManager	FAIRGeom	FAIR geometry

That probably solves the problem of code migration to shipDet_conf.
Still I don't know if there a way to pass TMemFile to ShipMuonShield constructor "ShipMuonShield(TString geofile, ...); auto f = TFile::Open(geofile, "read");"

olantwin:
Passing a TMemFile will require changing the constructor unfortunately. Instead of passing a filename, the interface would be to pass a TFile or object exposing the same interface (such as TMemFile.

ThomasRuf:
? this python. Will accept any pointer.

ThomasRuf:
ok, this one. Just add another constructor.

ThomasRuf:
What about passing empty string, then if empty received look in gRoot for mem file.

ThomasRuf:
TMemfile can be found via gROOT:
>>> import ROOT
>>> X= ROOT.TMemFile('output', 'recreate')
>>> X
<cppyy.gbl.TMemFile object at 0x56534e436e20>
>>> ROOT.gROOT.FindObject('output')
<cppyy.gbl.TMemFile object at 0x56534e436e20>

ThomasRuf:
Studying a bit more the code, I had overlooked that this combi.root file is read twice. Once in geometry_config.py to get the parameters needed to position the other detector elements, and then again in ShipMuonShield. In case a geofile exists, it is not needed to create and load the ShipMuonShield.  Therefore, it is only required once to extract the parameters from the geofile, no need to create an tmp root file. 
By the way, was the current fix tested? I am wondering if there is an issue of creating twice the TGeomanager, once when extracting the parameters for the combi.root file, and then again in the standard processing.

ThomasRuf:
it seems the fix in [shieldGeoFixDesign8](https://github.com/ShipSoft/FairShip/tree/shieldGeoFixDesign8) works, it creates the TGeo manager twice, as expected, but ROOT does not complain.

python -i $FAIRSHIP/macro/ShipAna.py -g /eos/experiment/ship/user/kap01/ecn3/geofile_full.conical.MuonBack-TGeant4.root -f /eos/experiment/ship/user/kap01/ecn3/individual_combi_ecn3_spill_11_ship_14318450_54_rec.root
[WARN] FairRun::SetOutputFile() deprecated. Use FairRootFileSink.
read params from geofile Name: root://eospublic.cern.ch//eos/experiment/ship/user/kap01/ecn3/geofile_full.conical.MuonBack-TGeant4.root Title: 
Info in <TGeoManager::CloseGeometry>: Geometry loaded from file...
Info in <TGeoManager::SetTopVolume>: Top volume is cave. Master volume is cave
Info in <TGeoNavigator::BuildCache>: --- Maximum geometry depth set to 100
Info in <TGeoManager::Voxelize>: Voxelizing...
Info in <TGeoManager::CountLevels>: max level = 6, max placements = 2000
Info in <TGeoManager::CloseGeometry>: 8347127 nodes/ 1268 volume UID's in FAIR geometry
Info in <TGeoManager::CloseGeometry>: ----------------modeler ready----------------
[INFO] Media file used: /afs/cern.ch/user/t/trufship/ShipBuild-2022/sw/slc7_x86-64/FairShip/master-local1/geometry/media.geo
muShieldDesign 8 detected and geometry already loaded, will not instantiate mushield object
 Mag Spectro Design 3
Info in <EmulsionMagnet::Chosen TP Design (0 no, 1 yes) >:  3
Info in <Target::SetDetectorDesign>:  to 3
Info in <ecal::ecal>: Geometry is read from file ecal_rect5x10m2.geoz3615.2000000000003.geo.
Info in <ecal::ecal>: Size of cell of type 2 is 5.970000 cm.
Info in <hcal::hcal>: Geometry is read from file hcal_rect.geoz3715.2000000000003.geo.
Info in <hcal::hcal>: Number of modules is 924, lightmap none
Info in <hcal::hcal>: Size of cell of type 22 is 23.959999 cm.
Calling addVMCFields
Creating field map called MainSpecMap using /afs/cern.ch/user/t/trufship/ShipBuild-2022/sw/slc7_x86-64/FairShip/master-local1/files/MainSpectrometerField.root
ShipBFieldMap::readMapFile() creating field MainSpecMap using file /afs/cern.ch/user/t/trufship/ShipBuild-2022/sw/slc7_x86-64/FairShip/master-local1/files/MainSpectrometerField.root
x limits: -470, 470, dx = 10
y limits: -550, 550, dy = 10
z limits: -990, 990, dz = 10
Offsets: x = 0, y = 0, z = 3068
Angles : phi = 0, theta = 0, psi = 0
Total number of bins = 2098455; Nx = 95, Ny = 111, Nz = 199
Setting the Global field
Adding field MainSpecMap to Global
The virtual MC pointer gMC is null! The global field can't be used by Geant4 but will work for track fitting and track extrapolation
TGeoManager::Init:0: RuntimeWarning: Deleting previous geometry: FAIRGeom/FAIR geometry
Info in <TGeoManager::CloseGeometry>: Geometry loaded from file...
Info in <TGeoManager::SetTopVolume>: Top volume is cave. Master volume is cave
Info in <TGeoManager::Voxelize>: Voxelizing...
Info in <TGeoManager::CountLevels>: max level = 6, max placements = 2000
Info in <TGeoManager::CloseGeometry>: 8347127 nodes/ 1268 volume UID's in FAIR geometry
Info in <TGeoManager::CloseGeometry>: ----------------modeler ready----------------
-I- ecalStructure::Construct(): calorimeter matrix created.

...

ThomasRuf:
Although both proposals fix the issue for running on simulated data, for which a geofile is present, there is no clean solution for running the simulation itself. How did people managed to have neutrino and DIS simulations running? They have used a combi.root file from some unknown source?
parser.add_argument("-g",        dest="geofile",       help="geofile for muon shield geometry, for experts only", required=False, default=None)
I strongly recommend to abandon this way of injecting parameters from a root file, and go to the recommended way of putting the parameters into shipDet_conf.py. I should never have accepted a push request for this change of philosophy.

olantwin:
> How did people managed to have neutrino and DIS simulations running? They have used a combi.root file from some unknown source? 

Using the root file provided by Eduard.

> parser.add_argument("-g", dest="geofile", help="geofile for muon shield geometry, for experts only", required=False, default=None) I strongly recommend to abandon this way of injecting parameters from a root file, and go to the recommended way of putting the parameters into shipDet_conf.py. I should never have accepted a push request for this change of philosophy.

As I've now repeated several times (and is also written in the help message you quote), this was never meant for general use. Unfortunately with ECN3 
it was suddenly necessary due to the rapidly changing muon shield configurations to expose more users to it.

This is not a change in philosophy, We all agree that this is an ugly method to do things, but as written in my original email, it was the least bad solution for an extremely narrow use case at the time. Nobody advocates to use this and everyone agrees that this should be replaced as soon as possible.

I only ever made a pull request for this so that the code used for the optimisation was accessible to everyone and no special branch was necessary for the optimisation.

ThomasRuf:
ok. I am worried by the mail of Richard today about revisiting SHiPs layout at SND@ECN3, maybe another muon shield change ahead. Who will implement the parameters in the config file?

olantwin:
Let's see what he asks and find a pragmatic solution then. I fear it will be a bigger change than just a new muon shield.

olantwin:
@PlaZMaD From your side, is this ready to be merged? Or do we want to revisit this solution with the new SC options on the horizon?

PlaZMaD:
> @PlaZMaD From your side, is this ready to be merged? Or do we want to revisit this solution with the new SC options on the horizon?

Well... It works fine for a current problem, won't be actual after the shield upcoming rework, but I'm not sure we can keep backward compatibility to work with current geofiles after the rework either. Maybe consider a temp branch for analysis purposes and not merge to master?

PR #420, opened 2023-03-09
ECN3 v2023

olantwin:
Changes needed for the new ECN3 geometry.

See here for dimensions of old geometry:
https://indico.cern.ch/event/1138547/contributions/4777211/attachments/2416232/4134679/2022_03_28_SHiP_LLO_study.pdf

And here for new dimensions:
https://indico.cern.ch/event/1256534/contributions/5285347/attachments/2607225/4503711/2023_03_08_SHiP_StatusPlans.pdf

Rescaling:
- [x] TimeDet
- [x] UBT
- [x] Calorimeter
- [x] Vessel
- [x] Lengths
- [x] SST
- [x] Muon

Other changes:
- [x] Update field map (https://github.com/olantwin/FairShip/pull/62)
- [x] Make new muon shield the default

Testing:
- [x] builds
- [x] Confirm beam height using measurements
- [x] visual
- [x] overlaps
- [ ] fields
- [ ] physics


olantwin:
@matclim, could you please check whether the scaling of the SplitCal looks sensible?
Do you happen to know whether the hcal geometry file is still used, or is that change superflous?

olantwin:
@antonioiuliano2 I did not adjust anything for the SND. I won't touch it so that you can do whatever you like.

olantwin:
image
image

Clearly the naive scaling was only partially successful.

It seems the calorimeters and magnet in particular need more work, but it seems everything else also slightly collides with the floor. Maybe an issue with the beamline height?

olantwin:
image

Despite beam adjustment, there are conflicts with the cavern.

matclim:
@olantwin the actual Splitcal layout would vary depending on the used layer (horizontal or vertical) as a baseline option for 6x4m^2. It would be 4x6.12 m^2 for horizontal layers and 4.08x6m^2 for vertical layers.
layout_std

I have myself have never used the hcal (as I'm focused on the SplitCal). In principle the SplitCal can also do hadronic calorimetry but you'd need extra layers, beyond the 20 X0 which are currently accounted for (a rough estimate implies ~15 times more lead than for just the eSplitCal). As such I think it's probably safe to just keep the hcal as it is for now.

olantwin:
@matclim Do you know whether the SplitCal should be in the simulation as the baseline? I just noticed that I have ECal and HCal in the render of the geometry.

ThomasRuf:
The problem with SplitCal, there is no reconstruction code for it. So, you can simulate it, but you will have no reco photons ore electrons to work with for reconstructing a signal.

olantwin:
Thanks for clarifying this, Thomas!

ThomasRuf:
In my mind, SplitCal still needs a proof of concept. It is an idea, but was never fully simulated, and the performances evaluated. This should be done outside of the default FairShip. For the moment, I think there is only the Shashlik calo option possible.

olantwin:
@leoredi pointed out that the second link was incorrect. This has been fixed.

olantwin:
Latest numbers from Richard

* UBT sensitive area: 1.20 x 2.90  (don't know what they have implemented as mechanics to hold the chambers but at least this makes the RPC overlap with the SBT
* Decay volume upstream (inner/outer): 1.00 x 2.70 / 1.40 x 3.10 (SBT thickness ~20cm)
* Decay volume downstream at 50m (inner/outer): 4.20 x 6.20 / 4.60 x 6.60 (SBT thickness ~20cm all  the way)
* Straw tracker sensitive area (all stations): 4.0 x 6.0 (don't know how much space mechanics takes in simulations)
* Ignore vacuum chamber through magnet
* Magnet 4x6m free aperture (inner/outer dimensions of yoke): 4.40 x 7.00 / 6.00 x 8.60 (the larger free vertical space around aperture is for coils)
* Timing detector sensitive area: 4.50 x 6.50 (don't know how much was reserved for mechanics in simulation)
* ECAL/MUON sensitiver area 5.00 x 6.50.

antonioiuliano2 (approved):
Thank you Oliver, it seems everything fine to me :)

fairlyons (changes requested):
geometry_config.py
135:          c.Veto.sensitiveThickness = 0.2 * u.m
131 - 134: can probably be deleted
138:          c.Veto.rib = 1.5 * u.cm

veto.cxx
95:     f_RibThickness(1.5 * cm),
187:   double wx1 =  920 * mm;
188:   double wx2 = 4120 * mm;
197:   double wy1 = 2620 * mm;
198:   double wy2 = 6120 * mm;
829:   double ribThick = 15*mm;
lid thickness seems to be off still, I have not looked closely at this

olantwin:
> 131 - 134: can probably be deleted

I will probably make a massive PR soon to remove all the TP era code. Until then I'll leave this.

Implemented all of your other changes. Were these updated numbers from Andrea or did I make a mistake while importing your earlier changes?

For the lid, we can discuss at the ECN3 meeting. I think it would be good to have at least the correct material budget for the front face. For the back lid, it matters less.

ThomasRuf:
the back window also matters, the material will affect the EM performance.

olantwin:
> the back window also matters, the material will affect the EM performance.

While I generally agree, none of the studies which are going to be launched imminently use detectors after the back window. I'll raise the issue today at the meeting so that we can take a decision.

fairlyons:
> Implemented all of your other changes. Were these updated numbers from Andrea or did I make a mistake while importing your earlier changes?

These are new updates from Andrea, we had a clarifying discussion on Friday. He suggested to keep the outer dimensions as specified so the inner has to be reduced to account for wall thickness.

olantwin:
Action items from the ECN3 meeting:

- [ ] Check material budget
- [ ] Adjust thickness and material of entrance window
- [ ] Check material around the tracking stations

olantwin:
As discussed during today's ECN3 meeting, this is now ready for pilot productions. Merging.

PR #421, opened 2023-03-16
SND Geometry for Muon Magnetic Spectrometer studies in SHiP

antonioiuliano2:
Dear All,

This is the geometry I showed at the Collaboration Meeting, which I employ for the simulation of performace study of magnetic spectrometer.

This introduces a NuTauTargetDesign version 4, which will be the design version for the studies in ECN3 from now on.
The main changes are basically two:

- The target is now simply a sequence of Emulsion/Tungsten Wall and Target Trackers;
- The muon filter is replaced with a magnetic spectrometer, with for now an uniform magnetic field of 3 T.

Please give a look at the code and give me advice, if possible. I have tried to clean it in the last days, but it still is quite messy (especially due to the large difference of the muon detector volume with respect to the previous configuration).

SND_ECN3_layout_magspectro_studies

Best Regards,
Antonio

### DIFF
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -713,19 +713,19 @@ with ConfigRegistry.register_config("basic") as c:
         c.NuTauTarget.col = 2
         c.NuTauTarget.wall = 13
         c.NuTauTarget.n_plates = 59
-        c.NuTauTarget.EmX = 19.2 * u.cm
-        c.NuTauTarget.EmY = 19.2 * u.cm
+        c.NuTauTarget.EmX = 20. * u.cm
+        c.NuTauTarget.EmY = 20. * u.cm
         c.NuTauTarget.BrPackX = 2 * 0.05 * u.cm
         c.NuTauTarget.BrPackY = 2 * 0.05 * u.cm
-        c.NuTauTarget.BrPackZ = 0
+        c.NuTauTarget.BrPackZ = 2.21 * u.cm
         c.NuTauTarget.BrX = c.NuTauTarget.BrPackX + c.NuTauTarget.EmX
         c.NuTauTarget.BrY = c.NuTauTarget.BrPackY + c.NuTauTarget.EmY
 
         c.NuTauTarget.EPlW = 2* c.NuTauTarget.EmTh + c.NuTauTarget.PBTh
         c.NuTauTarget.AllPW = c.NuTauTarget.LeadTh + c.NuTauTarget.EPlW
 
-        #c.NuTauTarget.BrZ = c.NuTauTarget.n_plates * c.NuTauTarget.AllPW + c.NuTauTarget.EPlW + c.NuTauTarget.BrPackZ
-        c.NuTauTarget.BrZ = 10. * u.cm #hardcoded wall width
+        c.NuTauTarget.BrZ = c.NuTauTarget.n_plates * c.NuTauTarget.AllPW + c.NuTauTarget.EPlW + c.NuTauTarget.BrPackZ #10 cm
+        #c.NuTauTarget.BrZ = 10. * u.cm #hardcoded wall width
         c.NuTauTarget.SingleEmFilm = False
 
  #TargetTrackers!
@@ -757,12 +757,12 @@ with ConfigRegistry.register_config("basic") as c:
         c.NuTauTT.scifimat_hor = c.NuTauTarget.WallXDim #13.045 * 7 + endpieces = length of hor. mats
         c.NuTauTT.scifimat_vert = c.NuTauTarget.WallYDim #13.045 * 11 + endpieces = length of vert. mats
         c.NuTauTT.scifimat_z = 0.145 * u.cm   # Scintillating fiber mat
-        c.NuTauTT.support_z = 0.02 * u.cm     # Support carbon composite
+        c.NuTauTT.support_z = 1.355 * u.cm     # More space for support (total TTZ 5 cm)
         c.NuTauTT.honeycomb_z = 2 * u.cm      # Airex (or Nomex)
         c.NuTauTT.TTX = c.NuTauTT.scifimat_hor
         c.NuTauTT.TTY = c.NuTauTT.scifimat_vert
         c.NuTauTT.TTZ = 2 * c.NuTauTT.support_z + 2 * c.NuTauTT.scifimat_z + c.NuTauTT.honeycomb_z 
-        c.NuTauTT.n = c.NuTauTarget.wall
+        c.NuTauTT.n = c.NuTauTarget.wall + 1 #starts and end with TT, no HPT
         # should be called after TTX, TTY
         c.NuTauTarget.xdim = c.NuTauTT.TTX 
         c.NuTauTarget.ydim = c.NuTauTT.TTY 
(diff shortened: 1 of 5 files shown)


geometry/geometry_config.py
```diff
@@ -601,8 +601,8 @@
 
         c.tauMudet.Xtot = c.tauMudet.XFe + 2 * c.tauMudet.LateralSupportX#now we need to include also supports.
         c.tauMudet.Ytot = c.tauMudet.YFe + c.tauMudet.UpperSupportY + c.tauMudet.LowerSupportY 
-        #c.tauMudet.Ztot = c.tauMudet.NRpc*c.tauMudet.ZRpc+c.tauMudet.NFethick*c.tauMudet.ZFethick + c.tauMudet.NFethin*c.tauMudet.ZFethin
-        c.tauMudet.Ztot = 3 * u.m #QUICK TEST FOR MAGNETIC SPECTROMETER
```
olantwin: Again, would be good to squash this with whichever commit introduced it.

macro/run_simScript.py
```diff
@@ -44,8 +44,9 @@
 
 globalDesigns = {'2016':{'dy':10.,'dv':5,'ds':7,'nud':1,'caloDesign':0,'strawDesign':4},\
                  '2018':{'dy':10.,'dv':6,'ds':9,'nud':3,'caloDesign':3,'strawDesign':10},\
-                 '2022':{'dy':8.,'dv':6,'ds':9,'nud':3,'caloDesign':3,'strawDesign':10}}
-default = '2022'
+                 '2022':{'dy':8.,'dv':6,'ds':9,'nud':3,'caloDesign':3,'strawDesign':10}, \
+                 '2023':{'dy':8.,'dv':6,'ds':9,'nud':4,'caloDesign':3,'strawDesign':10}}
+default = '2023'
```
olantwin, line 49: This will certainly cause a merge conflict with #420 , but I can take care of it there.

nutaudet/Target.cxx
```diff
@@ -385,16 +388,30 @@ void Target::ConstructGeometry()
...
+   volTungsten->SetLineColor(kGray);
+   //volLead->SetField(magField2);
+
+  }
     
   for(Int_t n=0; n<NPlates; n++)
     {
-      volBrick->AddNode(volLead, n, new TGeoTranslation(0,0,-BrickZ/2+BrickPackageZ/2+ EmPlateWidth + LeadThickness/2 + n*AllPlateWidth)); //LEAD
+      if (fDesign < 4) volBrick->AddNode(volLead, n, new TGeoTranslation(0,0,-BrickZ/2+BrickPackageZ/2+ EmPlateWidth + LeadThickness/2 + n*AllPlateWidth)); //LEAD
+      else volBrick->AddNode(volTungsten, n, new TGeoTranslation(0,0,-BrickZ/2+BrickPackageZ/2+ EmPlateWidth + LeadThickness/2 + n*AllPlateWidth)); //LEAD
```
olantwin, line 414: The comment here should probably be `TUNGSTEN`?

olantwin, line 414: An alternative way to do this woiuld be

```C++
volBrick->AddNode(fDesign < 4 ? volLead: volTungsten, n...)
```

python/shipDet_conf.py
```diff
@@ -187,42 +187,48 @@ def configure(run,ship_geo):
...
+     taumuondetector.SetCrossDimensions(ship_geo.tauMudet.XCross, ship_geo.tauMudet.YCross, ship_geo.tauMudet.ZCross, ship_geo.tauMudet.WidthArm)
+     taumuondetector.SetRpcOuterDimensions(ship_geo.tauMudet.XRpc_outer, ship_geo.tauMudet.YRpc_outer, ship_geo.tauMudet.ZRpc_outer)
+     taumuondetector.SetRpcInnerDimensions(ship_geo.tauMudet.XRpc_inner, ship_geo.tauMudet.YRpc_inner, ship_geo.tauMudet.ZRpc_inner)
+     taumuondetector.SetRpcGapDimensions(ship_geo.tauMudet.XRpcGap, ship_geo.tauMudet.YRpcGap, ship_geo.tauMudet.ZRpcGap)
+     taumuondetector.SetPillarDimensions(ship_geo.tauMudet.PillarX,ship_geo.tauMudet.PillarY, ship_geo.tauMudet.PillarZ)   
+    if ship_geo.nuTauTargetDesign==3:
+     if hasattr(ship_geo.tauMudet, "deltax"): #now replaced with veto taggers
+      taumuondetector.SetRpcDimDifferences(ship_geo.tauMudet.deltax, ship_geo.tauMudet.deltay)
+    if ship_geo.nuTauTargetDesign<3:
+     taumuondetector.SetReturnYokeDimensions(ship_geo.tauMudet.XRyoke,ship_geo.tauMudet.YRyoke, ship_geo.tauMudet.ZRyoke)
```
olantwin, line 219: Not sure what's going on with the white space here...

olantwin:
Generally looks OK, although not all changes are obvious to understand. We should definitely document the code better in future.

- [ ] everything builds
- [ ] simulation runs with 2022
- [ ] simulation runs with 2023
- [ ] geometry looks OK by eye

python/shipDet_conf.py
olantwin: TODO for later: we need to fix the white space on nearly all python files...

ThomasRuf:
All these set methods should go when moving to the system applied in sndsw.

macro/run_simScript.py
olantwin, line 1: It would be good to add the option 4 also to the help text for the command line argument

olantwin (changes requested):
Apart from the comments I left, everything looks good to me.

antonioiuliano2:
> Apart from the comments I left, everything looks good to me.

Thank you very much for your comments, I will start applying them one by one

nutaudet/Target.cxx
antonioiuliano2: Ok, I applied this condition and fixed the comment

antonioiuliano2: Yes, sorry my mistake

geometry/geometry_config.py
antonioiuliano2: Ok, I did it with git rebase -i, then I put the squash to all the related commits

python/shipDet_conf.py
antonioiuliano2: In this case the increase of white space is intentional, since I have added another if before:
if( ship_geo.nuTauTargetDesign<4)

macro/run_simScript.py
antonioiuliano2: Yes, also option 3 (CDS) was missing. I have added both

python/shipDet_conf.py
olantwin: Ah, I see. OK, but we really need to start using some normal python white space convention (2 or 4 spaces) to make this kind of thing readable. Probably need to change this one file at a time though...

olantwin (approved):
I think this is now OK as a first implementation.

olantwin:
Still builds and still no new overlaps after all comments were addressed, so I'll merge this to allow progressing on #420 .

PR #437, opened 2023-06-13
Sc magnet

PlaZMaD:
* remove  trench

* Added SND planes, SND blocks, and a script for geofile generation

* Modified muon shield for SC configuration

* added a 1m gap between the SC magnet and the warm magnet

* Added SC key

---------


olantwin:
> * remove  trench

In the code I read I don't see any changes to the trench apart from changed whitespace? Did I miss something?

>     * Added SND planes, SND blocks, and a script for geofile generation

I'm not sure I understand what you mean here. Why are you modifying the SND?

I'm looking through the other changes right now. I'll submit the review in a few mind.

python/shipDet_conf.py
```diff
@@ -451,6 +452,15 @@ def configure(run,ship_geo):
...
 
+ z_coord = [-3100, -2800, -2600]
+  #z_coord = []
+ for i in range(len(z_coord)):
+  scorplane = ROOT.ScoringPlane("sc_pl_" + str(i), ROOT.kTRUE, ROOT.kFALSE, 116.7, 253.5, 0.1)
+  scorplane.SetVetoPointName("sco" + "_" + str(i))
+  scorplane.SetZposition(z_coord[i])
+  detectorList.append(scorplane)
+
+
```
olantwin, line 463: Again, let's leave the scoring plane separate from the changes for the SC magnet, please.

shipgen/GenLinkDef.h
```diff
@@ -16,4 +16,5 @@
 #pragma link C++ class  NuageGenerator+;
 #pragma link C++ class  FixedTargetGenerator+;
 #pragma link C++ class  ALPACAGenerator+;
+// #pragma link C++ class  UnrolledMuonBackGenerator+;
```
olantwin, line 19: Remove this.

olantwin (changes requested):
@eduard322 @PlaZMaD Thanks a lot for this!

The SC code looks good (or at least good enough for now).

I'll build this now and have a look at the geometry, but apart from the unrelated changes which are not necessary for the SC magnets, this seems to be in a good enough shape to launch some simulations.

olantwin:
Is this what the geometry is supposed to look like? 
image

Are any special options needed to run with the V6 shield?

macro/run_simScript.py
```diff
@@ -108,6 +108,7 @@
 parser.add_argument("-D", "--display", dest="eventDisplay", help="store trajectories", required=False, action="store_true")
 parser.add_argument("--stepMuonShield", dest="muShieldStepGeo", help="activate steps geometry for the muon shield", required=False, action="store_true", default=False)
 parser.add_argument("--coMuonShield", dest="muShieldWithCobaltMagnet", help="replace one of the magnets in the shield with 2.2T cobalt one, downscales other fields, works only for muShieldDesign >2", required=False, type=int, default=0)
+parser.add_argument("--SC", dest="SC_mag", help="Activate SC muon shield. Configuration: 1 SC magnet (3*B_warm) + 3 warm magnets with inverted fields", required=False, default=True)
```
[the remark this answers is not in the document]
olantwin: If there is no sensible default SC configuration yet, we should also require that a geofile is given at the same time.

olantwin:
Could you please provide me with instructions and the geofile ASAP, so that I can test this? @eduard322 @PlaZMaD

I didn't notice initially, but this also targets the wrong branch. I would prefer merging tnis into the ECN3 branch to test everything together first.

olantwin:
@eduard322 @PlaZMaD If it's easier for you, we can discuss on Telegram as well.

olantwin:
Closing in favour of https://github.com/olantwin/FairShip/pull/66

PR #442, opened 2023-07-03
fixing overlap/extrusions in volMuFilter

antonioiuliano2:
Dear all,

During test with --debug 2 this weekend, before launching neutrino simulations, I found some internal overlaps/extrusions in the muon detector (volMuFilter extruded by VolCoil and overlapping with volFeYoke).

This fixes these overlaps, in the meantime I am working on splitting the volumes to add the return magnetic fields.

Best Regards,
Antonio
overlaps volMuFilter extruded


olantwin (approved):
In future, I would suggest adding proper spacing around operators, but as this is an urgent fix, I'll merge it as is.

olantwin:
Also, please copy/paste output from commands (in a code block), instead of posting screenshots, so that the code is copy-able and searchable.

PR #444, opened 2023-07-25
Add header include guard to ShipGlobals.h

olantwin:
This prevents including this header multiple times.

### DIFF
--- a/shipdata/ShipGlobals.h
+++ b/shipdata/ShipGlobals.h
@@ -1,4 +1,8 @@
+#ifndef SHIPGLOBALS_H
+#define SHIPGLOBALS_H
+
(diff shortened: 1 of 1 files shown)


antonioiuliano2 (approved):
Great, always good to hadd #ifndef when needed :)

antonioiuliano2 (approved):

antonioiuliano2:
P.S. I do not have write permissions for FairShip, so I do not think my review did count for that quota xD

olantwin:
> P.S. I do not have write permissions for FairShip, so I do not think my review did count for that quota xD

I should fix that. I trust that you'll use the power responsibly.

PR #463, opened 2024-06-11
first version of the helium-filled decay volume

yabezsh:
This is the first implementation of the option to run the simulation with helium-filled decay volume. The default behavior is the decay volume under vacuum, to call the helium-filled decay volume you should use the flag "--helium". In this configuration, the helium is only placed up to the tracking stations.

To run:     python $FAIRSHIP/macro/run_simScript.py -n 1 --helium

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -138,7 +138,7 @@ parser.add_argument("--noSC", dest="SC_mag", help="Deactivate SC muon shield. Co
 parser.add_argument("--scName", help="The name of the SC shield in the database", default="sc_v6")
 parser.add_argument("--MesonMother",   dest="MM",  help="Choose DP production meson source", required=False,  default=True)
 parser.add_argument("--debug",  help="1: print weights and field 2: make overlap check", required=False, default=0, type=int, choices=range(0,3))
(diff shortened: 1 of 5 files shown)


olantwin:
Thanks @yabezsh, I'll review it tomorrow.

Sorry for the malfunctioning bot, it's now been fixed (hopefully).

macro/run_simScript.py
```diff
@@ -138,6 +138,7 @@
 parser.add_argument("--scName", help="The name of the SC shield in the database", default="sc_v6")
 parser.add_argument("--MesonMother",   dest="MM",  help="Choose DP production meson source", required=False,  default=True)
 parser.add_argument("--debug",  help="1: print weights and field 2: make overlap check", required=False, default=0, type=int, choices=range(0,3))
+parser.add_argument("--helium",  help="use helium bag for vacuum tank", required=False, action="store_const", const="helium", default="vacuums")
```
olantwin, line 141: The `required=False` is redundant here.

Nice solution with the `store_const`

I would rephrase the help message a bit, as it might be misunderstood as replacing the tank structure as well (and break the long line), but this is really nitpicking.

olantwin (approved):
Looks good to me! Thanks for doing this so quickly!

PR #466, opened 2024-06-20
(WIP) importing AdvSND geometry into SHiP

antonioiuliano2:
Dear all,

This is a draft of the AdvSND target geometry imported into SHiP.

The work is still very at an early stage, and it is currently mostly used for first studies of acceptance, before design is optimized. This preliminary pull request is only to facilitate the discussion of the development and addressing the relevant issues.


olantwin:
Please try to avoid unnecessary merges.

olantwin:
@antonioiuliano2 If there is something worth merging from this PR, we should probably cherry-pick it and open a new PR.

As it seems that this configuration will definitely not be the final one, I am closing the PR for now, but I recommend you keep the branch around.

PR #468, opened 2024-06-25
updated Decay Vessel geometry to Naples design 2024

anupama-reghunath:


veto/veto.cxx
```diff
@@ -624,94 +754,94 @@ TGeoVolume* veto::MakeMagnetSegment(Int_t seg){
...
+	if(isVert)tMagVol->AddNode(VertMagCover_X, 4 , new TGeoCombiTrans(0,-ypos,-zpos,new TGeoRotation("r",0,0,0)));
 
 
     double dxOoutCover=thiknes/2;
     double dyOoutCover=604.4*cm;
     double dzOoutCover=56.0*cm/2;
-    double xPosOutetCover = xPos-(358.2-250.4)*cm/2;
-    double yPosOutetCover = 602.4*cm;
-    double zPosOutetCover = zPos-dzOoutCover-thiknes/2;
+    double xposOutetCover = xpos-(358.2-250.4)*cm/2;
```
olantwin, line 771: What does "Outet" mean? "Oout" also is very strange. Would be good if we could give things somewhat self-explanatory names.

veto/veto.cxx
```diff
@@ -720,60 +850,60 @@ TGeoVolume* veto::MakeMagnetSegment(Int_t seg){
...
+    xpos=-255.0*cm;
+    if(isLongOutCover)tMagVol->AddNode(LongOutCover_Y4, 1 , new TGeoTranslation(xpos,ypos,zpos));
+    if(isLongOutCover)tMagVol->AddNode(LongOutCover_Y4, 2 , new TGeoTranslation(xpos,ypos,-zpos));
+    if(isLongOutCover)tMagVol->AddNode(LongOutCover_Y4, 3 , new TGeoTranslation(-xpos,ypos,zpos));
+    if(isLongOutCover)tMagVol->AddNode(LongOutCover_Y4, 4 , new TGeoTranslation(-xpos,ypos,-zpos));
 
 
     /*
-    dxOoutCover=fabs(xPosOutetCover)-thiknes;
+    dxOoutCover=fabs(xposOutetCover)-thiknes;
```
olantwin, line 906: This dead code should probably be removed instead of updated.

veto/veto.cxx
```diff
@@ -1151,6 +1272,7 @@ void veto::ConstructGeometry()
       T6Lid->SetLineColor(2);
       T6Lid->SetFillColor(2);
       tMaGVol->AddNode(T6Lid, 1, new TGeoTranslation(0, 0,fTub6z+fTub6length+f_LidThickness/2.+0.1*cm - zStartMagVol));
+      //tDecayVol->AddNode(T6Lid, 1, new TGeoTranslation(0, 0,50+-ribthick?));
```
olantwin, line 1275: Remove.

veto/vetoHit.cxx
```diff
@@ -78,13 +78,13 @@ TGeoNode* vetoHit::GetNode()
...
-        if(Zlayer<10)path+="0";
-        path+=Zlayer;
-   }
-   else {
-        path+="00";
-   }
+   //if(ShapeType<3){
+   if(Zlayer<10)path+="0";
+   path+=Zlayer;
+   //}
```
olantwin, line 84: Remove commented code.

veto/veto.cxx
```diff
@@ -385,217 +427,305 @@ int veto::liscId(TString ShapeTypeName,int blockNr ,int Zlayer,int number, int p
 
     if(ShapeType<0) id =  999999;
     else id = ShapeType*100000 + blockNr*10000 + Zlayer*100 + number*10 + position;
+    
+    //std::cout << "Generated ID for " << ShapeTypeName << ": " << id << std::endl;
```
olantwin, line 431: Remove.

veto/veto.cxx
```diff
@@ -198,23 +207,32 @@ double wx(double z){
...
+                              double dB,
+                              Int_t color,
+                              TGeoMedium *material,
+                              Bool_t sens=kFALSE)
+    {//TGeoVolume* LiScX = GeoSideObj(name, half space between ribs along z, width in X at the beginning,width in Y at the beginning,width in X at the end,width in Y at the end, change in width along X,change in width along y, kMagenta - 10, vetoMed, true);
+
+      //a1- width in X, at the beginning
+      //b1- width in Y, at the beginning
+      //a2- width in X, at the end
+      //b2- width in Y, at the end
```
olantwin, line 235: It would be best to reformat this comment in such a way that Doxygen can parse it for the automatic documentation, see e.g. here: https://en.wikipedia.org/wiki/Doxygen#Example_code

olantwin (changes requested):
With all the formatting changes and renames, it's hard to tell what significant changes have been made.

Could you maybe add what this geometry now looks like in the event display to the discussion?

If might also be good to link your slides from the software meeting.

For the formatting, I would suggest running clang-format over veto.cxx, as its formatting is very inconsistent.

olantwin:
I'll try to build and look at the geometry in more detail towards the end of the week.

fairlyons:
I did a visual inspection of the geometry and looked over the code changes. Everything looks good to me.

python/shipDet_conf.py
```diff
@@ -1,486 +1,906 @@
 #!/usr/bin/env python
 # -*- coding: latin-1 -*-
-import ROOT,os
+from __future__ import print_function
+from __future__ import division
```
olantwin, line 4: Why did you readd these? They should no longer have any function.

geometry/veto_config.yaml
olantwin, line 1: This looks great as a first example!

- It might be good to remove one of the blank lines at the top.
- Do we use "Aluminum" as a material? I think we should rename it to "Aluminium" at some point, but that's beyond the scope of this PR.

veto/veto.cxx
```diff
@@ -1,1205 +1,1151 @@
...
+
+    vetoMed = gGeoManager->GetMedium(vetoMed_name);   //! medium of veto counter, liquid or plastic scintillator
+    supportMedIn = gGeoManager->GetMedium(supportMedIn_name);       //! medium of support structure, iron, balloon
+    supportMedOut = gGeoManager->GetMedium(supportMedOut_name);     //! medium of support structure, aluminium, balloon
+    decayVolumeMed = gGeoManager->GetMedium(decayVolumeMed_name);   // decay volume, air/helium/vacuum
+
+    TGeoVolume* tDecayVol = new TGeoVolumeAssembly("DecayVolume");
+
+    // TGeoVolume* T1Lid = MakeLidSegments(1, VetoStartInnerX / 2. + wallThick, VetoStartInnerY / 2. + wallThick);
+    // tDecayVol->AddNode(T1Lid, 1, new TGeoTranslation(0, 0, zpos - zStartDecayVol+f_LidThickness/2.1));
```
olantwin, line 1105: Is there a reason to keep these lines around?

veto/vetoContFact.cxx
olantwin, line 1: Does this actually do anything?

olantwin: Maybe we could use this for geometry constants and alignment parameters? Probably worth looking at what the purpose normally is in FairRoot.

veto/vetoHit.cxx
```diff
@@ -1,106 +1,110 @@
...
+#include "TVector3.h"
+#include "veto.h"
 
 #include <iostream>
 #include <math.h>
 using std::cout;
 using std::endl;
 
-Double_t speedOfLight = TMath::C() *100./1000000000.0 ; // from m/sec to cm/ns
+Double_t speedOfLight = TMath::C() * 100. / 1000000000.0;   // from m/sec to cm/ns
```
olantwin, line 18: Shouldn't we take this value from shipunit?

veto/vetoHit.cxx
```diff
@@ -1,106 +1,110 @@
...
 TVector3 vetoHit::GetXYZ()
 {
     TGeoNavigator* nav = gGeoManager->GetCurrentNavigator();
     TGeoNode* node = GetNode();
-    TGeoBBox* shape =  (TGeoBBox*)node->GetVolume()->GetShape();
-    Double_t origin[3] = {shape->GetOrigin()[0],shape->GetOrigin()[1],shape->GetOrigin()[2]};
-    Double_t master[3] = {0,0,0};
-    nav->LocalToMaster(origin,master);
-    TVector3 pos = TVector3(master[0],master[1],master[2]);
+    TGeoBBox* shape = (TGeoBBox*)node->GetVolume()->GetShape();
```
olantwin, line 43: Please don't use C-style casts.

veto/vetoHit.cxx
```diff
@@ -1,106 +1,110 @@
...
-
-
 // -----   Public method Print   -------------------------------------------
 void vetoHit::Print(Int_t detID) const
 {
-  cout << "-I- vetoHit: veto hit " << " in detector " << fDetectorID << endl;
-  cout << "  ADC " << fdigi << " ns" << endl;
+    cout << "-I- vetoHit: veto hit " << " in detector " << fDetectorID << endl;
+    cout << "  ADC " << fdigi << " ns" << endl;
 }
```
olantwin, line 108: Let's use FairLogger here, with the INFO severity. We should avoid using cout.

veto/vetoHitOnTrack.h
olantwin, line 1: What does this class do? I never noticed it before...

veto/vetoLinkDef.h
olantwin, line 1: We need to be careful here, I'm not sure how this is parsed. Autoformatting the linkdefs using clang-format might not be a good idea.

veto/vetoPoint.cxx
```diff
@@ -4,38 +4,41 @@
...
-       << ") cm" << endl;
-  cout << "    Momentum (" << fPx << ", " << fPy << ", " << fPz
-       << ") GeV" << endl;
-  cout << "    Time " << fTime << " ns,  Length " << fLength
-       << " cm,  Energy loss " << fELoss*1.0e06 << " keV" << endl;
+    cout << "-I- vetoPoint: veto point for track " << fTrackID << " in detector " << fDetectorID << endl;
+    cout << "    Position (" << fX << ", " << fY << ", " << fZ << ") cm" << endl;
+    cout << "    Momentum (" << fPx << ", " << fPy << ", " << fPz << ") GeV" << endl;
+    cout << "    Time " << fTime << " ns,  Length " << fLength << " cm,  Energy loss " << fELoss * 1.0e06 << " keV"
+         << endl;
```
olantwin, line 42: FairLogger

geometry/veto_config.yaml
olantwin: So far this is not used yet, correct?

olantwin (changes requested):
Generally much more readable, and a great example of how hopefully all detector code will look like soon.

For the geometry changes, can you confirm that there are no overlaps?

Some git nitpicks:
- It's confusing that you committed as two different authors. Maybe something is misconfigured somewhere?
- I would be good if we could undo the merge, and rebase instead.

python/shipDet_conf.py
anupama-reghunath: I believe this was because I used a version before it was cleaned out. So when I rebased this popped back up.

geometry/veto_config.yaml
anupama-reghunath: Yes, I used the Aluminum option that was already available. It is used, you can see the usage in shipDet_conf.py .
image

veto/veto.cxx
anupama-reghunath: Well the code is technically obsolete, but I kept it just incase we plan to place anything in the future. Maybe I take it out then?

veto/vetoHit.cxx
anupama-reghunath: will remove this, I think its remnant from oldcode after rebasing with origin post clean up.

veto/veto.cxx
```diff
@@ -1,1327 +1,1113 @@
...
 }
 
 void veto::Initialize()
 {
-  FairDetector::Initialize();
-//  FairRuntimeDb* rtdb= FairRun::Instance()->GetRuntimeDb();
-//  vetoGeoPar* par=(vetoGeoPar*)(rtdb->getContainer("vetoGeoPar"));
+    FairDetector::Initialize();
+    //  FairRuntimeDb* rtdb= FairRun::Instance()->GetRuntimeDb();
+    //  vetoGeoPar* par=(vetoGeoPar*)(rtdb->getContainer("vetoGeoPar"));
```
olantwin, line 96: This code can also be removed.

geometry/veto_config.yaml
olantwin: Ah, great! I missed that in all the changes to shipDet_conf.py, and because it was in a different commit from the one introducing the yaml.

veto/vetoContFact.cxx
olantwin: To be investigated at a later time.

veto/veto.cxx
olantwin: Probably a new lid would be implemented differently anyway.
I think we can safely remove this. We can always look at older commits, if necessary.

python/shipDet_conf.py
```diff
@@ -1,486 +1,906 @@
...
+    Veto.SetVesselDimensions(
+        veto_geo.xstartInner,
+        veto_geo.xendInner,
+        veto_geo.ystartInner,
+        veto_geo.yendInner,
+        veto_geo.z0
+
+    )
+    Veto.SetLiquidVeto(1)  
+    Veto.SetPlasticVeto(1)
```
olantwin, line 111: This line surprises me. Does it actually do anything?

python/shipDet_conf.py
```diff
@@ -1,486 +1,906 @@
...
+        fn.close()
+    return EcalZSize, sz
+
+
+def posEcal(z, efile):
+    EcalZSize, sz = makeEcalGeoFile(z, efile)
+    ecal = ROOT.ecal("Ecal", ROOT.kTRUE, sz)
+    return ecal, EcalZSize
+
+def configure_veto(yaml_file):
```
olantwin, line 94: We might eventually want to load the yaml directly in the C++ file, and here only for information needed by other systems/globally.
But for now, this is a nice solution that requires no changes elsewhere.

olantwin (changes requested):
This only changed the authorship and rebased instead of merged, right?

The other nitpicks remain.

veto/veto.cxx
anupama-reghunath: ok will do.

olantwin:
#498 can now be used to rebase this and split the massive commits in a nicer way.

veto/vetoHit.cxx
```diff
@@ -1,106 +1,116 @@
...
 
-
-
 // -----   Public method Print   -------------------------------------------
 void vetoHit::Print(Int_t detID) const
 {
-  cout << "-I- vetoHit: veto hit " << " in detector " << fDetectorID << endl;
-  cout << "  ADC " << fdigi << " ns" << endl;
+    LOG(DEBUG) << "-I- vetoHit: veto hit " << " in detector " << fDetectorID ;
```
olantwin, line 112: I would remove the `-I-` here, as it indicates the INFO severity, and make the severity for the LOG(INFO) to match the past behaviour. DEBUG isn't printed by default.

veto/vetoHit.h
```diff
@@ -1,16 +1,19 @@
...
-#include "FairVolume.h"
+#ifndef VETO_VETOHIT_H_
+#define VETO_VETOHIT_H_ 1
+//#include "FairVolume.h"
 #include "ShipHit.h"
-#include "vetoPoint.h"
-#include "TObject.h"
-#include "TGeoShape.h"
-#include "TGeoPhysicalNode.h"
+//#include "vetoPoint.h"
```
olantwin, line 5: If it builds, can you remove the commented headers, please?

veto/vetoLinkDef.h
olantwin, line 1: Please don't modify this file.

veto/vetoPoint.cxx
```diff
@@ -1,41 +1,44 @@
...
 {
-  cout << "-I- vetoPoint: veto point for track " << fTrackID
-       << " in detector " << fDetectorID << endl;
-  cout << "    Position (" << fX << ", " << fY << ", " << fZ
-       << ") cm" << endl;
-  cout << "    Momentum (" << fPx << ", " << fPy << ", " << fPz
-       << ") GeV" << endl;
-  cout << "    Time " << fTime << " ns,  Length " << fLength
-       << " cm,  Energy loss " << fELoss*1.0e06 << " keV" << endl;
+  LOG(DEBUG) << "-I- vetoPoint: veto point for track " << fTrackID << " in detector " << fDetectorID;
```
olantwin, line 38: See above.

olantwin (approved):

PR #470, opened 2024-06-28
Fixed the typo in the density of the helium

yabezsh:
The density of helium was defined as 1.78e-3 g/cm3 while it should have been 1.78E-4 g/cm3.


olantwin (approved):

olantwin:
Thanks!

PR #473, opened 2024-07-02
#471 remove pyfuturepast

### LINKED ISSUE #471: Stop using future/past python modules
These are almost always no longer needed, and a leftover of when we had to support python 2 and 3 simultaneously.

The pickler probably needs to change in a breaking way, everything else should be trivial!

egamberini:
Closes #471


olantwin (dismissed):
Thanks a lot!

Are you coming to the dinner this evening? I owe you a drink or two of your choice!

olantwin (approved):

PR #481, opened 2024-07-30
Fixing warnings during building related to nutaudet folder, about unused or uninitialized variables

antonioiuliano2:
Dear all FairShip users,

This commit fixes all nutaudet warnings happening during building the code.

They were about either unused or uninitialized variables: most of the times the warnings were genuine, i.e. the variable was actually "forgotten" in the code and never used. In other instances, however, the object was actually used, even if the variable was not. This happens a lot with TGeoShapes, since we define the variable when we build the shape, but we do not use the variable itself in the code. I fixed these warnings by addind a redundant SetName call after building the shape.

I attach the initial log with the warnings, as provided by @olantwin, for reference. I also attach the new log, without nutaudet warnings.

I will fix the charmdet warnings in another branch, since they are related to a completely different geometry.

Best Regards,
Antonio

[FairShip_Oliver.log](https://github.com/user-attachments/files/16425858/FairShip_Oliver.log)
[fixingnutaudetwarnings_build.log](https://github.com/user-attachments/files/16425870/fixingnutaudetwarnings_build.log)

### DIFF
--- a/nutaudet/NuTauMudet.cxx
+++ b/nutaudet/NuTauMudet.cxx
@@ -504,13 +504,11 @@ void NuTauMudet::ConstructGeometry()
       TGeoVolumeAssembly *volMudetBox = new TGeoVolumeAssembly("volTauNuMudet");
       tTauNuDet->AddNode(volMudetBox, 1, new TGeoTranslation(0,0,fZcenter));
 
-      TGeoBBox *IronLayer = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFe/2);
+      TGeoBBox *IronLayer = new TGeoBBox("MUDETIRON",fXFe/2, fYFe/2, fZFe/2);
       TGeoVolume *volIron = new TGeoVolume("volIron",IronLayer,Iron);
       volIron->SetLineColor(kGray);
-      IronLayer->SetName("MUDETIRON");
 
-      auto* IronLayer1 = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFethin/2);
-      IronLayer1->SetName("MUDETIRON1");
+      auto* IronLayer1 = new TGeoBBox("MUDETIRON1",fXFe/2, fYFe/2, fZFethin/2);
       //***********************ADDING EXTERNAL DETAILS TO THE MUON FILTER
 
       //********UPPER COVER***********


nutaudet/Target.cxx
```diff
@@ -352,6 +339,9 @@ void Target::ConstructGeometry()
 	  magField2->SetFieldValue(fField,0,0.);
 	  MagnetVol=gGeoManager->GetVolume("NudetMagnet");
 	}
+    else{
+    MagnetVol = NULL; //initialize to nothing, fixing -Wmaybe-uninitialized warnings
```
olantwin, line 343: Should be initialised directly in line 322.

Please use `nullptr`, `NULL` should not be used in modern code.

nutaudet/Target.cxx
```diff
@@ -398,12 +388,15 @@ void Target::ConstructGeometry()
    volLead->SetTransparency(1);
    volLead->SetLineColor(kGray);
    //volLead->SetField(magField2);
+   volTungsten = NULL; //fixes -Wmaybe-uninitialized warning
```
olantwin, line 391: Maybe better to have a volume `absorber`, which then is either tungsten or lead depending on the config?

Then the other volume is not needed.

nutaudet/Target.cxx
```diff
@@ -650,7 +641,7 @@ Bool_t  Target::ProcessHits(FairVolume* vol)
 
     Int_t motherV[MaxL];
 //   Bool_t EmTop = 0, EmBot = 0, EmCESTop = 0, EmCESBot = 0;
-    Bool_t EmBrick = 0, EmCES = 0, EmTop;
+    Bool_t EmBrick = 0, EmTop;
```
olantwin, line 644: Please don't define multiple things on the same line, this is unreadable.
Also, `Bool_t` should be initialised to `false`, not `0` for readability.

nutaudet/Target.cxx
```diff
@@ -748,7 +732,7 @@ Bool_t  Target::ProcessHits(FairVolume* vol)
 
 void Target::DecodeBrickID(Int_t detID, Int_t &NWall, Int_t &NRow, Int_t &NColumn, Int_t &NPlate, Bool_t &EmCES, Bool_t &EmBrick, Bool_t &EmTop)
 {
-  Bool_t BrickorCES = 0, TopBot = 0;
+  Bool_t BrickorCES = 0;
```
olantwin, line 735: `false`, not `0`

nutaudet/NuTauMudet.cxx
```diff
@@ -511,8 +508,7 @@ void NuTauMudet::ConstructGeometry()
       TGeoVolume *volIron = new TGeoVolume("volIron",IronLayer,Iron);
       volIron->SetLineColor(kGray);
 
-      TGeoBBox *IronLayer1 = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFethin/2);
-      TGeoVolume *volIron1 = new TGeoVolume("volIron1",IronLayer1,Iron);
+      TGeoBBox *IronLayer1 = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFethin/2);
```
olantwin, line 511: No need to repeat the class name, `auto*` would make this less verbose

olantwin (changes requested):
First of all, thanks!

Some remarks:
- For unused variables that are needed (e.g. the different shapes/volumes): Please use the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute instead of adding unnecessary statements.
- I don't think we need to make these fixes for the `charmdet` dir, as I believe they are outdated relative to the `muonflux` branch anyway. I would propose removing `charmdet` from the `master` branch instead.

nutaudet/Target.cxx
antonioiuliano2: You are right, thanks!

nutaudet/Target.cxx
antonioiuliano2: Ok, I did not change however the volume names, so we keep the geometry files consistent.

nutaudet/Target.cxx
antonioiuliano2: Ok, done.

nutaudet/NuTauMudet.cxx
antonioiuliano2: Ok! For now I have just done to this line, but I guess it would apply to all of geometry code lines in the future.

nutaudet/Target.cxx
antonioiuliano2: Ok

nutaudet/EmulsionMagnet.cxx
```diff
@@ -513,25 +508,23 @@ void EmulsionMagnet::ConstructGeometry()
       TGeoCombiTrans* combleft = new TGeoCombiTrans(transleft,rot1);
       combleft->SetName("combleft");
       combleft->RegisterYourself();
-      TGeoCompositeShape *cutleft;
+      [[maybe_unused]] 	TGeoCompositeShape *cutleft;
```
olantwin, line 511: single space, why not directly initialise?

olantwin (changes requested):
Nearly perfect.

Could you please also add a line to the change log?

nutaudet/Target.cxx
```diff
@@ -389,27 +376,25 @@ void Target::ConstructGeometry()
   volBrick->SetLineColor(kCyan);
   volBrick->SetTransparency(1);
 
-  TGeoBBox *Lead, *Tungsten; //need to separate the two cases
-  TGeoVolume *volLead, *volTungsten;
+  TGeoBBox *Absorber; //need to separate the two cases
+  TGeoVolume *volAbsorber;
```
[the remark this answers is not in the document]
antonioiuliano2: Ok, I have used the ternary statement with the following syntax:

```
TGeoBBox *Absorber = (fDesign < 4) ? true : false
```

nutaudet/Target.cxx
```diff
@@ -650,7 +633,8 @@ Bool_t  Target::ProcessHits(FairVolume* vol)
 
     Int_t motherV[MaxL];
 //   Bool_t EmTop = 0, EmBot = 0, EmCESTop = 0, EmCESBot = 0;
-    Bool_t EmBrick = 0, EmCES = 0, EmTop;
+    Bool_t EmBrick = false;
+    Bool_t EmTop;
```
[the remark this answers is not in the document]
antonioiuliano2: Not at all, I have added the initialization for EmTop too

nutaudet/Target.cxx
```diff
@@ -375,18 +375,11 @@ void Target::ConstructGeometry()
...
-  }
-  else{
-   Absorber = new TGeoBBox("W", EmulsionX/2, EmulsionY/2, LeadThickness/2);
-   volAbsorber = new TGeoVolume("Tungsten",Absorber,tungsten);
-  }
+  //need to separate the two cases, now with a ternary operator
+  TGeoBBox *Absorber = (fDesign < 4) ? 
+    new TGeoBBox("Pb", EmulsionX/2, EmulsionY/2, LeadThickness/2) : new TGeoBBox("W", EmulsionX/2, EmulsionY/2, LeadThickness/2);
+  TGeoVolume *volAbsorber = (fDesign < 4) ?
+    new TGeoVolume("Lead",Absorber,lead) : new TGeoVolume("Tungsten",Absorber,tungsten);
```
olantwin, line 382: Do we maybe need `[[maybe_unused]]` for these later, or are they both "used" in a way that's visible to the compiler? Please check whether there is a warning.

You could use `auto` here as well, now that the initialisation determines the type in an unambiguous way.

We could probably simplify this further by only changing the material of `volAbsorber` and keeping everything else the same.

olantwin (changes requested):
See comment, everything else looks good to me now.

Before the final merge, you'd have to rebase to deal with the conflict in the CHANGELOG.md.

nutaudet/Target.cxx
olantwin: This is also still outstanding.

nutaudet/Target.cxx
antonioiuliano2: Absorber and volAbsorber are always used, since they are part of the main Target. 
The only difference is the material, if it is lead (originally planned for SHiP) or tungsten (the most updated designs). 
I agree for the `auto` comment

antonioiuliano2:
> See comment, everything else looks good to me now.
> 
> Before the final merge, you'd have to rebase to deal with the conflict in the CHANGELOG.md.

Done, I have do force push afterwards, because git complained about the changed history.

nutaudet/Target.cxx
antonioiuliano2: Sorry, I have changed it.

olantwin:
> Done, I had to force push afterwards, because git complained about the changed history.

In general, you'll have to do a force push after a rebase. Next time, please do it as the last thing, once everything else is resolved, as it makes it harder to review the changes.

This time it's fine though, as there were very few outstanding issues.

nutaudet/NuTauMudet.cxx
```diff
@@ -511,8 +508,7 @@ void NuTauMudet::ConstructGeometry()
       TGeoVolume *volIron = new TGeoVolume("volIron",IronLayer,Iron);
       volIron->SetLineColor(kGray);
 
-      TGeoBBox *IronLayer1 = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFethin/2);
-      TGeoVolume *volIron1 = new TGeoVolume("volIron1",IronLayer1,Iron);
+      auto* IronLayer1 = new TGeoBBox("Iron",fXFe/2, fYFe/2, fZFethin/2);
```
olantwin, line 511: Is this used anywhere?

nutaudet/Target.cxx
```diff
@@ -388,28 +375,18 @@ void Target::ConstructGeometry()
   TGeoVolume *volBrick = new TGeoVolume("Brick",Brick,air);
   volBrick->SetLineColor(kCyan);
   volBrick->SetTransparency(1);
+  //need to separate the two cases, now with a ternary operator
+  auto *Absorber = new TGeoBBox("Absorber", EmulsionX/2, EmulsionY/2, LeadThickness/2);
+  auto *volAbsorber = (fDesign < 4) ?
+    new TGeoVolume("volAbsorber",Absorber,lead) : new TGeoVolume("volAbsorber",Absorber,tungsten);
```
olantwin, line 381: To avoid repeating the things that are the same:

```suggestion
  auto *volAbsorber = new TGeoVolume("volAbsorber", Absorber, (fDesign < 4) ? lead : tungsten);
```

nutaudet/HPT.cxx
```diff
@@ -232,12 +232,9 @@ void Hpt::ConstructGeometry()
...
         TGeoVolume *volMagRegion=gGeoManager->GetVolume("volMagRegion");
-        TGeoVolume *volTarget =gGeoManager->GetVolume("volTarget");
-        TGeoVolume *tTauNuDet = gGeoManager->GetVolume("tTauNuDet");
+        TGeoVolume *volTarget =gGeoManager->GetVolume("volTarget");        
 
         Double_t DZMagnetizedRegion = ((TGeoBBox*) volMagRegion->GetShape())->GetDZ() *2;
-        Double_t DYMagnetizedRegion = ((TGeoBBox*) volMagRegion->GetShape())->GetDY() *2;
-        Double_t DXMagnetizedRegion = ((TGeoBBox*) volMagRegion->GetShape())->GetDX() *2;
 
         Double_t DZTarget = ((TGeoBBox*) volTarget->GetShape())->GetDZ() *2;
```
olantwin, line 239: Could you fix this C-style cast please?

nutaudet/HPT.cxx
```diff
@@ -232,12 +232,9 @@ void Hpt::ConstructGeometry()
     if (fDesign == 3){
         //Trackers that in design 3 follow the target --------------------------------------------------------------------------------------
         TGeoVolume *volMagRegion=gGeoManager->GetVolume("volMagRegion");
-        TGeoVolume *volTarget =gGeoManager->GetVolume("volTarget");
-        TGeoVolume *tTauNuDet = gGeoManager->GetVolume("tTauNuDet");
+        TGeoVolume *volTarget =gGeoManager->GetVolume("volTarget");        
 
         Double_t DZMagnetizedRegion = ((TGeoBBox*) volMagRegion->GetShape())->GetDZ() *2;
```
olantwin, line 237: Could you fix this C-style cast please?

olantwin (changes requested):
99% perfect ;)

Sorry for being so nitpicky!

nutaudet/NuTauMudet.cxx
olantwin: Turns out it is used in line 573. I would suggest skipping the rename on line 551, which makes it harder to find that use and give it the name you want immediately.
Same for IronLayer.

olantwin:
@antonioiuliano2 Can you give an estimate of when the requested changes can be implemented (or let me know if you disagree with any comments?)

This PR has now been open for 3 months.

nutaudet/HPT.cxx
antonioiuliano2: Sorry, I did not get this comment. 

The cast is required in order to access the GetDX() method being required from the TGeoBBox class, while TGeoVolume::GetShape() returns a generic TGeoShape object:

https://root.cern/doc/master/classTGeoVolume.html#aaf5dc92206d7bc56e7bdb9f84287c2fe

nutaudet/NuTauMudet.cxx
antonioiuliano2: Thank you, I have moved the SetName() lines above.

olantwin: Thanks, but what I meant is, why don't you just `new TGeoBBox("MUDETIRON"...)`? You can just set the name in the constructor.

nutaudet/HPT.cxx
olantwin: C-style casts can easily lead to issues. A cast here is needed, but should be a `static_cast`, which is stricter and easier to search for.

See e.g.:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast

olantwin: See also https://github.com/ShipSoft/FairShip/wiki/CPP-guidelines

antonioiuliano2: I confirm that after the rebase the compilation and test simulation worked succesfully. 
I am now going to force push to apply it.

nutaudet/NuTauMudet.cxx
olantwin: This is now the only outstanding issue. Otherwise things look good to be merged.

antonioiuliano2: Ah, I understand now, sorry!

PR #483, opened 2024-08-08
Update geometry constants for SST

ukmartylee:
Update geometry constants for SST

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ it in future.
 ### Changed
 
 * Update getGeoInformation command line interface
+* Update SST geometry constants #483
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


olantwin (changes requested):
Thanks for the changes Marty!

Some nitpicks:

* Could you please remove the extra spaces before the equals sign on the you modify? 
* If you could also add a space before/after the multiplication operators that would be great. 
* Finally, could you please rephrase the commit message? Maybe "Update geometry constants for SST" or something similar would be good. Please also add this to the CHANGELOG.md.

Could you maybe briefly highlight the changes at the meeting this Monday?

ukmartylee:
Sure thing Oliver!
If you can send me the link for the Monday’s meeting, that would be great!

olantwin:
> If you can send me the link for the Monday’s meeting, that would be great!

https://indico.cern.ch/event/1445547/

Issue #489, opened 2024-08-21
Issue with SST parameters


anupama-reghunath:
With the new parameters for the SST, I'm having trouble running simulations. DeltazPlane and YPlaneOffset are now missing in geometry_config.py, but are used in the shipDet_conf.py and subsequently in strawtubes.cxx,
For the timebeing and also for backward compatibility, I've set them as
```
     c.strawtubes.DeltazPlane  = 0*u.cm
     c.strawtubes.YPlaneOffset = 0*u.cm
```
but not sure if the implementation is accurate.
Also, stereoangle expected is an integer in the .cxx and the respective header, and not a float value(4.57). 

Tested after these changes, seems to be working.

olantwin:
CC @ukmartylee

ukmartylee:
Setting them to zero at the moment, I think, would be fine. My concern is that by the definition in strawtubes.cxx, would the fact that two planes are created and placed at the same position (overlapping each other) be an issue? Otherwise, we’ll have to remove the plane loop in strawtubes.cxx completely.

And it would be sufficient for the stereo angle to just be 5, since this value is not fixed at the moment.

olantwin:
Overlaps should be avoided. Removing the plane loop sounds like the best solution.

We should probably allow for non-integer angles, should be a very simple change.

ukmartylee:
The interim solution could be that we’ll just perform 1 plane loop. Meanwhile, I’ll try to remove all plane-related variables and codes.

olantwin:
Thanks a lot!

Could you please implement this interim solution, so that the simulation and reconstruction work again?

PR #490, opened 2024-08-23
Implement interim solution with SST parameters

ukmartylee:
Retain plane setup for SST momentarily

Change stereoangle to double


olantwin:
Sorry, for the late reply. I somehow missed this.

I assume this fixes #489 ?

olantwin (changes requested):
The changes look good.

Please test that there are no overlaps.

It would be good to squash these changes into a single commit and add a line to the changelog.

anupama-reghunath (approved):

olantwin (approved):

PR #493, opened 2024-09-04
Add BzSign for ShipBFieldMap

olantwin:
As reported by Massi (https://indico.cern.ch/event/1447048/), The octant symmetry is not correct for the Bz field.


olantwin:
@Gfrisella Could you please check that this bug fix works as intended? Massi should be able to explain to you how to reproduce the issue.

Gfrisella:
I confirm the bug has been fixed. 
Bz_fieldmap

PR #510, opened 2024-09-12
Timing Correction for MuonDIS generation

anupama-reghunath:
Attached are the SBT response plots produced for the previous MuonDIS(involving backwards travelling muons), and with the suggested corrections with the new MuonDIS. SBThits are color coded wrt to the time of DIS. New method preserves the incoming muon's time as well as its veto response, and ensures that the event is temporally consistent.

Before:
SBT_event_oldmuDIS
After:
SBT_event_newmuDIS

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -510,7 +510,7 @@ if options.dryrun: # Early stop after setting up Pythia 8
  sys.exit(0)
 gMC = ROOT.TVirtualMC.GetMC()
 fStack = gMC.GetStack()
-EnergyCut = 10.*u.MeV if options.mudis else 100.*u.MeV
+EnergyCut = 10. * u.MeV if options.mudis else 100. * u.MeV
 
 if MCTracksWithHitsOnly:
  fStack.SetMinPoints(1)
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -1,6 +1,4 @@
 #!/usr/bin/env python
-from __future__ import print_function
-from __future__ import division
 import os
 import sys
 import ROOT
@@ -32,7 +30,7 @@ MCTracksWithHitsOnly   = False  # copy particles which produced a hit and their 
 MCTracksWithEnergyCutOnly = True # copy particles above a certain kin energy cut
 MCTracksWithHitsOrEnergyCut = False # or of above, factor 2 file size increase compared to MCTracksWithEnergyCutOnly
 
-charmonly    = False  # option to be set with -A to enable only charm decays, charm x-sec measurement  
+charmonly    = False  # option to be set with -A to enable only charm decays, charm x-sec measurement
 HNL          = True
 
 inputFile    = "/eos/experiment/ship/data/Charm/Cascade-parp16-MSTP82-1-MSEL4-978Bpot.root"
@@ -71,8 +69,7 @@ globalDesigns = {
 }
 default = '2023'
 
-inactivateMuonProcesses = False   # provisionally for making studies of various muon background sources
-inactivateMuonNuclearProcess = False #Only True for muonDIS
+inactivateMuonProcesses = False         # provisionally for making studies of various muon background sources
 
 parser = ArgumentParser()
 group = parser.add_mutually_exclusive_group()
@@ -152,6 +149,7 @@ parser.add_argument(
     const="vacuums",
     default="helium"
 )
+
 parser.add_argument("--SND", dest="SND", help="Activate SND.", action='store_true')
 parser.add_argument("--noSND", dest="SND", help="Deactivate SND. NOOP, as it currently defaults to off.", action='store_false')
 
@@ -172,7 +170,7 @@ if options.A != 'c':
      if options.A =='b': inputFile = "/eos/experiment/ship/data/Beauty/Cascade-run0-19-parp16-MSTP82-1-MSEL5-5338Bpot.root"
      if options.A.lower() == 'charmonly':
            charmonly = True
-           HNL = False 
+           HNL = False
      if options.A not in ['b','c','bc','meson','pbrem','qcd']: inclusive = True
 if options.MM:
      motherMode=options.MM
@@ -199,11 +197,11 @@ if options.testFlag:
 
 
 #sanity check
-if (HNL and options.RPVSUSY) or (HNL and options.DarkPhoton) or (options.DarkPhoton and options.RPVSUSY): 
+if (HNL and options.RPVSUSY) or (HNL and options.DarkPhoton) or (options.DarkPhoton and options.RPVSUSY):
  print("cannot have HNL and SUSY or DP at the same time, abort")
  sys.exit(2)
 
-if (simEngine == "Genie" or simEngine == "nuRadiography") and defaultInputFile: 
+if (simEngine == "Genie" or simEngine == "nuRadiography") and defaultInputFile:
   inputFile = "/eos/experiment/ship/data/GenieEvents/genie-nu_mu.root"
 if simEngine == "muonDIS" and defaultInputFile:
   print('input file required if simEngine = muonDIS')
@@ -245,12 +243,11 @@ else: tag = simEngine+"-"+mcEngine
 if charmonly: tag = simEngine+"CharmOnly-"+mcEngine
 if options.eventDisplay: tag = tag+'_D'
 if options.dv > 4 : tag = 'conical.'+tag
-elif dy: tag = str(options.dy)+'.'+tag 
 if not os.path.exists(options.outputDir):
   os.makedirs(options.outputDir)
 outFile = f"{options.outputDir}/ship.{tag}.root"
 
-# rm older files !!! 
+# rm older files !!!
 for x in os.listdir(options.outputDir):
   if not x.find(tag)<0: os.system(f"rm {options.outputDir}/{x}" )
 # Parameter file name
@@ -267,18 +264,17 @@ timer.Start()
 run = ROOT.FairRunSim()
 run.SetName(mcEngine)  # Transport engine
 run.SetSink(ROOT.FairRootFileSink(outFile))  # Output file
-run.SetUserConfig("g4Config.C") # user configuration file default g4Config.C 
-rtdb = run.GetRuntimeDb() 
+run.SetUserConfig("g4Config.C") # user configuration file default g4Config.C
+rtdb = run.GetRuntimeDb()
 # -----Create geometry----------------------------------------------
 # import shipMuShield_only as shipDet_conf # special use case for an attempt to convert active shielding geometry for use with FLUKA
 # import shipTarget_only as shipDet_conf
 import shipDet_conf
-
 modules = shipDet_conf.configure(run,ship_geo)
 # -----Create PrimaryGenerator--------------------------------------
 primGen = ROOT.FairPrimaryGenerator()
 if simEngine == "Pythia8":
- primGen.SetTarget(ship_geo.target.z0, 0.) 
+ primGen.SetTarget(ship_geo.target.z0, 0.)
 # -----Pythia8--------------------------------------
  if HNL or options.RPVSUSY:
   P8gen = ROOT.HNLPythia8Generator()
@@ -301,7 +297,7 @@ if simEngine == "Pythia8":
    pythia8_conf.configurerpvsusy(P8gen,options.theMass,[theCouplings[0],theCouplings[1]],
                                 theCouplings[2],options.RPVSUSYbench,inclusive,options.deepCopy)
   P8gen.SetParameters("ProcessLevel:all = off")
-  if inputFile: 
+  if inputFile:
    ut.checkFileExists(inputFile)
 # read from external file
    P8gen.UseExternalFile(inputFile, options.firstEvent)
@@ -314,7 +310,7 @@ if simEngine == "Pythia8":
   import pythia8darkphoton_conf
   passDPconf = pythia8darkphoton_conf.configure(P8gen,options.theMass,options.theDPepsilon,inclusive, motherMode, options.deepCopy)
   if (passDPconf!=1): sys.exit()
- if HNL or options.RPVSUSY or options.DarkPhoton: 
+ if HNL or options.RPVSUSY or options.DarkPhoton:
   P8gen.SetSmearBeam(1*u.cm) # finite beam size
   P8gen.SetLmin((ship_geo.Chamber1.z - ship_geo.chambers.Tub1length) - ship_geo.target.z0 )
   P8gen.SetLmax(ship_geo.TrackStation1.z - ship_geo.target.z0 )
@@ -322,7 +318,7 @@ if simEngine == "Pythia8":
   primGen.SetTarget(0., 0.) #vertex is set in pythia8Generator
   ut.checkFileExists(inputFile)
   if ship_geo.Box.gausbeam:
-   primGen.SetBeam(0.,0., 0.5, 0.5) #more central beam, for hits in downstream detectors    
+   primGen.SetBeam(0.,0., 0.5, 0.5) #more central beam, for hits in downstream detectors
    primGen.SmearGausVertexXY(True) #sigma = x
   else:
    primGen.SetBeam(0.,0., ship_geo.Box.TX-1., ship_geo.Box.TY-1.) #Uniform distribution in x/y on the target (0.5 cm of margin at both sides)
@@ -344,7 +340,7 @@ if simEngine == "FixedTarget":
  primGen.AddGenerator(P8gen)
 if simEngine == "Pythia6":
 # set muon interaction close to decay volume
- primGen.SetTarget(ship_geo.target.z0+ship_geo.muShield.length, 0.) 
+ primGen.SetTarget(ship_geo.target.z0+ship_geo.muShield.length, 0.)
 # -----Pythia6-------------------------
  test = ROOT.TPythia6() # don't know any other way of forcing to load lib
  P6gen = ROOT.tPythia6Generator()
@@ -367,7 +363,7 @@ if simEngine == "EvtCalc":
     )
 
 # -----Particle Gun-----------------------
-if simEngine == "PG": 
+if simEngine == "PG":
   myPgun = ROOT.FairBoxGenerator(options.pID,1)
   myPgun.SetPRange(options.Estart,options.Eend)
   myPgun.SetPhiRange(0, 360) # // Azimuth angle range [degree]
@@ -377,9 +373,8 @@ if simEngine == "PG":
 # -----muon DIS Background------------------------
 if simEngine == "muonDIS":
  ut.checkFileExists(inputFile)
- primGen.SetTarget(0., 0.) 
+ primGen.SetTarget(0., 0.)
  DISgen = ROOT.MuDISGenerator()
- 
  # from nu_tau detector to tracking station 2
  # mu_start, mu_end =  ship_geo.tauMudet.zMudetC,ship_geo.TrackStation2.z
  #
@@ -390,8 +385,6 @@ if simEngine == "muonDIS":
  DISgen.Init(inputFile,options.firstEvent)
  primGen.AddGenerator(DISgen)
  options.nEvents = min(options.nEvents,DISgen.GetNevents())
- inactivateMuonProcesses = False 
- inactivateMuonNuclearProcess = True #avoid double counting
  print('Generate ',options.nEvents,' with DIS input', ' first event',options.firstEvent)
 # -----neutrino interactions from nuage------------------------
 if simEngine == "Nuage":
@@ -410,7 +403,7 @@ if simEngine == "Nuage":
  nZcells = ntt -1
  startx = -ship_geo.NuTauTarget.xdim/2. + nXcells*ship_geo.NuTauTarget.BrX
  endx = -ship_geo.NuTauTarget.xdim/2. + (nXcells+1)*ship_geo.NuTauTarget.BrX
- starty = -ship_geo.NuTauTarget.ydim/2. + nYcells*ship_geo.NuTauTarget.BrY 
+ starty = -ship_geo.NuTauTarget.ydim/2. + nYcells*ship_geo.NuTauTarget.BrY
  endy = - ship_geo.NuTauTarget.ydim/2. + (nYcells+1)*ship_geo.NuTauTarget.BrY
  startz = ship_geo.EmuMagnet.zC - ship_geo.NuTauTarget.zdim/2. + ntt *ship_geo.NuTauTT.TTZ + nZcells * ship_geo.NuTauTarget.CellW
  endz = ship_geo.EmuMagnet.zC - ship_geo.NuTauTarget.zdim/2. + ntt *ship_geo.NuTauTT.TTZ + nZcells * ship_geo.NuTauTarget.CellW + ship_geo.NuTauTarget.BrZ
@@ -428,7 +421,7 @@ if simEngine == "Genie":
  ut.checkFileExists(inputFile)
  primGen.SetTarget(0., 0.) # do not interfere with GenieGenerator
  Geniegen = ROOT.GenieGenerator()
- Geniegen.Init(inputFile,options.firstEvent) 
+ Geniegen.Init(inputFile,options.firstEvent)
  Geniegen.SetPositions(ship_geo.target.z0, ship_geo.tauMudet.zMudetC-5*u.m, ship_geo.TrackStation2.z)
  primGen.AddGenerator(Geniegen)
  options.nEvents = min(options.nEvents,Geniegen.GetNevents())
@@ -438,7 +431,7 @@ if simEngine == "nuRadiography":
  ut.checkFileExists(inputFile)
  primGen.SetTarget(0., 0.) # do not interfere with GenieGenerator
  Geniegen = ROOT.GenieGenerator()
- Geniegen.Init(inputFile,options.firstEvent) 
+ Geniegen.Init(inputFile,options.firstEvent)
  # Geniegen.SetPositions(ship_geo.target.z0, ship_geo.target.z0, ship_geo.MuonStation3.z)
  Geniegen.SetPositions(ship_geo.target.z0, ship_geo.tauMudet.zMudetC, ship_geo.MuonStation3.z)
  Geniegen.NuOnly()
@@ -449,7 +442,7 @@ if simEngine == "nuRadiography":
  pdg.AddParticle('W','Ion', 1.71350e+02, True, 0., 74, 'XXX', 1000741840)
 #
  run.SetPythiaDecayer('DecayConfigPy8.C')
- # this requires writing a C macro, would have been easier to do directly in python! 
+ # this requires writing a C macro, would have been easier to do directly in python!
  # for i in [431,421,411,-431,-421,-411]:
  # ROOT.gMC.SetUserDecay(i) # Force the decay to be done w/external decayer
 if simEngine == "Ntuple":
@@ -463,10 +456,10 @@ if simEngine == "Ntuple":
  print('Process ',options.nEvents,' from input file')
 #
 if simEngine == "MuonBack":
-# reading muon tracks from previous Pythia8/Geant4 simulation with charm replaced by cascade production 
+# reading muon tracks from previous Pythia8/Geant4 simulation with charm replaced by cascade production
  fileType = ut.checkFileExists(inputFile)
  if fileType == 'tree':
- # 2018 background production 
+ # 2018 background production
   primGen.SetTarget(ship_geo.target.z0+70.845*u.m,0.)
  else:
   primGen.SetTarget(ship_geo.target.z0+50*u.m,0.)
@@ -486,20 +479,20 @@ if simEngine == "MuonBack":
  options.nEvents = min(options.nEvents,MuonBackgen.GetNevents())
  MCTracksWithHitsOnly = True # otherwise, output file becomes too big
  print('Process ',options.nEvents,' from input file, with Phi random=',options.phiRandom, ' with MCTracksWithHitsOnly',MCTracksWithHitsOnly)
- if options.followMuon :  
+ if options.followMuon :
     options.fastMuon = True
     modules['Veto'].SetFollowMuon()
  if options.fastMuon :    modules['Veto'].SetFastMuon()
 
  # optional, boost gamma2muon conversion
- # ROOT.kShipMuonsCrossSectionFactor = 100. 
+ # ROOT.kShipMuonsCrossSectionFactor = 100.
 #
 if simEngine == "Cosmics":
  primGen.SetTarget(0., 0.)
  Cosmicsgen = ROOT.CosmicsGenerator()
  import CMBG_conf
  CMBG_conf.configure(Cosmicsgen, ship_geo)
- if not Cosmicsgen.Init(Opt_high): 
+ if not Cosmicsgen.Init(Opt_high):
       print("initialization of cosmic background generator failed ",Opt_high)
       sys.exit(0)
  Cosmicsgen.n_EVENTS = options.nEvents
@@ -525,17 +518,17 @@ if MCTracksWithHitsOnly:
 elif MCTracksWithEnergyCutOnly:
  fStack.SetMinPoints(-1)
  fStack.SetEnergyCut(100.*u.MeV)
-elif MCTracksWithHitsOrEnergyCut: 
+elif MCTracksWithHitsOrEnergyCut:
  fStack.SetMinPoints(1)
  fStack.SetEnergyCut(100.*u.MeV)
-elif options.deepCopy: 
+elif options.deepCopy:
  fStack.SetMinPoints(0)
  fStack.SetEnergyCut(0.*u.MeV)
 
 if options.eventDisplay:
  # Set cuts for storing the trajectories, can only be done after initialization of run (?!)
   trajFilter = ROOT.FairTrajFilter.Instance()
-  trajFilter.SetStepSizeCut(1*u.mm);  
+  trajFilter.SetStepSizeCut(1*u.mm)
   trajFilter.SetVertexCut(-20*u.m, -20*u.m,ship_geo.target.z0-1*u.m, 20*u.m, 20*u.m, 200.*u.m)
   trajFilter.SetMomentumCutP(0.1*u.GeV)
   trajFilter.SetEnergyCut(0., 400.*u.GeV)
@@ -561,7 +554,7 @@ if options.debug == 1:
 #fieldMaker.plotField(1, ROOT.TVector3(-9000.0, 6000.0, 50.0), ROOT.TVector3(-300.0, 300.0, 6.0), 'Bzx.png')
 #fieldMaker.plotField(2, ROOT.TVector3(-9000.0, 6000.0, 50.0), ROOT.TVector3(-400.0, 400.0, 6.0), 'Bzy.png')
 
-if inactivateMuonProcesses :
+if inactivateMuonProcesses:
  ROOT.gROOT.ProcessLine('#include "Geant4/G4ProcessTable.hh"')
  mygMC = ROOT.TGeant4.GetMC()
  mygMC.ProcessGeantCommand("/process/inactivate muPairProd")
@@ -573,10 +566,11 @@ if inactivateMuonProcesses :
  gProcessTable = ROOT.G4ProcessTable.GetProcessTable()
  procmu = gProcessTable.FindProcess(ROOT.G4String('muIoni'),ROOT.G4String('mu+'))
  procmu.SetVerboseLevel(2)
-if inactivateMuonNuclearProcess:
-  ROOT.gROOT.ProcessLine('#include "Geant4/G4ProcessTable.hh"')
-  mygMC = ROOT.TGeant4.GetMC()
-  mygMC.ProcessGeantCommand("/process/inactivate muonNuclear")
+
+if simEngine == "muonDIS":
+    ROOT.gROOT.ProcessLine('#include "Geant4/G4ProcessTable.hh"')
+    mygMC = ROOT.TGeant4.GetMC()
+    mygMC.ProcessGeantCommand("/process/inactivate muonNuclear")
 # -----Start run----------------------------------------------------
 run.Run(options.nEvents)
 # -----Runtime database---------------------------------------------
@@ -600,7 +594,7 @@ if options.debug == 2:
  fGeo.CheckOverlaps(0.1)  # 1 micron takes 5minutes
  fGeo.PrintOverlaps()
  # check subsystems in more detail
- for x in fGeo.GetTopNode().GetNodes(): 
+ for x in fGeo.GetTopNode().GetNodes():
    x.CheckOverlaps(0.0001)
    fGeo.PrintOverlaps()
 # -----Finish-------------------------------------------------------
@@ -611,11 +605,11 @@ print(' ')
 print("Macro finished successfully.")
 if "P8gen" in globals() :
     if (HNL): print("number of retries, events without HNL ",P8gen.nrOfRetries())
-    elif (options.DarkPhoton): 
+    elif (options.DarkPhoton):
         print("number of retries, events without Dark Photons ",P8gen.nrOfRetries())
         print("total number of dark photons (including multiple meson decays per single collision) ",P8gen.nrOfDP())
 
-print("Output file is ",  outFile) 
+print("Output file is ",  outFile)
 print("Parameter file is ",parFile)
 print("Real time ",rtime, " s, CPU time ",ctime,"s")
 
@@ -637,11 +631,11 @@ if simEngine == "MuonBack":
  nEvents = 0
  pointContainers = []
  for x in sTree.GetListOfBranches():
-   name = x.GetName() 
+   name = x.GetName()
    if not name.find('Point')<0: pointContainers.append('sTree.'+name+'.GetEntries()') # makes use of convention that all sensitive detectors fill XXXPoint containers
  for n in range(t.GetEntries()):
      rc = t.GetEvent(n)
-     empty = True 
+     empty = True
      for x in pointContainers:
         if eval(x)>0: empty = False
      if not empty:
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -373,11 +373,6 @@ if simEngine == "muonDIS":
  ut.checkFileExists(inputFile)
  primGen.SetTarget(0., 0.)
  DISgen = ROOT.MuDISGenerator()
- #  add Carbon to PDG
- pdg = ROOT.TDatabasePDG.Instance()
- pdg.AddParticle("C12", "Carbon-12", 12.0, True, 0, 6.0, "nucleus", 1000060120)
- pdg.AddParticle("C13", "Carbon-13", 13.003355, True, 0, 6.0, "nucleus", 1000060130)
-
  # from nu_tau detector to tracking station 2
  # mu_start, mu_end =  ship_geo.tauMudet.zMudetC,ship_geo.TrackStation2.z
  #
@@ -515,15 +510,17 @@ if options.dryrun: # Early stop after setting up Pythia 8
  sys.exit(0)
 gMC = ROOT.TVirtualMC.GetMC()
 fStack = gMC.GetStack()
+EnergyCut = 10.*u.MeV if options.mudis else 100.*u.MeV
+
 if MCTracksWithHitsOnly:
  fStack.SetMinPoints(1)
  fStack.SetEnergyCut(-100.*u.MeV)
 elif MCTracksWithEnergyCutOnly:
  fStack.SetMinPoints(-1)
- fStack.SetEnergyCut(100.*u.MeV)
+ fStack.SetEnergyCut(EnergyCut)
 elif MCTracksWithHitsOrEnergyCut:
  fStack.SetMinPoints(1)
- fStack.SetEnergyCut(100.*u.MeV)
+ fStack.SetEnergyCut(EnergyCut)
 elif options.deepCopy:
  fStack.SetMinPoints(0)
  fStack.SetEnergyCut(0.*u.MeV)
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+from __future__ import print_function
+from __future__ import division
 import os
 import sys
 import ROOT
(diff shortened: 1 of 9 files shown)


muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
+import ROOT as r
+import os
+import argparse
+import math
```
olantwin, line 4: Remove unused import

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+from ShipGeoConfig import ConfigRegistry
+import shipunit as u
+
+r.PyConfig.IgnoreCommandLineOptions = True
+
+# Histogram
+hPmuon = r.TH1F('hPmuon', 'Momentum of muons hitting the SBT;P [GeV];Entries', 400, 0., 400)
+
+# Argument parser
+parser = argparse.ArgumentParser(description='Script to collect muons hitting the Tracking Station')
```
olantwin, line 15:
```suggestion
parser = argparse.ArgumentParser(description=__doc__)
```

muonDIS/make_nTuple_Tr.py
olantwin, line 1: Are the SBT and Tr cases so different that two different scripts are needed?

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+
+if args.testing_code: 
+    print("test code, output file name overwritten as: muonsProduction_wsoft_Tr_test.root")
+    args.outputfile='muonsProduction_wsoft_Tr_test.root'
+    selectedmuons='SelectedMuonsTr_test.txt'
+else:
+    selectedmuons='SelectedMuonsTr.txt'
+
+# Output file and tree setup
+output_file = r.TFile(args.outputfile, 'recreate')
```
olantwin, line 35:
```suggestion
output_file = r.TFile.Open(args.outputfile, 'recreate')
```

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+
+muon_vetoPoints    = r.TClonesArray("vetoPoint")
+output_tree.Branch('muon_vetoPoints', muon_vetoPoints)
+
+path = args.path
+
+fsel = open(selectedmuons, "w")
+
+# Process each input file
+for inputFile in os.listdir(path):
```
olantwin, line 52: Comments from other script apply

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+        print(f"Error: {e}")
+        continue
+
+    for event in tree:
+        
+        ev += 1
+        
+        #saving soft tracks
+        track_array.Clear()
+        muon_id = next((itrk for itrk in range(event.MCTrack.GetEntries()) if abs(event.MCTrack[itrk].GetPdgCode()) == 13), None)
```
olantwin, line 70: cryptic

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+fsel.close()
+
+
+print("------------------------------------------------------file saved, reading",args.outputfile, " now----------------------------------------------------------------")
+
+
+file = r.TFile(args.outputfile, 'read')
+if not file or file.IsZombie():
+    print(f"Failed to open file ")
+    exit(1)
```
olantwin, line 156:
```suggestion
file = r.TFile.Open(args.outputfile, 'read')
```

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+
+tree = file.Get("MuonAndSoftInteractions")
```
olantwin, line 158:
```suggestion
tree = file.MuonAndSoftInteractions
```

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+if not tree:
+    print(f"Tree 'MuonAndSoftInteractions' not found in file")
+    file.Close()
+    exit(1)
```
olantwin, line 162: Probably redundant

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+    time_hit = imuondata[8]
+
+    num_tracks = len(event.tracks) 
+    
+    print(f"Muon PID: {pid},  x: {x}, y: {y}, z: {z}, t_muon: {time_hit}, Number of soft tracks in this event: {num_tracks}")
+    
+
+file.Close()
+
+#python make_nTuple_Tr.py -p <path to muonbackground files>
```
olantwin, line 187: Remove, make sure --help works instead

shipgen/MuDISGenerator.cxx
```diff
@@ -12,14 +12,16 @@
 #include "TGeoEltu.h"
 #include "TVectorD.h"
 #include "TGeoCompositeShape.h"
+#include "FairLogger.h"   // for FairLogger, etc
```
olantwin, line 15:
```suggestion
#include "FairLogger.h"
```

shipgen/MuDISGenerator.cxx
```diff
@@ -12,14 +12,16 @@
...
+
+
+
+const Double_t c_light = 29.9792458;//speed of light in cm/ns
```
olantwin, line 19:
```suggestion
const Double_t c_light = 29.9792458;  // speed of light in cm/ns
```

shipgen/MuDISGenerator.cxx
```diff
@@ -28,31 +30,46 @@ MuDISGenerator::MuDISGenerator() {}
...
+    dPart = 0; 
+    dPartSoft = 0; 
+
+    if (0 == strncmp("/eos",fileName,4)) {
+        TString tmp = gSystem->Getenv("EOSSHIP");
+        tmp += fileName;
+        fInputFile = TFile::Open(tmp); 
+        LOGF(info, "Open external file on eos: %s", tmp.Data());
+    } else {
+        fInputFile = new TFile(fileName);
```
olantwin, line 46:
```suggestion
        fInputFile = TFile::Open(fileName, "read");
```

shipgen/MuDISGenerator.cxx
```diff
@@ -28,31 +30,46 @@ MuDISGenerator::MuDISGenerator() {}
...
+    LOGF(info, "Opening input file %s", fileName);
+
+    iMuon = 0;
+    dPart = 0; 
+    dPartSoft = 0; 
+
+    if (0 == strncmp("/eos",fileName,4)) {
+        TString tmp = gSystem->Getenv("EOSSHIP");
+        tmp += fileName;
+        fInputFile = TFile::Open(tmp);
```
olantwin, line 43:
```suggestion
        fInputFile = TFile::Open(tmp, "read"); 
```

shipgen/MuDISGenerator.cxx
```diff
@@ -28,31 +30,46 @@ MuDISGenerator::MuDISGenerator() {}
...
+        fInputFile = TFile::Open(tmp); 
+        LOGF(info, "Open external file on eos: %s", tmp.Data());
+    } else {
+        fInputFile = new TFile(fileName);
+    }
+
+    if (!fInputFile || fInputFile->IsZombie()) {
+        LOG(FATAL) << "Error opening input file";
+        return kFALSE;
+    }
```
olantwin, line 52: Not necessary when using TFile::Open

shipgen/MuDISGenerator.cxx
```diff
@@ -28,31 +30,46 @@ MuDISGenerator::MuDISGenerator() {}
...
+
+    fTree = (TTree *)fInputFile->Get("DIS");
```
olantwin, line 54: dynamic_cast instead of C-style cast

shipgen/MuDISGenerator.cxx
```diff
@@ -216,36 +234,42 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
+    if (fn%10==0) {
+      LOG(INFO) << "Info MuDISGenerator: MuDIS event-nr "<< fn;
     }
+    
+
     int nf = dPart->GetEntries();
-    //cout << "*********************************************************" << endl;
-    //cout << "muon DIS Generator debug " << iMuon->GetEntries()<< " "<< iMuon->AddrAt(0) << " nf "<< nf << " fn=" << fn <<endl;
+    int ns = dPartSoft->GetEntries(); 
+    //LOG(DEBUG) << "*********************************************************" ;
```
olantwin, line 244: Uncomment or remove

shipgen/MuDISGenerator.cxx
```diff
@@ -216,36 +234,42 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
 
     //some start/end positions in z (f.i. emulsion to Tracker 1)
     Double_t start[3]={0.,0.,startZ};
     Double_t end[3]={0.,0.,endZ};
 
-// incoming muon  array('d',[pid,px,py,pz,E,x,y,z,w])
+// incoming muon  array('d',[pid,px,py,pz,E,x,y,z,w,t])
     TVectorD* mu = dynamic_cast<TVectorD*>(iMuon->AddrAt(0));
-    //cout << "muon DIS Generator in muon " << int(mu[0][0])<< endl;
+    //LOG(DEBUG) << "muon DIS Generator in muon " << int(mu[0][0]);
```
olantwin, line 253: uncomment or remove

shipgen/MuDISGenerator.cxx
```diff
@@ -216,36 +234,42 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
-    //cout << "MuDIS: mu xyz position " << x << ", " << y << ", " << z << endl;
-    //cout << "MuDIS: mu pxyz position " << mu[0][1] << ", " << mu[0][2] << ", " << mu[0][3] << endl;
-    //cout << "MuDIS: mu weight " << w << endl;
-    //cout << "MuDIS: start position " << start[0] << ", " << start[1] << ", " << start[2] << endl;
-    //cout << "MuDIS: end position " << end[0] << ", " << end[1] << ", " << end[2] << endl;
+    //LOG(INFO) << "MuDIS: mu xyz position " << x << ", " << y << ", " << z;
+    //LOG(INFO) << "MuDIS: mu pxyz position " << mu[0][1] << ", " << mu[0][2] << ", " << mu[0][3];
+    //LOG(INFO) << "MuDIS: mu weight " << w;
+    //LOG(INFO) << "MuDIS: start position " << start[0] << ", " << start[1] << ", " << start[2];
+    //LOG(INFO) << "MuDIS: end position " << end[0] << ", " << end[1] << ", " << end[2];
```
olantwin, line 272: Maybe make DEBUG and uncomment. Otherwise, remove.

shipgen/MuDISGenerator.cxx
```diff
@@ -255,8 +279,10 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
     Double_t ymu;
     Double_t zmu;
     Int_t count=0;
-    //cout << "Info MuDISGenerator Start prob2int while loop, bparam= " << bparam << ", " << bparam*1.e8 <<endl;
-    //cout << "Info MuDISGenerator What was maximum density, mparam[7]= " << mparam[7] << ", " << mparam[7]*1.e8 <<endl;
+    //LOG(INFO) << "Info MuDISGenerator Start prob2int while loop, bparam= " << bparam << ", " << bparam*1.e8 ;
+    //LOG(INFO) << "Info MuDISGenerator What was maximum density, mparam[7]= " << mparam[7] << ", " << mparam[7]*1.e8;
+
```
olantwin, line 284: Make DEBUG and uncomment or remove.

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
       TGeoNode *node = gGeoManager->FindNode(xmu,ymu,zmu);
       TGeoMaterial *mat = 0;
       if (node && !gGeoManager->IsOutside()) mat = node->GetVolume()->GetMaterial();
-      //cout << "Info MuDISGenerator: mat " <<  count << ", " << mat->GetName() << ", " << mat->GetDensity() << endl;
+      //LOG(DEBUG) << "Info MuDISGenerator: mat " <<  count << ", " << mat->GetName() << ", " << mat->GetDensity();
```
olantwin, line 294: Uncomment

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
       //density relative to Prob largest density along this trajectory, i.e. use rho(Pt)
       prob2int= mat->GetDensity()/mparam[7];
-      if (prob2int>1.) cout << "***WARNING*** MuDISGenerator: prob2int > Maximum density????" << prob2int << " maxrho:" << mparam[7] << " material: " <<  mat->GetName() << endl;
+      if (prob2int>1.) LOG(WARNING) << "***WARNING*** MuDISGenerator: prob2int > Maximum density????" << prob2int << " maxrho:" << mparam[7] << " material: " <<  mat->GetName();
```
olantwin, line 297:
```suggestion
      if (prob2int>1.) LOG(WARNING) << "MuDISGenerator: prob2int > Maximum density????" << prob2int << " maxrho:" << mparam[7] << " material: " <<  mat->GetName();
```
Also, single statement per line please.

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
-    cpg->AddTrack(int(mu[0][0]),mu[0][1],mu[0][2],mu[0][3],xmu,ymu,zmu,-1,false,mu[0][4],0.,w);
-    // in order to have a simulation of possible veto detector hits, let Geant4 follow the muon backward
-    // due to dE/dx, as soon as the muon travers thick material, this approximation will become bad.
-    // a proper implementation however would be to have this better integrated in Geant4, follow the muon, call DIS event, continue
-    cpg->AddTrack(int(mu[0][0]),-mu[0][1],-mu[0][2],-mu[0][3],xmu,ymu,zmu,0,true,mu[0][4],0.,w);
-// outgoing particles, [did,dpx,dpy,dpz,E], put density along trajectory as weight, g/cm^2
-    w=mparam[0]*mparam[4];
+    
+    //LOG(DEBUG) << "Info MuDISGenerator: prob2int " << prob2int << ", " << count;
+    //LOG(DEBUG) << "MuDIS: put position " << xmu << ", " << ymu << ", " << zmu ;
```
olantwin, line 302: Uncomment

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
-         //cout << "muon DIS Generator out part mom " << Part[0][1]<<" " << Part[0][2] <<" " << Part[0][3] << " " << Part[0][4] << endl;
-         //cout << "muon DIS Generator out part pos " << mu[0][5]<<" " << mu[0][6] <<" " << mu[0][7] << endl;
-         //cout << "muon DIS Generator out part w " << mu[0][8] << endl;
-         cpg->AddTrack(int(Part[0][0]),Part[0][1],Part[0][2],Part[0][3],xmu,ymu,zmu,0,true,Part[0][4],0.,w);
-         //cout << "muon DIS part added "<<endl;
-       }
+         //LOG(DEBUG)<< "muon DIS Generator out part " << int(Part[0][0]); 
+         //LOG(DEBUG) << "muon DIS Generator out part mom " << Part[0][1]<<" " << Part[0][2] <<" " << Part[0][3] << " " << Part[0][4]; 
+         //LOG(DEBUG) << "muon DIS Generator out part pos " << mu[0][5]<<" " << mu[0][6] <<" " << mu[0][7]; 
+         //LOG(DEBUG) << "muon DIS Generator out part w " << mu[0][8] ;
```
olantwin, line 334: Uncomment

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
+    }
+
+    Double_t t_DIS=(t_muon+t_rmu)/1e9; //time taken in seconds to reach [xmu,ymu,zmu]
+    
+    cpg->AddTrack(int(mu[0][0]),mu[0][1],mu[0][2],mu[0][3],xmu,ymu,zmu,-1,false,mu[0][4],t_DIS,w);//set time of the track start as t_muon from the input file
+
+    // outgoing particles, [did,dpx,dpy,dpz,E], put density along trajectory as weight, g/cm^2
+    w=mparam[0]*mparam[4];//modify weight, by multiplying with average densiy along track??
+    
     for(int i=0; i<nf; i++)
```
olantwin, line 328: Might be better to do a range-for loop over the particles container

shipgen/MuDISGenerator.cxx
```diff
@@ -265,37 +291,67 @@ Bool_t MuDISGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
+         //LOG(DEBUG)<< "muon DIS Generator out part " << int(Part[0][0]); 
+         //LOG(DEBUG) << "muon DIS Generator out part mom " << Part[0][1]<<" " << Part[0][2] <<" " << Part[0][3] << " " << Part[0][4]; 
+         //LOG(DEBUG) << "muon DIS Generator out part pos " << mu[0][5]<<" " << mu[0][6] <<" " << mu[0][7]; 
+         //LOG(DEBUG) << "muon DIS Generator out part w " << mu[0][8] ; 
+         cpg->AddTrack(int(Part[0][0]),Part[0][1],Part[0][2],Part[0][3],xmu,ymu,zmu,0,true,Part[0][4],t_DIS,w);        }
+    
+    //Soft interaction tracks
+    
+
+    for (int i = 0; i < ns; i++) {
```
olantwin, line 340: Prefer doing a range-for loop over the container

shipgen/MuDISGenerator.h
```diff
@@ -41,12 +41,17 @@ class MuDISGenerator : public FairGenerator
...
+  TClonesArray* dPartSoft ;
   FairLogger*  fLogger; //!   don't make it persistent, magic ROOT command
   TFile* fInputFile;
   TTree* fTree;
   int fNevents;
   int fn;
   bool fFirst;
+
+  const Double_t c_light = 29.9792458;//speed of light in cm/ns
+  const Double_t muon_mass = 0.10565999895334244;  //muon mass in GeV
```
olantwin, line 53: If these are defined here, they are not needed in the implementation file.
If they are not needed in the header, it would be good to delete them here.

olantwin (changes requested):
Very nice improvements. Generally looking good, but some nitpicks.

Once all other issues are resolved, it would be good to run clang-format and ruff over the touched code.

I would suggest fixing issues with --fixup commits, and then at the end we can do a rebase and automatically squash the fixes.
You might want to check out [git-absorb](https://github.com/tummychow/git-absorb) to make that process a bit simpler.

olantwin:
> _ No description provided. _

Could you add some explanation to your first post, maybe with the before/after plots you sent me?

olantwin:
This change would also benefit from a description in the CHANGELOG (probably an entire paragraph!)

olantwin:
@maksymovchynnikov Could you please have a look and ideally test these changes? It would be good to merge these ASAP.

olantwin:
@maksymovchynnikov Your monthly reminder that your feedback would be very welcome on this pull request.

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,204 @@
...
+            if a == args.nDIS // 2:
+                myPythia.Initialize('FIXT', mutype[pid], 'n0', p)
+                isProton = 0
+                target='n0'
+                print(f"Switching to neutron interaction")
+
+            dPartDIS.Clear()
+            iMuon.Clear()
+            muPart[9] = isProton
+            iMuon[0] = muPart
```
olantwin, line 127: This should be adapted for ROOT 6.32 as done in #559

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,204 @@
...
+                did = myPythia.GetK(itrk, 2)
+                dpx, dpy, dpz = rotate(ctheta, stheta, cphi, sphi, myPythia.GetP(itrk, 1), myPythia.GetP(itrk, 2), myPythia.GetP(itrk, 3))
+                psq = dpx ** 2 + dpy ** 2 + dpz ** 2
+                E = r.TMath.Sqrt(getMasssq(did) + psq)
+                m = array('d', [did, dpx, dpy, dpz, E])
+                part = r.TVectorD(5, m)
+                nPart = dPartDIS.GetEntries()
+                if dPartDIS.GetSize() == nPart:
+                    dPartDIS.Expand(nPart + 10)
+                dPartDIS[nPart] = part
```
olantwin, line 145: This should be adapted for ROOT 6.32 as done in https://github.com/ShipSoft/FairShip/pull/559

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,204 @@
...
+                    dPartSoft.Expand(nPart + 10)
+                dPartSoft[nPart] = part
+            
+            muon_vetoPoints.Clear()
+            index=0
+
+            for hit in sTree.muon_vetoPoints:
+                if muon_vetoPoints.GetSize() == index:
+                    muon_vetoPoints.Expand(index+1)
+                muon_vetoPoints[index] = hit
```
olantwin, line 187: This should be adapted for ROOT 6.32 as done in https://github.com/ShipSoft/FairShip/pull/559

muonDIS/make_nTuple_Tr.py
```diff
@@ -0,0 +1,189 @@
...
+        index=0
+        muon_vetoPoints.Clear()
+        for hit in event.vetoPoint:
+            
+            detID = hit.GetDetectorID()
+            pid = hit.PdgCode()
+            if 1000 < detID < 999999 and abs(pid) == 13:
+                if muon_vetoPoints.GetSize() == index:
+                    muon_vetoPoints.Expand(index+1)
+                muon_vetoPoints[index] = hit
```
olantwin, line 85: This should be adapted for ROOT 6.32 as done in https://github.com/ShipSoft/FairShip/pull/559

muonDIS/make_nTuple_SBT.py
```diff
@@ -0,0 +1,179 @@
...
+    if args.testing_code and ev >= 100000:
+        break
+
+    f = r.TFile.Open(os.path.join(path, inputFile), 'read')
+
+    if not f or f.IsZombie():
+        print(f"Failed to open file {inputFile}")
+        continue
+
+    tree = f.Get("cbmsim")
```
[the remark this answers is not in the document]
olantwin: Then it would be good to catch the error using a try/except.

macro/run_simScript.py
```diff
@@ -407,7 +410,8 @@
  DISgen.Init(inputFile,options.firstEvent)
  primGen.AddGenerator(DISgen)
  options.nEvents = min(options.nEvents,DISgen.GetNevents())
- inactivateMuonProcesses = True # avoid unwanted hadronic events of "incoming" muon flying backward
+ inactivateMuonProcesses = False
```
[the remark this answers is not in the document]
anupama-reghunath: yes, but also this flag is never True. Think we should remove the it completely?

muonDIS/add_muonresponse.py
```diff
@@ -0,0 +1,71 @@
...
+import argparse
+
+parser = argparse.ArgumentParser(description='Script to collect muons hitting the SBT')
+parser.add_argument('-f', '--inputfile', default='ship.conical.muonDIS-TGeant4.root')
+parser.add_argument('-m', '--muonfile', default='muonDis_0.root')
+args = parser.parse_args()
+
+# Open the existing ROOT file in update mode
+inputFile = r.TFile.Open(args.inputfile, "UPDATE") 
+sTree = inputFile.Get("cbmsim")
```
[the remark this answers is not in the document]
anupama-reghunath: No reason other than to maintain cohesivity with the rest of FairShip code which uses sTree.

muonDIS/add_muonresponse.py
```diff
@@ -0,0 +1,71 @@
...
+    newTree.Fill()
+
+# Write the updated tree to the output file
+newTree.Write()
+outputFile.Close()
+inputFile.Close()
+muonFile.Close()
+
+# Replace the old file with the new file
+os.replace('ship.conical.muonDIS-TGeant4_temp.root', args.inputfile)
```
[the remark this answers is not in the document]
anupama-reghunath: I'll open it up for reading in this case then.

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
olantwin: I think we don't need to be consistent with bad naming practices. If we remained completely consistent with old code, we would not be able to clean up anything.

I would prefer following usual python naming practices (short, self-descriptive names, snake_case) unless there is a good reason not to.

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
olantwin: I don't quite remember, but I think I meant that we might as well just update the input file, at least if it's local instead of writing a new file and overwriting.

muonDIS/add_muonresponse.py
```diff
@@ -0,0 +1,71 @@
...
+# Open the existing ROOT file in update mode
+inputFile = r.TFile.Open(args.inputfile, "UPDATE") 
+sTree = inputFile.Get("cbmsim")  
+
+# Open the external file with additional vetoPoints
+muonFile = r.TFile.Open(args.muonfile, "READ")  
+muonTree = muonFile.Get('DIS') 
+
+# Create a new file for the updated tree
+outputFile = r.TFile('ship.conical.muonDIS-TGeant4_temp.root', "RECREATE")
```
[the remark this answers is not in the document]
anupama-reghunath: Removed this entirely since the inputfile is now updated directly.

macro/run_simScript.py
```diff
@@ -596,10 +593,11 @@
...
  procmu.SetVerboseLevel(2)
-if inactivateMuonNuclearProcess:
-  ROOT.gROOT.ProcessLine('#include "Geant4/G4ProcessTable.hh"')
-  mygMC = ROOT.TGeant4.GetMC()
-  mygMC.ProcessGeantCommand("/process/inactivate muonNuclear")
+
+if simEngine == "muonDIS":
+    ROOT.gROOT.ProcessLine('#include "Geant4/G4ProcessTable.hh"')
+    mygMC = ROOT.TGeant4.GetMC()
+    mygMC.ProcessGeantCommand("/process/inactivate muonNuclear")
```
olantwin, line 600: Now that we don't include the original muon anymore, is this necessary? It should be fairly negligible that the outgoing muon undergoes DIS again and in the few cases where it does, I don't think it's an issue.

muonDIS/add_muonresponse.py
```diff
@@ -27,8 +27,8 @@
 inputFile = r.TFile.Open(args.inputfile, "READ")
 
 try:
-    sTree = inputFile.cbmsim
-except AttributeError as e:
+    inputTree = inputFile.cbmsim
```
olantwin, line 30: `input_tree` would be more pythonic

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
olantwin: Did you already commit this change? To me it seems like you are still creating a new file.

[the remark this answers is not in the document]
anupama-reghunath: Sorry, the commit is now pushed.

macro/run_simScript.py
anupama-reghunath: That is true, will remove this. maybe also the inactivateMuonProcesses, I dont find it being used anywhere anymore.

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,204 @@
...
+    mutype = {-13: 'gamma/mu+', 13: 'gamma/mu-'}
+
+    myPythia.SetMSTU(11, 11)
+    print(f"Processing events from {nStart} to {nEnd}...")
+
+    nMade = 0
+    
+    print(nStart, nEnd)
+    
+    for k in range(nStart, nEnd):
```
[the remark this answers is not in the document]
anupama-reghunath: yes, but the index helps in choosing a specific range of muons from the same muon file for parallelised DIS generation while using HTCondor.

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
olantwin: We might however have to be careful with files on EOS, updating them in place might not be ideal. Maybe we can add a warning/error to tell the user to perform the update locally?

macro/run_simScript.py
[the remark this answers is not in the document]
olantwin: Probably yes.

muonDIS/make_nTuple_SBT.py
```diff
@@ -0,0 +1,179 @@
...
+output_tree.Branch('imuondata', imuondata)
+
+track_array = r.TObjArray()
+output_tree.Branch('tracks', track_array)
+
+muon_vetoPoints    = r.TClonesArray("vetoPoint")
+output_tree.Branch('muon_vetoPoints', muon_vetoPoints)
+
+
+for inputFile in os.listdir(path):
```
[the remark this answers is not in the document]
anupama-reghunath: I've rewritten this to match the latest MuonBackground files produced "/eos/experiment/ship/simulation/bkg/MuonBack_2024helium/sc_v6_10_spills", and can be reused for any new production using the same condor scripts.

muonDIS/make_nTuple_SBT.py
```diff
@@ -0,0 +1,179 @@
...
+        continue
+
+    for event in tree:
+
+        ev += 1
+        numHitsPermuon = 0
+        
+        #saving soft tracks
+        track_array.Clear()
+        muon_id = next((itrk for itrk in range(event.MCTrack.GetEntries()) if abs(event.MCTrack[itrk].GetPdgCode()) == 13), None)
```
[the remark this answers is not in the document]
anupama-reghunath: The muon track index is saved in order to save all the soft interaction tracks( mother id ==index).

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
anupama-reghunath: yes, I'm also wondering of ways to block rewrite if it is already added (this will reapply the muon hits on the file if rerun). Perhaps a new naming? But this will break the default input file name for any existing analysis scripts.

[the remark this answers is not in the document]
olantwin: Could you check for the presence of a branch?
Alternatively, we could add some information to the file that could be checked, e.g. to the TaskList

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,204 @@
...
+    pyr = sphi * px1 + cphi * py
+    return pxr, pyr, pzr
+
+def makeMuonDIS():
+    """Main function to generate DIS events."""
+    parser = argparse.ArgumentParser(description='Script to generate DIS events')
+
+    parser.add_argument('-f', '--inputFile', help='Input file to use', required=True)
+    parser.add_argument('-nJob', '--nJob', type=int, help='Process ID, gives muon index', required=True)
+    parser.add_argument('-nPerJobs', '--nPerJobs', type=int, help='The number of muons per file', required=False)
```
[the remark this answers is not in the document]
anupama-reghunath: Set this as required =True now. These are always provided when making muonDIS.

muonDIS/makeMuonDIS.py
anupama-reghunath: Will be corrected later as a new PR.

muonDIS/make_nTuple_SBT.py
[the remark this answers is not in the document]
anupama-reghunath: This is now rewritten in a simpler way.

muonDIS/make_nTuple_SBT.py
```diff
@@ -0,0 +1,179 @@
...
+                        fsel.write(f"{pid} {hit.GetPx() / u.GeV} {hit.GetPy() / u.GeV} {hit.GetPz() / u.GeV} {hit.GetX() / u.m} {hit.GetY() / u.m} {hit.GetZ() / u.m} {weight}\n")
+
+                else:
+                    numHitsPermuon += 1
+                    
+        if numHitsPermuon != 0:
+            hnumSegPermmuon.Fill(numHitsPermuon)
+
+    f.Close()
+    gc.collect()
```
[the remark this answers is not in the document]
anupama-reghunath: can't remember anymore. removed it since it works fine without it :)

muonDIS/make_nTuple_Tr.py
anupama-reghunath: The Tr script only saves muons which hit the first tracking station(excludes muons hitting the SBT). There is no such selection used in the SBT script.

shipgen/MuDISGenerator.h
anupama-reghunath: They have been removed from the .cxx

muonDIS/add_muonresponse.py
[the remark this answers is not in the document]
anupama-reghunath: > We might however have to be careful with files on EOS, updating them in place might not be ideal. Maybe we can add a warning/error to tell the user to perform the update locally?

Since these are only intended for muonDIS files, should'nt this be allowed to edit in EOS?

anupama-reghunath:
@olantwin I have addressed the open comments. Can you please let me know if anything else is needed?

shipgen/MuDISGenerator.h
```diff
@@ -42,12 +42,17 @@ class MuDISGenerator : public FairGenerator
...
```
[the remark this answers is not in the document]
anupama-reghunath: Well, I'm not certain if these are available as constants elsewhere, I see that for GenieGenerator, c_light is used without definition for calculations. This could cause lack of clarity (timing bug in neuDIS), so I had them in the header file, instead of the implementation file.

muonDIS/make_nTuple_SBT.py
```diff
@@ -0,0 +1,202 @@
...
+    args.outputfile = "muonsProduction_wsoft_SBT_test.root"
+    selectedmuons = "SelectedMuonsSBT_test.txt"
+else:
+    selectedmuons = "SelectedMuonsSBT.txt"
+
+if args.path:
+    path = args.path
+else:
+    path = "/eos/experiment/ship/simulation/bkg/MuonBack_2024helium/sc_v6_10_spills"
+
```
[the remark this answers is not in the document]
anupama-reghunath: Better readability.

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,228 @@
...
+    parser.add_argument(
+        "-nJob", "--nJob", type=int, help="Process ID, gives muon index", required=True
+    )
+    parser.add_argument(
+        "-nPerJobs",
+        "--nPerJobs",
+        type=int,
+        help="The number of muons per file",
+        required=True,
+    )
```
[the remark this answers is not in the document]
anupama-reghunath: This maybe a bit tricky since to handle both SBT and Tr cases in the makeMuonDIS script, we adjust the generation by modifying the number of muons the script takes based on the case type. For the Tr case, where there are fewer muons, we limit the number of muons (essentially to 1 mostly), since nDISPerMuon ~1000. But for the SBT case, since we have a larger muon sample we allow more muons given that the nDISPerMuon ~100 or so. This allows for some appropriate balance in the final output as well as production time.  We  _could_ have a start event and end event instead of these, but this will require a rewrite of the existing algorithm and retesting.

shipgen/MuDISGenerator.h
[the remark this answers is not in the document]
olantwin: It's fine to have them in this header, the question was rather why these are member variables/attributes rather than normal variables.

muonDIS/make_nTuple_SBT.py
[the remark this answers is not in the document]
olantwin: I don't see how this improves readability, and like this the default value is not visible in the `--help` output

muonDIS/makeMuonDIS.py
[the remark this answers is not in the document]
olantwin: Maybe you can explain this in more detail at the CM, but this still seems like something better manages in e.g. the HTCondor submission scripts.

muonDIS/makeMuonDIS.py
```diff
@@ -0,0 +1,228 @@
...
+                E = r.TMath.Sqrt(masssq + psq)
+                m = array("d", [did, dpx, dpy, dpz, E])
+                part = r.TVectorD(5, m)
+                nPart = dPartDIS.GetEntries()
+                if dPartDIS.GetSize() == nPart:
+                    dPartDIS.Expand(nPart + 10)
+                # dPartDIS.ConstructedAt(nPart).Use(part) #to be adapted later
+                dPartDIS[nPart] = part
+                if itrk == 1:
+                    with open(f"sigmadata_{args.nJob}.txt", "a") as fcross:
```
anupama-reghunath, line 173: Reasoning for having an external textfile to save the DIS cross section from Pythia6:
1) As is, the DIS cross section otherwise is not saved anywhere
2) Consider generating N >>1 DIS events for the single muon. Since Pythia generates the cross-section value events-by-event, it converges from some meaningless value for the 1st event to the proper value for the Nth event. In the proper implementation, all the events must share the same correct cross-section. As a temporary fix,we achieve this by reading the Nth DIS cross section entry for a given muon to calculate the weight.

anupama-reghunath: My plan of action to avoid this external file is:

1.Directly write the cross section to the muonDis file ( which is passed as the input for muonDIS production)
2. Set weight of the scattered muon to the cross section in the simulation (atm both the incoming and scattered have the same weight from the spill).

Perhaps also set a minimum value for the number of DIS per muon to allow for convergence?

Please let me know what you think! @olantwin

anupama-reghunath: Heres a sample from 200 DIS events per muon.
Crosssections

olantwin: Ok, this is now much clearer, thanks!

I agree, that this information should be directly in the sample. Is there a weight field in the event header? Otherwise, using the weight of the scattered muon seems like an OK workaround (maybe we can preserve the weight of the incoming muon from the target sim).

We do not use the cross section anywhere else, right?

Could you prepare that plot also relative to the final cross section? That allow us to estimate a good default number of events to generate.

anupama-reghunath: > I agree, that this information should be directly in the sample. Is there a weight field in the event header? 
I couldnt find one in ShipEventHeader.
> Otherwise, using the weight of the scattered muon seems like an OK workaround (maybe we can preserve the weight of the incoming muon from the target sim).
At the moment I save the cross section onto the scattered muon and also refactor the incoming muon's weight to include the DIS multiplicity such that we maintain the event weight's normalisation to a spill. Previously this refactoring was explicitly done in the analysis, but the meta data (nDIS per muon) needed to be explicitly plugged in to calculate weight for any sim file.

> We do not use the cross section anywhere else, right?
> 
> Could you prepare that plot also relative to the final cross section? That allow us to estimate a good default number of events to generate.

olantwin:
@anupama-reghunath How shall we proceed with this PR?

muonDIS/make_nTuple_SBT.py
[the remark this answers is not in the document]
anupama-reghunath: this is now added as default.

muonDIS/makeMuonDIS.py
anupama-reghunath: I've removed the usage of an external file, and now save the DIS cross section to MCTrack[1] instead of scattered muon(simply out of convenience for easy calculation of weights).

muonDIS/make_nTuple_Tr.py
anupama-reghunath: @olantwin also to note, the current scripts for SBT and Tr  are used to produce the simulations. If we change this, we may also need to change the subsequent scripts since the nDIS per SBT muon < nDIS per Tr muon. What do you think is the right way to go?

anupama-reghunath:
Summary of the latest commits:

- An overlooked bug with saving the incoming muon's vetopoint information. The issue arises if we have multiple muons in the events and then the vetoPoint information muon hitting the SBT/Tr may not be correctly chosen. This has now been rectified.
- Adding Carbon onto the PDG Database in run_simScript.py (From Coulomb scattering of LiSc (Linear Alkyl Benzene))
- Lowering the MCTrack Energy Cut to 10MeV in run_simScript.py , this helps with studying veto efficiencies, but also avoids losing Tracks with vital info ( MCTrack[1].GetWeight() contains the DIS cross section of the incoming muon).

muonDIS/make_nTuple_Tr.py
olantwin: In the longer term, I think they should definitely be merged, unless they are very different.
In the shorter term, it would require updating the entire chain, so it would probably take quite a bit of work to transition, which I don't think anyone can invest right now.
Maybe we make an issue for it, so that we can get to it later, but for this PR combining both is definitely out of scope.
Maybe you could have a look whether any of the new code you add could be put into functions that can be reused in both, instead of duplicating.

macro/run_simScript.py
```diff
@@ -375,6 +375,9 @@
  ut.checkFileExists(inputFile)
  primGen.SetTarget(0., 0.)
  DISgen = ROOT.MuDISGenerator()
+ #  add Carbon to PDG
+ pdg = ROOT.TDatabasePDG.Instance()
+ pdg.AddParticle("C", "Carbon_Nucleus", 12.0, True, 0, 6.0, "nucleus", 1000060120)
```
olantwin: Maybe we would want to add this where we add the other particles to the database: https://github.com/ShipSoft/FairShip/blob/2eb3c2da286a64fefe6566362898da6b344fec27/python/shipRoot_conf.py#L63

macro/run_simScript.py
```diff
@@ -519,15 +519,20 @@
  sys.exit(0)
 gMC = ROOT.TVirtualMC.GetMC()
 fStack = gMC.GetStack()
+if options.mudis:
+  EnergyCut=10.*u.MeV
+else:
+  EnergyCut=100.*u.MeV
+
```
olantwin, line 526: Maybe better as a ternary statement to avoid possible issues of variable scope.

olantwin:
Are you adding any new TVectorD's compared to what was used before?

macro/run_simScript.py
anupama-reghunath: wasnt sure where to add this, maybe we should then also add line 445: `pdg.AddParticle('W','Ion', 1.71350e+02, True, 0., 74, 'XXX', 1000741840)`

olantwin:
> > Are you adding any new TVectorD's compared to what was used before?
> 
> ah yes, but this is metadata -> ( number of muons within the original MuonBack event that pass the threshold), didn't know how else to save this other than in muonDis.root , incase of recalculation of weights.

It's fine to store it in the file, but please use a `std::vector` instead, see #562.

anupama-reghunath:
> > > Are you adding any new TVectorD's compared to what was used before?
> > 
> > 
> > ah yes, but this is metadata -> ( number of muons within the original MuonBack event that pass the threshold), didn't know how else to save this other than in muonDis.root , incase of recalculation of weights.
> 
> It's fine to store it in the file, but please use a `std::vector` instead, see #562.

I have added this for now to avoid making new changes to the rest of the code implementation.To be specific, there is no new TVectorD but just an additional element in the existing TVectorD.

olantwin:
> > > > Are you adding any new TVectorD's compared to what was used before?
> > > 
> > > 
> > > ah yes, but this is metadata -> ( number of muons within the original MuonBack event that pass the threshold), didn't know how else to save this other than in muonDis.root , incase of recalculation of weights.
> > 
> > It's fine to store it in the file, but please use a `std::vector` instead, see #562.
> 
> To be specific, there is no new TVectorD but just an additional element in the existing TVectorD. I did so to avoid making new changes to the rest of the code implementation.

Ok, that's fine then. If we're just extending an existing object I can live with it.

olantwin:
From my side this looks good now. Probably now would be a good time to autosquash the fixups and fix the merge conflict by rebasing on master.

@maksymovchynnikov I'd really appreciate it, if you could help test and review this.

muonDIS/make_nTuple_Tr.py
anupama-reghunath: Issue opened #596

anupama-reghunath:
>     * Lowering the MCTrack Energy Cut to 10MeV in run_simScript.py , this helps with studying veto efficiencies, but also avoids losing Tracks with vital info ( MCTrack[1].GetWeight() contains the DIS cross section of the incoming muon).

With some testing, I noticed that in some cases the MCTrack on which the DIS cross section is available is not saved because it doesnt meet the threshold. We also increase the risk losing this info if we raise the MCTrack threshold to 100MeV ( since the files with 10MeV threshold are getting very large and if I'm not wrong kept only for veto related studies). @olantwin do you have any other suggestion on what to do? Perhaps within the EventHeader since it is metadata?

olantwin:
> > * Lowering the MCTrack Energy Cut to 10MeV in run_simScript.py , this helps with studying veto efficiencies, but also avoids losing Tracks with vital info ( MCTrack[1].GetWeight() contains the DIS cross section of the incoming muon).
> 
> With some testing, I noticed that in some cases the MCTrack on which the DIS cross section is available is not saved because the track Energy does not meet the threshold. We also increase the risk losing this info if we ever raise the threshold to 100MeV ( since the files with 10MeV threshold are getting very large and if I'm not wrong kept only for veto related studies). @olantwin do you have any other suggestion on what to do? Perhaps within the EventHeader since it is metadata?

Adding it to the event header or adding a scalar branch per event would probably be the best.
I've never been a fan of hiding this kind of information in track weights, which require special knowledge to decode.

olantwin:
> I don't see a suitable field within FairEventHeader, so maybe it makes sense to save this as a scalar branch until we have a custom event header. (I think however this is better as a separate PR).

Agreed. Then let's discuss this particular issue further in #634

olantwin (dismissed):
@anupama-reghunath I assume you already have run several test productions?

It would be good, maybe with the help of @maksymovchynnikov, to check whether the background numbers are compatible. This could maybe be done with a small production, if the statistical uncertainties are propagated properly.

shipgen/MuDISGenerator.h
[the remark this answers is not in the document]
olantwin: This should be fixed. These should probably just be constants in the implementation file.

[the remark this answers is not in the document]
anupama-reghunath: Pushed the changes now.

anupama-reghunath:
@olantwin yes I have done some simple tests productions so far and it looks okay, and then I will launch a larger production to compare the numbers from Maksym

olantwin (approved):

PR #511, opened 2024-09-12
fix typo (== instead of =)

lfpc:
If DecayVolumeMedium is not passed to ship_geo, code will crash due to typo (it should be = instead of ==)


olantwin (approved):
Thanks for catching this!

PR #513, opened 2024-09-12
Remove preshower and CaloDesign 1

olantwin:
@matclim


olantwin:
If there are no objections, this PR will be merged on Nov 4 after the software meeting. Silence will be interpreted as approval.

matclim:
Everything seems to be in order, I think you should be able to merge.

matclim (approved):

PR #532, opened 2024-10-09
shipStrawTracking: Move to argparse

olantwin:


fairlyons (changes requested):
Works the same as before, "dummy" file is still created.

olantwin:
> Works the same as before, "dummy" file is still created.

Changing the output behaviour (and fixing the write to the input file) is a matter for another PR.
A lot of cleanup needs to be done to the file.

olantwin:
So if everything works as before, this PR works as intended.

fairlyons (approved):
I have tested and the functionality is unchanged.

PR #533, opened 2024-10-10
Changes for standalone genfit

olantwin:


fairlyons:
Do charmdet and preshower also need to be removed from CMakeLists? I could not build with them

olantwin:
Indeed. I'll update this PR to include those changes. I'll probably then open a separate PR to remove the other charmdet stuff.
It would be neater to remove charmdet and preshower first, but it will probably require a lot of prodding to make that happen on a reasonable timescale...

olantwin:
Should be fixed now.

@fairlyons : If you could test that the tracking:
- [ ] runs
- [ ] has performance consistent with the integrated genfit

we should be ready to merge and finally tick off at least this todo item.

fairlyons:
Current code builds successfully with the correct shipdist but fails with many errors on reconstruction. Several genfit files cannot be located.

antonioiuliano2 (dismissed):
Ok for me, thank you!

olantwin:
> Ok for me, thank you!

Thanks for your review! 
This will need some fairly extensive testing (@fairlyons is on it), before it's ready though. It seems the new version isn't quite the drop-in replacement I had hoped...

fairlyons:
It seems that now none of the tracks are properly fitted, it may take some time to do a thorough investigation.

olantwin:
With @fairlyons this is now ready to be merged. Will fix merge conflicts and publish a new stack.

@antonioiuliano2, could you please start building a new cvmfs stack for Ubuntu for the 24.10 tag?

macro/ShipReco.py
```diff
@@ -77,6 +77,10 @@ def mem_monitor():
  options.geoFile = tmp.replace('_rec','')
 
 fgeo = ROOT.TFile.Open(options.geoFile)
+ROOT.gSystem.Load("libgenfit2")
+ROOT.gInterpreter.Declare('#include "TGeoMaterialInterface.h"')
+ROOT.gInterpreter.Declare('#include "FieldManager.h"')
+ROOT.gInterpreter.Declare('#include "MaterialEffects.h"')
```
olantwin, line 83: Can probably be removed now that the autoloading works correctly.

fairlyons: Yes this can be removed.

olantwin:
Tracking and vertexing performance now exactly the same as before, thanks to fixes by @fairlyons

fairlyons (approved):
Performance confirmed to be the same as before for both tracking and vertexing.

wcmartylee (approved):

PR #546, opened 2024-10-24
Remove conditions DB

olantwin:
If there are no objections, this PR will be merged on Nov 4 after the software meeting. Silence will be interpreted as approval.


olantwin:
As there were no objections, merging.

PR #551, opened 2024-11-04
Fix division by 0 error

fairlyons:


olantwin (approved):
A few things we should follow up on:

- [ ] Right now we only skip filling the histogram. Should we maybe discard the track entirely, if `nmeas == 0`?
- [ ] the chi2 is reduced chi2, if I am not mistaken, we should probably make sure the histogram is labelled correctly

CC @evanherwijnen

PR #556, opened 2024-11-10
millepede removed from CMakeLists.txt in the main folder

eduard322:
Hi,
I removed the ```add_subdirectory (millepede)``` from ./CMakeLists.cxx because it caused the errors during the build of Fairship due to the last commit


olantwin (approved):
My bad, sorry. I was shuffling around different commits and the change to the CMakeLists.txt was lost somewhere when resolving conflicts.

PR #565, opened 2024-11-22
Add CERN-hosted runner workflow for building and running execs

egamberini:
Add a workflow triggering on:
* pull requests to master
* push to master and to v* tags
* manually

The workflow runs on a CERN-hosted runner with access to `/cvmfs`.

Container image is at [registry.cern.ch/ship/gha-runner:latest](url).

The triggering branch is used to build and run the sim/reco/ana pipeline.
Produced `.root` file are uploaded on success as artifacts and can be downloaded and inspected manually.


olantwin (approved):
Thanks a lot, this will make our life a lot easier in the medium term :)

PR #570, opened 2024-11-27
fixed bug of unit conversion in neuDIS Generator

anupama-reghunath:


shipgen/GenieGenerator.cxx
```diff
@@ -548,7 +548,7 @@ Bool_t GenieGenerator::ReadEvent(FairPrimaryGenerator* cpg)
     //cout << "Info GenieGenerator: prob2int " << prob2int << ", " << count << endl;
 
     Double_t zrelative=z-ztarget;
-    Double_t tof=TMath::Sqrt(x*x+y*y+zrelative*zrelative)/2.99792458e+6;
+    Double_t tof=TMath::Sqrt(x*x+y*y+zrelative*zrelative)/2.99792458e+10;
```
olantwin, line 551: Could you please auto-format this line and either put the units in a comment, or better, use the unit constants so that there can be no confusion about the units in future?

olantwin (changes requested):
Thanks a lot, this is a very important fix.

olantwin (approved):

PR #571, opened 2024-11-28
correct default settings within geometry_config.py

anupama-reghunath:
Since within `run_simScript.py` we always pass the default value, the default settings within `geometry_config.py` were never used.

### DIFF
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -59,7 +59,7 @@ if "SC_mag" not in globals():
 if "shieldName" not in globals():
     shieldName = None
 if "DecayVolumeMedium" not in globals():
-    DecayVolumeMedium = "helium"
+    raise ValueError("DecayVolumeMedium is not defined, possibly old (incompatible) geometry!")
 if "SND" not in globals():
     SND = False
 
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -58,8 +58,6 @@ if "SC_mag" not in globals():
     SC_mag = False
 if "shieldName" not in globals():
     shieldName = None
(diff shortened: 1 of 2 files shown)


olantwin:
These defaults are for backwards compatibility, to allow e.g. reconstructing old files before helium and removal of the SND were options, so I don't think these need updating. 
Instead, we might consider throwing an error when trying to reconstruct these old files (and maybe give information this commit it was generated with, so that it can be reconstructed with the correct version).

anupama-reghunath:
atleast within the context of DecayVolume, the current version is not backwards compatible. 'vacuums' as default uses the current geometry for vacuum as opposed any  of the older geometries with 'vacuums'  as decayVolumeMedium.

olantwin:
Then at least for the DecayVolume we should probably just raise an error (with a helpful error message), instead of constructing an incompatible geometry.

olantwin (changes requested):
Since we can still construct the SND, its default value for old geometries (which had an SND) should be True. @antonioiuliano2 what do you think?

Looks good to me otherwise.

olantwin:
@anupama-reghunath @antonioiuliano2 How shall we proceed with this PR?

olantwin:
Any updates? It would be nice to get this merged.

olantwin:
Closing because of lack of response.

olantwin (approved):

olantwin:
@anupama-reghunath Could you add a CHANGELOG entry for this? Dropping support for geometries without DecayVolumeMedium is worth making explicit there. The backward compatibility didn't actually exist, as even the vacuums case has changed relative to geometries without that variable.

PR #575, opened 2024-12-05
SST boxes medium will now follow DecayVolumeMedium

olantwin:
@evanherwijnen noticed that the boxes containing the stations were still `vaccums`. This change should now make the medium follow the one used for the decay volume.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ it in future.
 ### Changed
 
 * Set Decay Volume Medium as helium (previously vacuums),can be explicitly switched to vacuum with --vacuums.
+* Medium of SST boxes will be the same as DecayVolumeMedium (previously, always vacuum)
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


wcmartylee:
> @evanherwijnen noticed that the boxes containing the stations were still `vaccums`. This change should now make the medium follow the one used for the decay volume.

Hi Oliver, thanks for this find! On a second thought, the stations are actually in air with the helium option.

olantwin:
Ok, so then is should be vacuum for the vacuum option and air otherwise?

wcmartylee:
> Ok, so then is should be vacuum for the vacuum option and air otherwise?

Yes.

olantwin:
Ok, now it should be the desired behaviour. Please test.

evanherwijnen:
Yes

On 5 Dec 2024, at 10:35, Oliver Lantwin ***@***.***> wrote:

﻿

Ok, so then is should be vacuum for the vacuum option and air otherwise?

wcmartylee:
Only sometimes these errors would appear by running Reco:
`shipDigiReco::findVetoHitOnTrack extrapolation did not worked`
`shipVertex: extrapolation did not work`

Might not be due to this change, but I’m not sure.

olantwin:
That is unrelated, or at least it already occurred occasionally before.
Then could you please approve?

wcmartylee (approved):

PR #578, opened 2024-12-11
Reco: Don't prune tracks

olantwin:
Before we were pruning with the CFL option. See [here](https://github.com/GenFit/GenFit/blob/e81adeb07c8643301a1d9f7ae25048557cc72dff/core/include/Track.h#L298) for the Genfit documentation.

We might want to reintroduce the pruning in future, but right now the small saving in on-disk size doesn't seem to be worth the loss in information.


evanherwijnen:
Hi Oliver
This works! Cheers and thanks Eric

olantwin:
Thanks Eric for the quick feedback and for spotting the original issue!

PR #597, opened 2025-02-07
Build FairShip with the separated libTPythia6.so library

antonioiuliano2:
Dear all,

This pull request is related to the update for ROOT 6.32 without TPythia6 support: 
https://github.com/ShipSoft/shipdist/pull/102.

Basically, this pull request will tell FairShip to load the TPythia6 from the shared library, now separated from ROOT.

It is used by

* TPythia6Generator (compiled in shipgen), and run_simScript.py which calls it
* makeCascade.py
* makeMuonDIS.py

Best Regards,
Antonio


shipgen/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ ${CMAKE_SOURCE_DIR}/generators
 ${CMAKE_SOURCE_DIR}/shipdata
 ${CMAKE_SOURCE_DIR}/veto
 ${genfit2_INCDIR}
+${TPYTHIA6_INCLUDE_DIR}
```
olantwin, line 11: Can we define a CMake Target as a dependency instead of manually including the libraries?

macro/makeCascade.py
```diff
@@ -74,7 +74,11 @@
    sys.exit('ERROR on input, exit')
 
 PDG = ROOT.TDatabasePDG.Instance()
-myPythia = ROOT.TPythia6()
+try:
+   myPythia = ROOT.TPythia6()
+except: #library not part of ROOT from 6.32 onward, need to load it externally
```
olantwin, line 79: Can we narrow this `except` to a single Exception type?

macro/makeCascade.py
```diff
@@ -287,7 +291,7 @@ def fillp1(hist):
          if stack[nstack][0]==idhist[i]:
             idpn=0
 #   decide on p or n target in Mo
-            if random.random>fracp: idpn=1
+            if random.random()>fracp: idpn=1
```
olantwin, line 294: Too many statements on one line, missing spaces around operators.

python/basiclibs.py
```diff
@@ -6,7 +6,7 @@
 #gSystem.Load("libGeomPainter.so")
 #gSystem.Load("libVMC.so")
 #gSystem.Load("libEG.so")
-gSystem.Load("libEGPythia6.so")
+#gSystem.Load("libEGPythia6.so")
```
olantwin, line 9: Delete instead of commenting.

gconfig/basiclibs.C
```diff
@@ -12,14 +12,14 @@ void basiclibs()
   gSystem->Load("libGeomPainter");
   gSystem->Load("libVMC");
   gSystem->Load("libEG");
-  gSystem->Load("libEGPythia6");
+  //gSystem->Load("libEGPythia6");
```
olantwin, line 15: Delete instead of commenting

gconfig/basiclibs.C
```diff
@@ -12,14 +12,14 @@ void basiclibs()
...
   gSystem->Load("libPythia6");
   gSystem->Load("libPhysics");
   gSystem->Load("libNet");
   gSystem->Load("libTree");
   gSystem->Load("libMinuit");
   gSystem->Load("libMathMore");
   gSystem->Load("libpythia8");
-  gSystem->Load("libgenfit.so");
+  //gSystem->Load("libgenfit.so");
```
olantwin, line 23: Remove this unrelated change.

gconfig/DecayConfigPy6.C
```diff
@@ -5,6 +5,9 @@ void DecayConfig() {
 	// specific types of decays defined below.
 
 	// Access the external decayer singleton and initialize it
+	if (TString(gSystem->DynamicPathName("libtpythia6.so", kTRUE)) != TString("")){
```
olantwin, line 8: I don't think this is needed, if the ROOT dictionaries are in the correct place.

macro/run_simScript.py
```diff
@@ -345,7 +345,11 @@
 # set muon interaction close to decay volume
  primGen.SetTarget(ship_geo.target.z0+ship_geo.muShield.length, 0.)
 # -----Pythia6-------------------------
- test = ROOT.TPythia6() # don't know any other way of forcing to load lib
+ try:
+   test = ROOT.TPythia6() # don't know any other way of forcing to load lib
+ except: #library not part of ROOT from 6.32 onward, need to load it externally
+   ROOT.gSystem.Load("libtpythia6.so")
```
olantwin, line 351: Probably not needed, if dictionaries are in the right location.

olantwin (changes requested):
We should also use find_package to specify TPythia6 as a dependency and use it to define include paths etc.

olantwin:
Builds fine on SLC9

olantwin:
@antonioiuliano2 Any updates on this?

olantwin:
It would be good to add a changelog entry. While nothing user-facing should change, this is quite a major under-the-hood change.

olantwin:
One thing we could try for backwards compatibility is to check whether ROOT has TPythia6, and if not search for the TPythia6 standalone package using find_package.
Right now you kind of do that implicitly by using only an additional include path, which might be blank. Unfortunately the lack of TPythia6 might only be noticed at runtime as a result.

gconfig/DecayConfigPy6.C
antonioiuliano2: You are right, 

I believe the issue with this and the similar comment is that I should use the same library name as the ROOT old TPythia6 library. libEGPythia6.so, instead of libtpythia6.so. 

I will do it and remove all these redundant if conditions.

olantwin: Indeed, using the same library name seems like a good idea.

Thanks for the update!

macro/run_simScript.py
```diff
@@ -345,11 +346,7 @@
 # set muon interaction close to decay volume
  primGen.SetTarget(ship_geo.target.z0+ship_geo.muShield.length, 0.)
 # -----Pythia6-------------------------
- try:
-   test = ROOT.TPythia6() # don't know any other way of forcing to load lib
- except: #library not part of ROOT from 6.32 onward, need to load it externally
-   ROOT.gSystem.Load("libtpythia6.so")
-   test = ROOT.TPythia6()
+ test = ROOT.TPythia6() # don't know any other way of forcing to load lib
```
olantwin, line 349: If the libraries are in the correct place, this should not be necessary.

muonShieldOptimization/makeMuonDIS.py
```diff
@@ -1,4 +1,8 @@
 import ROOT,time,os,sys
+
+ROOT.gROOT.LoadMacro("$VMCWORKDIR/gconfig/basiclibs.C")
```
olantwin, line 3: Can we make this a header and load it as such?
I'd really like to limit our use to macros as much as possible.

antonioiuliano2: For now (and to minimize errors on my changes), I have followed the procedure already in place for makeCascade.py 

But I believe it should work as you suggested

olantwin: Ok, then it's fine for now.

olantwin:
I think once we confirm that this build fine with:
- [ ] ROOT 6.30 with builtin TPythia6
- [ ] ROOT 6.32 with standalone TPythia6

And does not build with:
- [ ] ROOT 6.32 without standalone TPythia6

We can probably merge this first version.

A changelog entry would still be appreciated ;)

olantwin:
@antonioiuliano2 Have you tried generating some muon DIS events with this branch?

antonioiuliano2:
> @antonioiuliano2 Have you tried generating some muon DIS events with this branch?

Yes, I did a test with the muonShieldOptimization/makeMuonDIS.py  script and the default muConcrete.root file. Nothing out of the order to report.



> It would be good to add a changelog entry. While nothing user-facing should change, this is quite a major under-the-hood change.

Added an entry to the CHANGELOG.

olantwin:
Ok, could you please rebase?

macro/run_simScript.py
```diff
@@ -2,6 +2,7 @@
 import os
 import sys
 import ROOT
+ROOT.gSystem.Load('libEGPythia6')
```
olantwin, line 5: Where is pythia6 used in run_simScript.py?

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ it in future.
 * chore: trim trailing whitespace
 * Fix mismatch dimension cavern ECN3 TCC8
 * fix: use standard library to expand env
+* Compatibility with ROOT 6.32, including an external TPythia6 package (not supported anymore by ROOT).
```
olantwin, line 23: Please move this to the *Changed* heading.

olantwin (changes requested):

shipgen/CMakeLists.txt
olantwin: This can be addressed later.

macro/run_simScript.py
olantwin: Auto-load should work fine, see #633

olantwin: If you didn't explicitly check that this is needed, I would remove this line until someone shows that it is needed.

olantwin (approved):

PR #592, opened 2025-01-27
Review of the dimension of the cave

Gfrisella:
Review of the dimension of the cave: 1) Uptated dimensions of TCC8 and ECN3. 2) Added a Step in ECN3. 3) Removed the coat of the absorber. 4) Modified the dimension of the HA and absorber.
Draft

Link to the presentation: [https://docs.google.com/presentation/d/1nRv-70tWRiKnH0pJtc1-iokxBEzhVT5Ow-1VYbAtBOk/edit#slide=id.g2d84c95faf5_2_0](url)

PR #598, opened 2025-02-07
Switch Shield

eduard322:
Here is the stable version in which the usage of ROOT file for setting the dinensions of Muon Shield is removed (at least now it takes info directly from the geometry_config.py script). Several things to mention and to discuss:

1. During the simulation the geometry registration is conducted twice for some reason, which seems very redundant. First time in run_simScript.py and then in shipDet_conf.py. If there is no hidden reason, I can remove these entities.
2. Reading the Muon Shield dimensions from the dictionary in geometry_config.py script doesn't look perfect since the geometry of Muon Shield is not fixed yet. During optimization me and Evgeny were using json files for storing different geometries, would be nice to repeat this implementation now. Or maybe use yaml file since the info about decay volume is stored in yaml files and it would be more consistent.
3. Current warm version overlaps with the floor. Floor can be excavated though, we need a permission to make so, don't we?

Screenshot 2025-02-07 at 16 29 23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ it in future.
 ### Changed
 
 * Don't special case EOS paths (fix #566)
+* Setting up the Muon shield geometry by ROOT files is completely replaced with the temporary solution of dict in the ```geometry/geometry_config.py```. Set up the shield name is now done by --shieldName key instead of --scName.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ it in future.
 ### Changed
 
 * Don't special case EOS paths (fix #566)
-* Setting up the Muon shield geometry by ROOT files is completely replaced with the temporary solution of dict in the ```geometry/geometry_config.py```. Set up the shield name is now done by --shieldName key instead of --scName.
+* Setting up the Muon shield geometry by ROOT files is completely replaced with the temporary solution of dict in the `geometry/geometry_config.py`.
+* Set up of the shield name is now done using the `--shieldName` flag instead of `--scName`.
 
 ### Removed
 
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -56,8 +56,8 @@ if "muShieldWithCobaltMagnet" not in globals():
     muShieldWithCobaltMagnet = 0
 if "SC_mag" not in globals():
     SC_mag = False
-if "scName" not in globals():
-    scName = None
+if "shieldName" not in globals():
+    shieldName = None
 if "DecayVolumeMedium" not in globals():
     DecayVolumeMedium = "vacuums"
 if "SND" not in globals():
@@ -69,7 +69,7 @@ with ConfigRegistry.register_config("basic") as c:
     c.SND = SND
 
     c.SC_mag = SC_mag
-    c.scName = scName
+    c.shieldName = shieldName
     # global muShieldDesign, targetOpt, strawDesign, Yheight
     c.Yheight = Yheight*u.m
     # decision by the SP
@@ -315,16 +315,19 @@ with ConfigRegistry.register_config("basic") as c:
         c.muShield.half_Y_max = 317 * u.cm
     elif muShieldDesign == 8:
         if not SC_mag:
-            assert muShieldGeo
-            c.muShieldGeo = muShieldGeo
-            print("Load geo")
-            f = r.TFile.Open(muShieldGeo)
-            params = r.TVectorD()
-            params.Read('params')
-            f.Close()
+            # assert muShieldGeo
+            # c.muShieldGeo = muShieldGeo
+            # print("Load geo")
+            # f = r.TFile.Open(muShieldGeo)
+            # params = r.TVectorD()
+            # params.Read('params')
+            # f.Close()
+            assert shieldName
+            params = shield_db[shieldName]
+            c.muShield.params = params
         else:
-            assert scName
-            params = shield_db[scName]
+            assert shieldName
+            params = shield_db[shieldName]
             c.muShield.params = params
         c.muShield.dZ1 = 0.35*u.m + zGap
         c.muShield.dZ2 = 2.26*u.m + zGap
--- a/geometry/geometry_config.py
+++ b/geometry/geometry_config.py
@@ -314,21 +314,9 @@ with ConfigRegistry.register_config("basic") as c:
         c.muShield.half_X_max = 179 * u.cm
         c.muShield.half_Y_max = 317 * u.cm
     elif muShieldDesign == 8:
-        if not SC_mag:
-            # assert muShieldGeo
-            # c.muShieldGeo = muShieldGeo
-            # print("Load geo")
-            # f = r.TFile.Open(muShieldGeo)
-            # params = r.TVectorD()
-            # params.Read('params')
-            # f.Close()
-            assert shieldName
-            params = shield_db[shieldName]
-            c.muShield.params = params
-        else:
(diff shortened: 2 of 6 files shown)


geometry/geometry_config.py
```diff
@@ -315,16 +315,19 @@
...
     elif muShieldDesign == 8:
         if not SC_mag:
-            assert muShieldGeo
-            c.muShieldGeo = muShieldGeo
-            print("Load geo")
-            f = r.TFile.Open(muShieldGeo)
-            params = r.TVectorD()
-            params.Read('params')
-            f.Close()
+            # assert muShieldGeo
```
olantwin, line 318: It seems like the conditional is no longer needed with your changes.

macro/run_simScript.py
```diff
@@ -201,8 +201,8 @@
  print("cannot have HNL and SUSY or DP at the same time, abort")
  sys.exit(2)
 
-if (simEngine == "Genie" or simEngine == "nuRadiography") and defaultInputFile:
-  inputFile = "/eos/experiment/ship/data/GenieEvents/genie-nu_mu.root"
+# if (simEngine == "Genie" or simEngine == "nuRadiography") and defaultInputFile:
+#   inputFile = "/eos/experiment/ship/data/GenieEvents/genie-nu_mu.root"
```
olantwin, line 205: This seems to be unrelated to the muon shield.
If you want to propose changes to the neutrino simulation, please make a separate pull request for that.

passive/ShipMuonShield.cxx
```diff
@@ -513,15 +513,16 @@ Int_t ShipMuonShield::Initialize(std::vector<TString> &magnetName,
...
+    params = shield_params;
+    // if (!fSC_mag)
+    // {
+    //   TVectorT<Double_t> paramsR;
+    //   auto f = TFile::Open(fGeofile, "read");
+    //   paramsR.Read("params");
+    //   for (int i = 0; i < 56; i++){params.push_back(paramsR[i]);}
+    // }else{
+    //   params = shield_params;
+    // }
```
olantwin, line 525: Don't comment, remove.

passive/ShipMuonShield.cxx
```diff
@@ -969,11 +970,11 @@ void ShipMuonShield::ConstructGeometry()
         continue;
       }
   Double_t ironField_s_SC = fField * fieldScale[nM] * tesla;
-  Bool_t SC_key = false;
+  // Bool_t SC_key = false;
```
olantwin, line 973: Remove

passive/ShipMuonShield.cxx
```diff
@@ -969,11 +970,11 @@ void ShipMuonShield::ConstructGeometry()
...
   if(nM == 3 && fSC_mag){
     Double_t SC_FIELD = 5.1;
     ironField_s_SC = SC_FIELD * fieldScale[nM] * tesla;
-    SC_key = true;
+    // SC_key = true;
```
olantwin, line 977: remove

olantwin (commented):
I'll need to do some testing, but some first comments that stood out while reading.

olantwin:
>     1. During the simulation the geometry registration is conducted twice for some reason, which seems very redundant. First time in run_simScript.py and then in shipDet_conf.py. If there is no hidden reason, I can remove these entities.

What do you mean by registration?
 
>     2. Reading the Muon Shield dimensions from the dictionary in geometry_config.py script doesn't look perfect since the geometry of Muon Shield is not fixed yet. During optimization me and Evgeny were using json files for storing different geometries, would be nice to repeat this implementation now. Or maybe use yaml file since the info about decay volume is stored in yaml files and it would be more consistent.

I'm not sure what our UZH colleagues are doing now. A yaml would probably be nicest, as it allows us to add comments etc. and is both flexible and readable. Ideally, a format we could share with the SBT/vessel/other subsystems would be nice.

>     3. Current warm version overlaps with the floor. Floor can be excavated though, we need a permission to make so, don't we?

I believe the muon shield optimisation now includes constraints to avoid overlaps with the floor. Unfortunately, none are implemented yet. @Gfrisella, maybe you can provide the dimensions of a configuration we can use via the dictionary/yaml?

eduard322:
The geometry is registered here https://github.com/eduard322/FairShip/blob/68fa73cede0025222016b4a05fd92f68a07f53cc/macro/run_simScript.py#L224
Then ship_geo goes to shipDet_conf.py script where another geometry registration is conducted https://github.com/eduard322/FairShip/blob/68fa73cede0025222016b4a05fd92f68a07f53cc/python/shipDet_conf.py#L151
with literally same parameters coming from the previous registration in run_simScript.py
Looks ambiguous

olantwin:
> The geometry is registered here https://github.com/eduard322/FairShip/blob/68fa73cede0025222016b4a05fd92f68a07f53cc/macro/run_simScript.py#L224 Then ship_geo goes to shipDet_conf.py script where another geometry registration is conducted https://github.com/eduard322/FairShip/blob/68fa73cede0025222016b4a05fd92f68a07f53cc/python/shipDet_conf.py#L151 with literally same parameters coming from the previous registration in run_simScript.py Looks ambiguous

Indeed, in the case of run_simScript.py it seems completely redundant, but in the case of loading old data for the reconstruction, it has its use for forward-compatibility.

Maybe we can find a way to avoid the unnecessary step in run_simScript.py.

macro/run_simScript.py
```diff
@@ -203,7 +203,7 @@
 
 if (simEngine == "Genie" or simEngine == "nuRadiography") and defaultInputFile:
   inputFile = "/eos/experiment/ship/data/GenieEvents/genie-nu_mu.root"
-            # "/eos/experiment/ship/data/GenieEvents/genie-nu_mu_bar.root"
+            "/eos/experiment/ship/data/GenieEvents/genie-nu_mu_bar.root"
```
olantwin, line 206: This line causes a runtime error due to an unexpected indent.

eduard322: sorry, extra info..

olantwin (dismissed):
From my side this looks OK.

Please add a change log entry describing your change.

Since this affects the muon shield, I will wait for an OK from @Gfrisella on behalf of the muon shield before merging.

olantwin:
pre-commit.ci run

olantwin:
> @olantwin sorry, stupid question: how is change log usually done? Should I add CHANGELOG.md to the commit or something?

It's just a text file. Add a succinct entry under *unreleased* describing the change, and commit the change to the pull request branch.

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ it in future.
 ### Added
 
 ### Changed
-
+* Setting up the Muon shield geometry by ROOT files is completely replaced with the temporary solution of dict in the ```geometry/geometry_config.py```. Set up the shield name is now done by --shieldName key instead of --scName.
```
olantwin, line 22: I think I would make an entry for each change (i.e. make a new bullet point after the first sentence.

For readability, I would also use single backticks for the filepath and add a blank line below the heading and after the list.

olantwin (changes requested):

olantwin:
The overlaps seem to be only the ones from #592 + #609 , no new overlaps are introduced :+1:

olantwin:
@Gfrisella Could you please review this pull request?

olantwin (approved):
Did some cleanup of the changelog and ran clang-format on the changed lines.

Looks good to me.

olantwin:
@Gfrisella The muon shield group is welcome to suggest improvements. For now this seems like a clear step in the right direction.

@eduard322 Thanks to making a first step towards improving the muon shield implementation.

PR #602, opened 2025-02-11
fix: Don't special case EOS paths

### LINKED ISSUE #566: Remove manual redirection to XROOTD for files on EOS
Since ROOT can do this automatically as of [6.28](https://root.cern/doc/v628/release-notes.html#faster-reading-from-eos), we should remove our hacky way of doing it ourselves.

olantwin:
Fixes #566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ it in future.
 
 ### Changed
 
+* Don't special case EOS paths (fix #566)
+
 ### Removed
 
 ## 25.01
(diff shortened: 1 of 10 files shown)


fairlyons (approved):
I tried the normal test on lxplus: runsimscript -> shipreco -> shipana and had no problems.

olantwin:
> I tried the normal test on lxplus: runsimscript -> shipreco -> shipana and had no problems.

Thanks for the quick feedback, it's highly appreciated!

olantwin:
Copying my comment from mattermost here for posterity:

> This has the potential to be breaking for users who depend on the special casing on systems that don't have EOS mounted via FUSE. If you specify the full paths or work on lxplus where the FUSE mount exist, nothing should change.
> This will be merged in a few days, if nobody makes any objections known to me.

Posted Monday, so probably "in a few days" = Friday or Monday.

Issue #599, opened 2025-02-10
Overlaps of cavern with muon shield and tracking stations, gaps in cavern


olantwin:
Introduced by #592.

```
=== Overlaps for FAIRGeom ===
 = Overlap ov00000: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MiddleMagL_1 ovlp=20.4398
 = Overlap ov00001: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MiddleMagR_1 ovlp=20.4398
 = Overlap ov00002: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MagRetL_1 ovlp=20.4398
 = Overlap ov00003: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MagRetR_1 ovlp=20.4398
 = Overlap ov00004: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MagTopLeft_1 ovlp=20.4398
 = Overlap ov00005: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MagTopRight_1 ovlp=20.4398
 = Overlap ov00006: cave/Cavern_1 overlapping cave/MuonShieldArea_1/Magn4_MagBotLeft_1 ovlp=20.4398
 = Overlap ov00007: cave/Cavern_1 overlapping cave/Tr4_4 ovlp=8.28469
 = Overlap ov00008: cave/Cavern_1 overlapping cave/Tr1_1 ovlp=5.89438
 = Overlap ov00009: cave/Cavern_1 overlapping cave/Tr2_2 ovlp=5.89438
```

This is also quite obvious visually:

![Image](https://github.com/user-attachments/assets/a3dc81cc-6b7b-4704-b917-fc7878a6dbc8)

![Image](https://github.com/user-attachments/assets/8700442f-2be9-49c6-b667-55f202512935)

olantwin:
@wcmartylee Please also check with the SST group, what the plan is for the straw trackers. Not sure whether the trench needs to move or changes to the tracker are needed.

olantwin:
Overlap with muon shield will be fixed by #609

olantwin:
The change to the cavern was premature and will be reverted, so this issue no longer applies.

olantwin:
Reopening as the decision on the revert will be taken Monday.

PR #606, opened 2025-02-11
Add overlap check to CI

olantwin:


olantwin:
Great, this works as expected, see the CI failure due to the overlaps from #599

egamberini:
oops too late

olantwin:
> Hello! The overlap check could be moved before the Sim step to fail fast.

Not sure, the overlap check takes quite some time, so for a quick sanity check, it might be better to first run the sim. We could even run it last or in parallel, as it might still be worth checking that nothing else breaks.

PR #609, opened 2025-02-11
Fix Cavern Mismatch

Gfrisella:
Fixed the cavern mismatch TCC8 and ECN3.
image(1)


olantwin (changes requested):
Good, this also fixes the overlaps with the muon shield.

Could you please move the changelog entry to the *Unreleased* heading?

PR #611, opened 2025-02-12
Use git-lfs to track ROOT files

olantwin:


olantwin:
@egamberini It seems like it causes issues for the CI if we use git-lfs to store ROOT files. Is this something we need to fix on the runners?

egamberini:
Weird, I can't reproduce the error locally while using the `gha-runner` container, so it doesn't seem to be a missing dependency in the container itself.
The runner VM and its OS shouldn't affect what happens inside the container, but I'll investigate further tomorrow....

olantwin:
> Weird, I can't reproduce the error locally while using the `gha-runner` container, so it doesn't seem to be a missing dependency in the container itself.
> The runner VM and its OS shouldn't affect what happens inside the container, but I'll investigate further tomorrow....

Thanks for the quick reply!

olantwin:
@egamberini Did you have a chance to investigate why the runner fails?

olantwin:
Alternatively, we might want to just move these files to EOS.

egamberini:
I've managed to reproduce the error on different environments:
- container in runner VM
- container on local computer
- local computer

Only works on lxplus in my tests.

Therefore there must be a missing dep (maybe a python package or pyrooot extension that's present on lxplus system-wide and needs to be explicitly installed on other environments).
I've tried installing `git-lfs` as a system package but doesn't help.

Executing `pip list | grep lfs` on lxplus doesn't match anything.

I'm at a loss of ideas.

olantwin:
@egamberini I think we're a step further, now it complains about the git-lfs executable missing.

egamberini:
Nice! I've pushed a new container with `git-lfs` installed. That fixed the CI seems like...

olantwin:
> Nice! I've pushed a new container with `git-lfs` installed. That fixed the CI seems like...

Thanks a lot!

PR #614, opened 2025-02-12
Revert premature geometry update

olantwin:
This reverts commit 20f685cf864fb62c2a59b44e74bd721f6544f1ed and commit b4d71c3eee59c5b4737a9000c95a4bcbe93c4926.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,10 @@ it in future.
 
 * chore: Fix file endings
 * chore: trim trailing whitespace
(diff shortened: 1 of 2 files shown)


olantwin:
CC @Gfrisella , @eduard322 

I suggest we discuss Monday how to proceed.

olantwin:
It was decided to not revert, update to final version by the collaboration meeting.

After that, we'll try to stay in sync with versioned layouts, ideally in such a way that we can support multiple layouts.

PR #622, opened 2025-02-13
CI use global version env

egamberini:
Use a global "version" env variable to avoid mismatches.

Bump to 25.01 in master.


olantwin (changes requested):
@egamberini There seems to be some trailing whitespace hiding somewhere. (See the pre-commit CI).

Apart from that, this looks good to me!

olantwin:
Fixed.

PR #630, opened 2025-02-20
Shorten frame width to avoid overlap

wcmartylee:
Shorten the width of SST frames temporarily to avoid overlaps with the cave


olantwin (approved):

olantwin:
Great, thanks a lot!

PR #640, opened 2025-03-05
build(main): Rename *LinkDef.h to LinkDef.h

olantwin:
As there is only one LinkDef.h file in the directories for each component, use a common name everywhere.


olantwin:
Seems like it's not as simple as it seemed, there seems to be interference between the different linkdefs.

PR #650, opened 2025-03-10
fix: Don't use TFile's deprecated attribute syntax

### LINKED ISSUE #644: The attribute syntax for TFile is deprecated and will be removed in ROOT 6.34
E.g.:
```
DeprecationWarning: The attribute syntax for TFile is deprecated and will be removed in ROOT 6.34. Please use TFile["FAIRGeom"] instead of TFile.FAIRGeom
  fgeo.FAIRGeom
```

olantwin:
Use the item-getting syntax instead.
Fixed for FAIRGeom and cbmsim. Other uses might have escaped notice. Fixes #644.


olantwin:
Of course the new syntax only works from 6.32 onward...

PR #657, opened 2025-03-12
Add scalar branch to save CrossSection for muonDIS

### LINKED ISSUE #634: Add MuDIS cross-section as scalar branch to data format
I don't see a suitable field within FairEventHeader, so maybe it makes sense to save this as a scalar branch until we have a custom event header. (I think however this is better as a separate PR).

_Originally posted by @anupama-reghunath in https://github.com/ShipSoft/FairShip/issues/510#issuecomment-2690829828_

anupama-reghunath:
Adds a custom CrossSection branch to save the DIS cross sections from the muonDIS file produced using #510. 
Resolves #634

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -652,34 +652,31 @@ if simEngine == "MuonBack":
  fin.SetWritable(False) # bpyass flush error
 
 if simEngine == "muonDIS":
-    f_outputfile = ROOT.TFile.Open(outFile, "read")
-    output_tree = f_outputfile.cbmsim
 
-    f_muonfile = ROOT.TFile.Open(inputFile, "read")
-    muondis_tree = f_muonfile.Get("DIS")
+    temp_filename = outFile.replace(".root", "_tmp.root")
 
-    iMuon = ROOT.TClonesArray("TVectorD")
-    muondis_tree.SetBranchAddress("InMuon", iMuon)
+    with (
+        ROOT.TFile.Open(outFile, "read") as f_outputfile,
+        ROOT.TFile.Open(inputFile, "read") as f_muonfile,
+        ROOT.TFile.Open(temp_filename, "recreate") as f_temp,
+    ):
+        output_tree = f_outputfile.cbmsim
 
-    temp_filename = outFile.replace(".root", ".tmp")
-    f_temp = ROOT.TFile.Open(temp_filename, "recreate")
+        muondis_tree = f_muonfile.DIS
 
-    new_tree = output_tree.CloneTree(0)
+        new_tree = output_tree.CloneTree(0)
 
-    cross_section = array("f", [0.0])
-    cross_section_leaf = new_tree.Branch(
-        "CrossSection", cross_section, "CrossSection/F"
-    )
+        cross_section = array("f", [0.0])
+        cross_section_leaf = new_tree.Branch(
+            "CrossSection", cross_section, "CrossSection/F"
+        )
 
-    for output_event, muondis_event in zip(output_tree, muondis_tree):
-        mu = muondis_event.InMuon[0]
-        cross_section[0] = mu[10]
-        new_tree.Fill()
+        for output_event, muondis_event in zip(output_tree, muondis_tree):
+            mu = muondis_event.InMuon[0]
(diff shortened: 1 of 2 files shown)


olantwin:
pre-commit.ci autofix

olantwin (approved):
Maybe better to use FairRoot I/O eventually, but good enough for a start.
To be merged after #510.

olantwin:
For iterating over the two trees, you might want to use zip to avoid the GetEntry business

macro/run_simScript.py
```diff
@@ -650,6 +650,41 @@
...
+
+if simEngine == "muonDIS":
+    f_outputfile = ROOT.TFile.Open(outFile, "read")
+    output_tree = f_outputfile.cbmsim
+
+    f_muonfile = ROOT.TFile.Open(inputFile, "read")
+    muondis_tree = f_muonfile.Get("DIS")
+
+    iMuon = ROOT.TClonesArray("TVectorD")
+    muondis_tree.SetBranchAddress("InMuon", iMuon)
```
olantwin, line 662: Is this still needed now?

macro/run_simScript.py
```diff
@@ -650,6 +650,41 @@
...
+
+    temp_filename = outFile.replace(".root", ".tmp")
```
olantwin, line 664: In case there is any chance that the file survives the process, it might be good to keep the ROOT extension. I'd rather add something before the extension.

macro/run_simScript.py
```diff
@@ -650,6 +650,41 @@
...
+        mu = muondis_event.InMuon[0]
+        cross_section[0] = mu[10]
+        new_tree.Fill()
+
+    new_tree.Write("", ROOT.TObject.kOverwrite)
+    f_temp.Close()
+    f_outputfile.Close()
+    f_muonfile.Close()
+
+    os.replace(temp_filename, outFile)
```
olantwin, line 684: It might be good to use context managers here to ensure all files are left in a good state even if something fails.

For the temporary file, maybe https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile would be good to use, but we can probably also handle the cleanup ourselves using a context manager and `finally`.

macro/run_simScript.py
anupama-reghunath: true! this is now removed.

macro/run_simScript.py
anupama-reghunath: changed this to _tmp.root

macro/run_simScript.py
```diff
@@ -650,6 +650,38 @@
...
+if simEngine == "muonDIS":
+
+    temp_filename = outFile.replace(".root", "_tmp.root")
+
+    with (
+        ROOT.TFile.Open(outFile, "read") as f_outputfile,
+        ROOT.TFile.Open(inputFile, "read") as f_muonfile,
+        ROOT.TFile.Open(temp_filename, "recreate") as f_temp,
+    ):
+        output_tree = f_outputfile.cbmsim
```
olantwin, line 663: Because of #644, it might be better to use `TFile::Get()` until we have switched to 6.32+ only.

macro/run_simScript.py
```diff
@@ -650,6 +650,38 @@
...
+
+        muondis_tree = f_muonfile.DIS
```
olantwin, line 665: Same here.

macro/run_simScript.py
```diff
@@ -650,6 +650,38 @@
...
+            mu = muondis_event.InMuon[0]
+            cross_section[0] = mu[10]
+            new_tree.Fill()
+
+        new_tree.Write("", ROOT.TObject.kOverwrite)
+
+    os.replace(temp_filename, outFile)
+    print("Successfully added DISCrossSection to the output file:", outFile)
+
+# ------------------------------------------------------------------------
```
olantwin, line 684: Remove this line.

olantwin (changes requested):

olantwin:
Please, also add a CHANGELOG entry.

olantwin (approved):

olantwin:
Thanks a lot!

PR #661, opened 2025-03-13
feat(digi): move digiStraw to STL vector

olantwin:


wcmartylee (approved):
Seems to be helping in fixing #580. The error messages dropped from (normally) 386 to ~140!

olantwin:
> Seems to be helping in fixing #580. The error messages dropped from (normally) 386 to ~140!

I find this puzzling...
If this changes anything, maybe the pattern recognition/tracking is more broken than we thought?

wcmartylee:
Might be a big hint on problems lying in `digitizeStrawTubes`. Either way, we start from there.

Issue #519, opened 2024-09-25
Segmentation Fault when running ShipReco.py


anupama-reghunath:

mferril:
Hi @olantwin @anupama-reghunath 

FYI: I have also seen some unexpected behaviour when launching ```ShipReco.py``` on just 2 events, as a test. The output file has been successfully created, it's uncorrupted. However the script ends in Segmentation violation:

```bash
[FairShip/latest] FairShip/macro $> python ShipReco.py -f test/ship.conical.EvtCalc-TGeant4.root -g test/geofile_full.conical.EvtCalc-TGeant4.root -n 2
configured to process  2  events from  test/ship.conical.EvtCalc-TGeant4.root  starting with event  0  with option Yheight =  None  with vertexing True  and real pattern reco  
[INFO] Media file used: sw/slc9_x86-64/FairShip/master-local1/geometry/media.geo
Info in <ecal::ecal>: Geometry is read from file ecal_rect5x10m2.geoz3613.6000000000004.geo.
Info in <ecal::ecal>: Size of cell of type 2 is 5.970000 cm.
Info in <hcal::hcal>: Geometry is read from file hcal_rect.geoz3713.6000000000004.geo.
Info in <hcal::hcal>: Number of modules is 924, lightmap none
Info in <hcal::hcal>: Size of cell of type 22 is 23.959999 cm.
Info in <TGeoManager::CloseGeometry>: Geometry loaded from file...
Info in <TGeoManager::SetTopVolume>: Top volume is cave. Master volume is cave
Info in <TGeoNavigator::BuildCache>: --- Maximum geometry depth set to 100
Info in <TGeoManager::Voxelize>: Voxelizing...
Info in <TGeoManager::CountLevels>: max level = 6, max placements = 897
Info in <TGeoManager::CloseGeometry>: 5342106 nodes/ 728 volume UID's in FAIR geometry
Info in <TGeoManager::CloseGeometry>: ----------------modeler ready----------------
Calling addVMCFields
Creating field map called MainSpecMap using sw/slc9_x86-64/FairShip/master-local1/files/MainSpectrometerField.root
ShipBFieldMap::readMapFile() creating field MainSpecMap using file sw/slc9_x86-64/FairShip/master-local1/files/MainSpectrometerField.root
x limits: -470, 470, dx = 10
y limits: -405, 405, dy = 10
z limits: -990, 990, dz = 10
Offsets: x = 0, y = 0, z = 3068
[...]

Issue #453, opened 2024-04-12
Segmentation fault on exit of ShipRec.py when using splitcal


olantwin:
Probably something is freed twice, so it should be possible to avoid this by disowning something somewhere...

No success yet figuring out what the culprit is.

CC: @matclim 

Discovered by Walter Bonivento

olantwin:
Possibly related to #519 ?

Something about not using the old detectors causes issues with the reconstruction.

PR #664, opened 2025-03-13
Fix ShipReco seg fault!

olantwin:


olantwin:
This does indeed fix #519 :champagne: 
And would also explain how to fix #453 .
Now we just need to update ShipAna to fail when ShipRPCPoints and TargetPoints are not present.

anupama-reghunath (approved):

PR #667, opened 2025-03-15
Fix dark photon production

amagnan:
Tested and fix DP production in all modes:

- fix parameter to choose meson mother by name
- fix change of outputDir taken from options


olantwin (approved):
Thanks a lot!

PR #669, opened 2025-03-19
Add CHANGELOG entry for removing support for old geometries

anupama-reghunath:


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ it in future.
 ### Removed
 
 * fix: Remove unused, unrunnable shipPatRec_prev.py
+* fix(reco): Dropping support for old geometries without DecayVolumeMedium explicitly set(pre 24.11 release case).
```
olantwin, line 43: I would use present or past tense. I'm also not sure whether `fix(reco)` really is the appropriate tag. [`feat(geometry)!`](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change) might be better.

olantwin (changes requested):

CHANGELOG.md
anupama-reghunath: thanks will do this, how are these tags different?

olantwin (approved):

PR #676, opened 2025-03-25
chore(CI): add EventCalc to CI

olantwin:


olantwin:
After 4 hours, it seems like only about 40 events are transported.
This makes the EventCalc generator far too slow to include in the CI, and until the performance is fixed, not suitable for general use.

olantwin:
Let's see whether HNLs run more quickly.

mferril:
> After 4 hours, it seems like only about 40 events are transported. This makes the EventCalc generator far too slow to include in the CI, and until the performance is fixed, not suitable for general use.

Something must be really off. I have tested it with a local build of FairShip and it takes ca. 15 minutes for 50 events (of which ca. 5 for the GEANT4 initialisation).

mferril:
> I can reproduce the crash on lxplus with 100 events (no crash for 1). The crash doesn't occur with other generators.

@olantwin can you please send me on mattermost the error log? I will look into this, as it seems quite strange to me. Thanks!

olantwin:
> > I can reproduce the crash on lxplus with 100 events (no crash for 1). The crash doesn't occur with other generators.
> 
> @olantwin can you please send me on mattermost the error log? I will look into this, as it seems quite strange to me. Thanks!

It's the same one as on the CI: https://github.com/ShipSoft/FairShip/actions/runs/14083794149/job/39442804814

BTW: I had one suspicion for the bad performance (reading from CVMFS instead of from a local file), but that does not seem to be the issue.

PR #679, opened 2025-03-28
Move latest old SND design to yaml file

amagnan:
- duplicated geometry_config to remove almost all parameters related to SND
- in shipDet_conf, move the SND-related stuff to a specific configure_snd_old method reading parameters from a Yaml file, similarly as what is done for SBT in the configure_veto method.
- Kept only parameters and design related to NuTauTargetDesign=4, the latest.
- in the run_simScript file, the simEngine "Nuage" will be broken as it uses  SND parameters from ship_geo, it needs to be updated appropriately if still used ?

- in principle this PR should come after the one from Antonio cleaning up old SND stuff everywhere to keep only design 4.


amagnan:
I think that for the new design, we may want separate yaml and configure methods for each part, Emulsion, SiW and MTC, what do you think ? Passing the z coordinate in some way, and depending on overall decision to use which subdet for the full SND...

olantwin:
pre-commit.ci autofix

antonioiuliano2 (approved):
Thank you Anne-Marie,
This should simplify the update to new SND configurations.

Will the new version also require yaml file, or all parameters will be in the geometry_config.py file?

I would like to avoid splitting parameters into multiple files, at least in the final/default version. Otherwise, people in the future may be unable to find where the geometry of their subdetector is located.

antonioiuliano2 (commented):
The automatic check seems to complain about extrusions-overlaps.
Let us check them again after my other pull request is merged.

amagnan:
@antonioiuliano2 my idea is to have a yaml file with the new SND, and avoid having parameters in the geometry_config, but OK to have only one yaml file with the 3 subdetectors, I think that is fine!

geometry/geometry_config_noSND.py
olantwin, line 1: I don't see the benefit of renaming this file.

macro/run_simScript.py
```diff
@@ -382,6 +382,8 @@
  primGen.SetTarget(0., 0.)
  Nuagegen = ROOT.NuageGenerator()
  Nuagegen.EnableExternalDecayer(1) #with 0 external decayer is disable, 1 is enabled
+
+ #CAMM - This is broken now, need info from dedicated SND geo...
```
olantwin, line 386: @antonioiuliano2 Can we drop the NuageGenerator? I don't remember anyone using it in the last decade, and I'm pretty sure it wouldn't work right now anyway.

olantwin (changes requested):
Looks good, but I would note rename `geometry_config.py`

geometry/geometry_config_noSND.py
amagnan: Now I removed the duplicate and just updated geometry_config.py.

python/shipDet_conf.py
```diff
@@ -141,14 +293,15 @@ def configure(run, ship_geo):
...
         ship_geo.cave.floorHeightTankA = 4.5 * u.m
         ship_geo.cave.floorHeightTankB = 2.0 * u.m
-    if not hasattr(ship_geo, "NuTauTT"):
-        ship_geo.NuTauTT = AttrDict(z=0 * u.cm)
-    if not hasattr(ship_geo.NuTauTT, "design"):
-        ship_geo.NuTauTT.design = 0
+    #if not hasattr(ship_geo, "NuTauTT"):
+    #    ship_geo.NuTauTT = AttrDict(z=0 * u.cm)
+    #if not hasattr(ship_geo.NuTauTT, "design"):
+    #    ship_geo.NuTauTT.design = 4
```
olantwin, line 299: Remove these lines.

olantwin (changes requested):
I think we just need to decide what to do about Nuage and then this can be merged.

python/shipDet_conf.py
amagnan: done

python/shipDet_conf.py
```diff
@@ -89,6 +89,158 @@ def posEcal(z, efile):
...
 
+def configure_snd_old(yaml_file,
+                      snd_tauMu_zTot,
+                      snd_tauMu_zMudetC,
+                      cave_floorHeightMuonShield):
+
+    with open(yaml_file) as file:
+        config = yaml.safe_load(file)
+
+    #print(config)
```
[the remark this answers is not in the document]
amagnan: done

amagnan:
For nuage, I'd be in favour to make it part of the next big update of the run_simScript ? I kept the comments that it is broken from now on (and maybe already was anyway...)

olantwin (approved):

PR #684, opened 2025-04-04
Added dump_event() function to analysis_toolkit

anupama-reghunath:
This is a modified `dump_event()` function inspired by the `dump()` function in ShipAna.py.
Output for an event (produced using `analysis_example.py `:
Screenshot from 2025-04-04 16-32-45


olantwin (approved):
Great addition. Thanks!

PR #687, opened 2025-04-08
Review and fix of pattern recognition

wcmartylee:
Some minor bugs with PatRec:

- hard-coded straw diameter
- double append of recognized tracks

Other changes/fix may follow.

### DIFF
--- a/python/shipPatRec.py
+++ b/python/shipPatRec.py
@@ -911,7 +911,6 @@ def tracks_combination_using_extrapolation(recognized_tracks_12, recognized_trac
             atrack['hits_y34'] = recognized_tracks_34[i_34]['hits_y']
             atrack['hits_stereo34'] = recognized_tracks_34[i_34]['hits_stereo']
             recognized_tracks_combo.append(atrack)
-            recognized_tracks_combo.append(atrack)
 
     return recognized_tracks_combo
 
--- a/python/shipPatRec.py
+++ b/python/shipPatRec.py
@@ -7,7 +7,7 @@ import global_variables
 ReconstructibleMCTracks = []
 theTracks = []
 
-r_scale = global_variables.ShipGeo.strawtubes.InnerStrawDiameter / 1.975
+r_scale = global_variables.ShipGeo.strawtubes.InnerStrawDiameter / 1.9928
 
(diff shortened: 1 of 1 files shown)


python/shipPatRec.py
```diff
@@ -352,7 +352,7 @@ def fast_hough_pat_rec_stereo_views(SmearedHits_stereo, recognized_tracks_y, min
                 if ahit2['digiHit'] in used_hits:
                     continue
 
-                if abs(ahit1['zx_projection']) > 300 or abs(ahit2['zx_projection']) > 300:
+                if abs(ahit1['zx_projection']) > 200 or abs(ahit2['zx_projection']) > 200:
```
olantwin, line 355: Maybe good to define a const instead of repeating these two numbers everywhere. Please use explicit units for clarity.

python/shipPatRec.py
```diff
@@ -911,7 +911,6 @@ def tracks_combination_using_extrapolation(recognized_tracks_12, recognized_trac
             atrack['hits_y34'] = recognized_tracks_34[i_34]['hits_y']
             atrack['hits_stereo34'] = recognized_tracks_34[i_34]['hits_stereo']
             recognized_tracks_combo.append(atrack)
-            recognized_tracks_combo.append(atrack)
```
olantwin, line 914: Please double-check where this is used, to ensure that nothing depends on entries being duplicated (I hope not, but better to make sure)

python/shipPatRec.py
```diff
@@ -7,7 +7,7 @@
 ReconstructibleMCTracks = []
 theTracks = []
 
-r_scale = global_variables.ShipGeo.strawtubes.InnerStrawDiameter / 1.975
+r_scale = 1.
```
olantwin, line 10: Do we need this scale factor?

olantwin (changes requested):
Looks, good. Would be nice to add changelog entries to the individual commits.

python/shipPatRec.py
wcmartylee: I’m not sure. Would need some time to see if I can find a reference.

olantwin (approved):

PR #692, opened 2025-04-14
feat(digi): add experimental RDF digitisation

olantwin:


olantwin:
This experiment won't be continued, but thanks to the ROOT team, proof-of-concept code is available.

PR #696, opened 2025-04-15
Add UBT hit info of incoming muon for MuonDIS

anupama-reghunath:


olantwin (approved):
LGTM, but we need to replace those TClonesArrays soon.

PR #698, opened 2025-04-15
Assume all tracks are muons during track fit

wcmartylee:
The MC track matching for tracks from real pattern recognition doesn’t work. For simplicity, use muon as particle hypothesis during track fit. Use of PID information for hypothesis to be studied.


olantwin (changes requested):
Please add a change-log entry about the assumption of having a muon to avoid using MC truth information

olantwin (approved):

Issue #354, opened 2020-06-05
TVector3 UpstreamTaggerHit::GetXYZ()  Does not work


ThomasRuf:
TVector3 UpstreamTaggerHit::GetXYZ()
{
    Double_t hit_final;
    Int_t mod;
    TGeoNavigator* nav = gGeoManager->GetCurrentNavigator();
    TGeoNode* node = GetNode(hit_final, mod);
    ....  crash in GetNode

TGeoNode* UpstreamTaggerHit::GetNode(Double_t &hit_final, Int_t &mod){

(this is a very complicated piece of code to derive the path to the detector element to which the hit belongs to. Are there no other ways to do this?)

anyhow, the reason why it crashes seems is here:
Double_t XHit = point_final[1]; 

Note: Double_t YHit = point_final[0];  is OK.

reading https://root.cern.ch/root/htmldoc/guides/users-guide/InputOutput.html , I think it is missing the declaration about the length of the array. 
Double_t point_final[3]; //[3] 

fixes the problem.

HOWEVER: the hit classes are supposed to be lite weight objects representing the output of the electronics. So, essentially the detector ID, an ADC or/and TDC value, nothing else. Your objects are heavy. I don't think the DAQ people would be happy to send all these bytes from the detector to the event builder. Could you please reconsider the design of this class? If you want to have another class more related to reconstructed objects, like we have for Ecal clusters for example, be free to do so.

PR #691, opened 2025-04-14
UBT simplified to a SP with 4x6m, all reference to RPC remove

ggvasquez:
UpstreamTagger now is a scoring plane (vacuum) material with a 440x640cm**2 with a thickness of 8cm.

PR #699, opened 2025-04-15
modified GetXYZ method on UBT Hit to work with scoring plane

ggvasquez:
This fix the GetXYZ method on UpstreamTaggerHit, based on the new geometry of UBT. There is no need of search for modules since now is a single volume. 
This fix issue #354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ it in future.
 * fix: don't use TFile's deprecated attribute syntax
 * fix(digi): Fix logic of SST digitisation (#662)
 * Fix double append of recognized stracks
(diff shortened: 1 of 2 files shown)


olantwin:
pre-commit.ci autofix

olantwin:
Does this depend on your other PR?

ggvasquez:
Yes, the modification dependes on previous PR. After the PR I will address the other two issues about UBT.

olantwin:
Given that you deleted the bodies of most of the methods of UpstreamTaggerHit in #691, and I don't see a definition of the GetNode without arguments, I'm not sure how this is supposed to work.

olantwin:
Any updates on this?

PR #705, opened 2025-04-17
Clean up muon shield flags

olantwin:


Gfrisella:
Why are we still using muShieldDesign flag? 
( muShieldDesign=8)
There are other dependencies?

olantwin:
> Why are we still using muShieldDesign flag? ( muShieldDesign=8) There are other dependencies?

Because it wasn't cleanup up yet. This removes it from run_simScript.py, next step would be to remove it in geometry_config.py etc.

PR #721, opened 2025-05-13
Charmdet run 4

olantwin:
Changes needed for the run 4 charm detector.

Step one: figure out what can be reused from old code.

New development (also as laboratory for FairShip changes) will happen in https://github.com/olantwin/ruffiano for now (which will be pure python, depending on FairShip, if things work as imagined).


olantwin:
pre-commit.ci autofix

PR #747, opened 2025-06-03
 Create new MS design for new HA design

Gfrisella:
Added MS design `New_HA_Design` based on the MS `warm_opt`  with new version of HA. I increased the length of the first section magnet to keep roughly the same integrated field in the first section.


olantwin:
pre-commit.ci autofix

PR #777, opened 2025-06-22
makeCascade: formatting and change of target

THanae:
Small updates to makeCascade
- Target default was changed from Mo to W to follow the June 2025 target config. The target composition was also added as an argument with two options: W and Mo (depending on the target composition, the proton ratio changes slightly to either 0.40 for W or 0.43 for Mo)
- Changed `getopt` to `argparse` to retrieve arguments from the command line (for consistency with other files, and in order to add a description for each argument)
- Added more arguments to retrieve from the command line: the choice of tune (LHCb or PoorE791), the target composition as mentioned above, and parameters used for generating the chi. There is always a default argument added so those parameters do not need to be used
- Tried to be a bit more consistent with the spacing to improve readability


olantwin (changes requested):
Generally looks good (well, at least better and more understandable than before)

olantwin (approved):

PR #778, opened 2025-06-23
Changed the definition of fsmearBeam in shipgen/MuonBackGenerator.cxx

Gfrisella:


shipgen/MuonBackGenerator.cxx
```diff
@@ -165,10 +165,13 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
   }
   if (fPhiRandomize){phi = gRandom->Uniform(0.,2.) * TMath::Pi();}
   if (fsmearBeam > 0) {
-     Double_t r = fsmearBeam + 0.8 * gRandom->Gaus();
-     Double_t _phi = gRandom->Uniform(0., 2.) * TMath::Pi();
-     dx = r * TMath::Cos(_phi);
-     dy = r * TMath::Sin(_phi);
+      Double_t x = gRandom->Gaus(0, 0.8);
+      Double_t y = gRandom->Gaus(0, 0.8);
```
olantwin, line 169: Why introduce these intermediate variables, if you only use them to assign to dx and dy?

olantwin (changes requested):
A future improvement would be to expose the smearing parameters as a config option or command-line argument to the simulation script.

Looks good apart from that.

shipgen/MuonBackGenerator.cxx
```diff
@@ -23,19 +23,19 @@ MuonBackGenerator::MuonBackGenerator() {
 // -------------------------------------------------------------------------
 // -----   Default constructor   -------------------------------------------
 Bool_t MuonBackGenerator::Init(const char* fileName) {
-  return Init(fileName, 0, false);
+  return Init(fileName, 0, 0.8, 5); // default values for firstEvent, fPaintBeam, fsmearBeam
 }
 // -----   Default constructor   -------------------------------------------
-Bool_t MuonBackGenerator::Init(const char* fileName, const int firstEvent, const Bool_t fl = false ) {
+Bool_t MuonBackGenerator::Init(const char* fileName, const int firstEvent, const Double_t fl, const Double_t sb) {
```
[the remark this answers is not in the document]
olantwin: Then let's take the opportunity to fix `fl` too

shipgen/MuonBackGenerator.cxx
```diff
@@ -23,19 +27,19 @@ MuonBackGenerator::MuonBackGenerator() {
 // -------------------------------------------------------------------------
 // -----   Default constructor   -------------------------------------------
 Bool_t MuonBackGenerator::Init(const char* fileName) {
-  return Init(fileName, 0, false);
+  return Init(fileName, 0); // default values for firstEvent, fPaintBeam, fsmearBeam
```
olantwin, line 30: Misleading comment, now that the arguments are removed

olantwin (changes requested):

shipgen/MuonBackGenerator.cxx
```diff
@@ -163,12 +167,15 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
-  if (fPhiRandomize){phi = gRandom->Uniform(0.,2.) * TMath::Pi();}
-  if (fsmearBeam > 0) {
-     Double_t r = fsmearBeam + 0.8 * gRandom->Gaus();
-     Double_t _phi = gRandom->Uniform(0., 2.) * TMath::Pi();
-     dx = r * TMath::Cos(_phi);
-     dy = r * TMath::Sin(_phi);
+  if (fsmearBeam > 0){
+    dx = gRandom->Gaus(0, fsmearBeam);
+    dy = gRandom->Gaus(0, fsmearBeam);
+    phi = gRandom->Uniform(0., 2*TMath::Pi());
```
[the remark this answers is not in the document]
Gfrisella: Truee! Indeed, it is not used here

shipgen/MuonBackGenerator.cxx
```diff
@@ -177,7 +184,7 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
          px = track->GetPx();
          py = track->GetPy();
          pz = track->GetPz();
-         if (fPhiRandomize) {
+         if (fPaintBeam > 0) {
```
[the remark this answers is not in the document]
Gfrisella: I'm not sure. If you don't give a phi to the PaintBeam it means that you are only giving an offset to the beam

shipgen/MuonBackGenerator.cxx
```diff
@@ -163,21 +167,23 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
+  } 
   if (id==-1){
      for (unsigned i = 0; i< MCTrack->GetEntries();  i++ ){
          auto* track = dynamic_cast<ShipMCTrack*>(MCTrack->At(i));
          Int_t abspid = TMath::Abs(track->GetPdgCode());
          px = track->GetPx();
          py = track->GetPy();
          pz = track->GetPz();
-         if (fPhiRandomize) {
+         if (fPaintBeam > 0) {
```
olantwin, line 186: This should be, and should *not* reuse the same $\phi$ as above

```suggestion
         if (fPhiRandomize) {
```

shipgen/MuonBackGenerator.h
```diff
@@ -45,7 +49,7 @@ class MuonBackGenerator : public FairGenerator
   int fNevents;
   float f_zOffset;      //!
   int fn;
-  Bool_t fPhiRandomize;
```
olantwin, line 48: Let's keep this and add a method to set this flag

olantwin (changes requested):
Sorry, that I initially didn't notice the confusion about the $\phi$ randomisation.

macro/run_simScript.py
```diff
@@ -72,6 +72,7 @@
 parser.add_argument("--FollowMuon",dest="followMuon", help="Make muonshield active to follow muons", required=False, action="store_true")
 parser.add_argument("--FastMuon",  dest="fastMuon",  help="Only transport muons for a fast muon only background estimate", required=False, action="store_true")
 parser.add_argument("--Nuage",     dest="nuage",  help="Use Nuage, neutrino generator of OPERA", required=False, action="store_true")
+parser.add_argument("--phiRandom", dest="phiRandom",  help="only relevant for muon background generator, random phi", required=False, action="store_true")
```
olantwin, line 75:
```suggestion
parser.add_argument("--phiRandom", help="only relevant for muon background generator, random phi", action="store_true")
```

shipgen/MuonBackGenerator.cxx
```diff
@@ -183,11 +184,13 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
          px = track->GetPx();
          py = track->GetPy();
          pz = track->GetPz();
-         if (fPaintBeam > 0) {
+         if (fPhiRandomize) {
+             gRandom->SetSeed(12345);  // TO BE FIXED: this should not be hardcoded
```
olantwin, line 188: Remove this line for now, as it will also affect the ability to reproduce the muon background.
If we can't set an independent seed, it's better to not set it.

shipgen/MuonBackGenerator.cxx
```diff
@@ -183,11 +184,13 @@ Bool_t MuonBackGenerator::ReadEvent(FairPrimaryGenerator* cpg)
...
+             double phi_rm = gRandom->Uniform(0., 2*TMath::Pi());
              Double_t phi0 = TMath::ATan2(py, px);
              Double_t pt = track->GetPt();
-             px = pt * TMath::Cos(phi + phi0);
-             py = pt * TMath::Sin(phi + phi0);
+             px = pt * TMath::Cos(phi_rm + phi0);
```
olantwin, line 192: Probably no need to calculate phi0 or performing the addition, as the distribution of (uniform(0, 2pi) + const) % 2pi should be the same as without the constant.

olantwin (changes requested):
Looks good, but for now I would not set the seed unless we can figure out how to set a second, independent one.

olantwin (approved):

PR #792, opened 2025-06-26
PG updated

eduard322:
sndsw-based PG


CHANGELOG.md
```diff
@@ -89,6 +90,9 @@ it in future.
 * Removed old stuff legacy in `shipMuonShield.cxx` (like LE and fFloor)
 * Event Display: Don't try to recreate geometry config
 * Geometry: Make the tungsten target the default (Jun25 config)
+=======
+* Particle Gun has been updated: now user can set the coordinates of the PG via keys --V{x,y,z} and use --D{x,y} to uniformly smear the signal in a given x and y range
```
olantwin, line 94: Please resolve the merge conflicts instead of committing the conflict markers

olantwin (changes requested):
Full review probably Monday

olantwin (changes requested):
Apart from the merge conflict, my main concern is that this leads to a further proliferation of command-line options. Maybe it would be cleaner if we make the PG a subcommand with it's option parsed using a second parser?

eduard322:
> Apart from the merge conflict, my main concern is that this leads to a further proliferation of command-line options. Maybe it would be cleaner if we make the PG a subcommand with it's option parsed using a second parser?

okay, I see. Will it be okay like this for example (with subparser):



```python $FAIRSHIP/macro/run_simScript.py PG --pID 11 --Estart 50 --multiple --Dx 10```



?

PR #800, opened 2025-07-04
Proximity shielding update

olantwin:
Todo:

- [x] update readme
- [x] fix overlap with pedestal
- [x] make sure distance to hadron absorber is correct

PR #803, opened 2025-07-06
Remove hadron absorber from fixed target

THanae:
Aim is to remove outdated hadron absorber from `FixedTarget` code. At the same time, files which define hadron absorber in `FixedTarget` are updated. 
Notes:
- `run_fixedTarget` was updated but is not in a running state (it isn't in a running state in the master version either) -- this is addressed separately in PR 730
- `run_simPgun` could not run with just the changes to the `FixedTarget` so a series of changes were added to make it run. In particular `ship_geo` was defined and `Veto` and `Muon` needed additional dimension/position information.

### DIFF
--- a/macro/run_simPgun.py
+++ b/macro/run_simPgun.py
@@ -1,28 +1,35 @@
+import yaml
+import ROOT,os
+import atexit
+import shipunit as u
+
+from ShipGeoConfig import ConfigRegistry, AttrDict
+
+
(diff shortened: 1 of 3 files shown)


olantwin:
It might be worth rebasing this on #800, as some changes are made already there, and that PR will likely be merged first.

olantwin:
What is missing for this PR?

PR #810, opened 2025-07-13
Updated if statement to account for target composition

THanae:
In the `if ` statement which determines the fraction of protons to neutrons, the wrong argument was used (`target` instead of `args.target_composition`. This is changed with this PR.


olantwin:
Thanks!

PR #818, opened 2025-07-16
Structure python directory as a python package

barskern:
A start of restructuring the python code to a package/module. This allows easier inclusions, cleaner interfaces and also makes it easier to work and maintain the code. The work has been done partly at the 2nd computing workshop of SHIP. The splitting and re-naming has been done at a best effort practice, and due to a lack of tools locally (specifically ROOT), these changes are not really tested. Hence please do clone these changes to try them out locally.

This PR also introduces UV, which is a great package manager for python. In this case, it seems as though "only" root and the standard libraries are used, but using UV could open the possibilities for easy use of other libraries. Also it allows managing multiple python versions seamlessly. See [their website](https://docs.astral.sh/uv/) for more info.


barskern:
@olantwin You are right about uv for the moment not really bringing anything to the table in this repo besides the minimal lockfile in which I simply included pytest for development purposes. Anyhow, it is how I generated the pyproject and it uses e.g. hatchling and some other modern tooling. As you say, the challenge will probably be to figure out how to define the dependencies that come from other sources than pypi.

barskern:
@olantwin So for the last part now, the fairship package has to be installed into the runtine environment setup by the `setUp.sh` script. I was looking a bit at it, but I don't know the intricacies of how its all put together. As you perhaps know already, could you tie it together?

olantwin:
> @olantwin So for the last part now, the fairship package has to be installed into the runtine environment setup by the `setUp.sh` script. I was looking a bit at it, but I don't know the intricacies of how its all put together. As you perhaps know already, could you tie it together?

Yes, I'll figure that part out.

PR #819, opened 2025-07-16
Remove deprecated pydocstyle

barskern:
https://github.com/PyCQA/pydocstyle

It was producing some warnings. Seems to be succeeded by ruff, which is already in the pre-commit config.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,0 +1,9 @@
+[lint]
+select = [
+    "D", # pydocstyle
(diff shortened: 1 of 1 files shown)


olantwin:
pydocstyle will also need to be removed from the skipped checks.

The pydocstyle rules are not enabled by ruff by default, we will need to enable the `D` set of rules.

olantwin (approved):

PR #821, opened 2025-07-24
Target production fixes

olantwin:


olantwin:
@kskovpen Could you explain why the change of target volume is needed?

I think we should specify the actual target, but maybe the naming changed with the modifications to the target. I don't think we want to have the muon shield here.

kskovpen:
@olantwin just committed your proposed changes in https://github.com/ShipSoft/FairShip/pull/821#issuecomment-3136541989

PR #823, opened 2025-07-30
Small fix to `run_fixedTarget`

mesmith75:


mesmith75:
Aha! #821 already fixes it.

PR #825, opened 2025-07-31
Use C++ style casts, don't respecify type where obvious.

olantwin:


SND/EmulsionTarget/Target.cxx
```diff
@@ -528,7 +528,7 @@ Bool_t Target::ProcessHits(FairVolume* vol)
                pdgCode);
 
         // Increment number of muon det points in TParticle
-        ShipStack* stack = (ShipStack*)gMC->GetStack();
+        ShipStack* stack = static_cast<ShipStack*>(gMC->GetStack());
```
egamberini, line 531: this could be `auto stack = static_cast<ShipStack*>(gMC->GetStack());` as the type is explicit in the `static_cast<>` for consistency.

PR #838, opened 2025-08-07
Make changelog more uniform to prepare for next release

olantwin:
This should make it easier to read, as it was very inconsistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -220,37 +220,32 @@ it in future.
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


olantwin:
Dropped the reformatting of the "Unreleased" section, which would have been a lot of work to resolve all the conflicts of.

~~This will be merged Monday at the latest, or probably earlier once the checks pass.~~ If nobody complained since August, I consider this approved.

Edit: Actually, no point in running the checks...

PR #841, opened 2025-08-07
feat!(sim): Use g4Config.yaml instead of deprecated .C

### LINKED ISSUE #543: Using gConfig.C macro DEPRACATED
```
[WARN] FairGenericVMCConfig::Setup() Using gConfig.C macro DEPRACATED.
[WARN] Check FairRoot/examples/common/gconfig/ for current YAML implementation.
```

This is something we should try to fix ASAP, before it limits what FairRoot version we can use.

olantwin:
Add YAML config including all settings from .C. Set ShipStack and its options explicitly in simulation scripts.

Thorough testing needed.

Fixes #543


olantwin:
Fails because `fairtools/MCConfigurator` isn't built when GEANT3 isn't found...

I'll return to this when I'm back (or left as exercise for the reader).

olantwin:
Build failure should be fixed once we switch the CI to a config using https://github.com/ShipSoft/shipdist/pull/115

PR #842, opened 2025-08-09
fix(ecal/hcal): include missing headers

olantwin:
@matclim How is the removal of HCAL/ECAL progressing? All this old code only causes problems.


matclim:
I had started to have a look and had some fixes but then got caught up with other things.

I'm unfortunately on vacation with limited access to the internet atm so I'm a bit limited in my ability to do much right now, I hope to remove the old ecal/hcal by early September at the latest.

PR #845, opened 2025-08-15
Updates on SST geometry

wcmartylee:
Several geometry updates on SST:
- Parameters moved to yaml
- Construct in cxx class updated
- Options of frame material added
- Straw sensitive area expanded for full coverage of aperture

This frame design is not final. An I-beam frame design will be integrated when ready.


wcmartylee:
> Could you prepare a presentation for next week's software meeting to give an overview of these changes?
> 
> At first glance this looks like a very nice cleanup, I'll try to review properly asap.

Sure thing!

wcmartylee:
@olantwin If this is ready to be merged, I’ll add an entry in CHANGELOG.

PR #867, opened 2025-09-23
Add generated changelog

barskern:
Here is an example `cliff.toml` to generate the CHANGELOG from the commit history.

It might be a worthty trade of to have a verbose CHANGELOG for the simplicity of not having to manually make changes to it. It does mean that the commit messages should be clear on what they do. Probably move to enforce conventional commits with a tool such as [`committed`](https://github.com/crate-ci/committed), and be more deliberate when merging PRs, that the commits are perhaps rebased or squashed to be meaningful.

The exact format and filtering of the changelog can be modified. For example, for now all commits concerning `style` or formatting are hidden from the changelog.

### DIFF
--- a/scripts/update-changelog.sh
+++ b/scripts/update-changelog.sh
@@ -3,6 +3,6 @@
 set -euxo pipefail
 
 # Commit ref were we want to start to use git-cliff for the changelog
-from_ref="f5620fe0"
+from_ref="0921ba5a"
 
 git-cliff -o CHANGELOG.md "$from_ref.."


olantwin:
Can we maybe configure this to start at commit X and have the old manually written changelog as footer?

For the script, it might be good to just assume git-cliff is installed. I'd probably install it locally on my different development machines.

barskern:
> Can we maybe configure this to start at commit X and have the old manually written changelog as footer?

Yeah, that we could certainly do! I will change this PR to implement that, and when you do the next release, we can do the necessary steps to start with the new generation.

> For the script, it might be good to just assume git-cliff is installed. I'd probably install it locally on my different development machines.

Sure, that is probably a bit cleaner. I will change it!

barskern:
Thank you for your patience regarding the update spam. I think I should meet your goals now. We can keep the PR open until you are ready to make the transition, and then we can rebase on master, update the footer, update the `from_ref` commit sha and make the merge.

Issue #796, opened 2025-07-02
Neutrino momentum truncated to 50GeV in 2018 pre-production


anupama-reghunath:
![Image](https://github.com/user-attachments/assets/a489374e-085d-4860-a86c-60fa1e97c3c2)

Attached are the two plots from the neutrino preproduction sample  (/eos/experiment/ship/data/Mbias/background-prod-2018/pythia8_Geant4_1.0_withCharm_nu.root). 

Does anyone know why we force the P_total < 50GeV in logP vs logPt (right plot) ?

Relevant code here, but no supporting comments unfortunately.
https://github.com/ShipSoft/FairShip/blob/5364119cdf3cf33c77d1ea46e091fe4fde972e6d/macro/makeDecay.py#L142C1-L143C75

antonioiuliano2:
Hello Anupama,
The 2D spectrum is used here:

https://github.com/ShipSoft/FairShip/blob/5364119cdf3cf33c77d1ea46e091fe4fde972e6d/shipgen/GenieGenerator.cxx#L486

It is used to generate the transverse component of the neutrino momenta, since GENIE will always generate them along z.
It does not impede to simulate neutrinos over 50 GeV (we simulated them so far), but it will consider it as a 50 GeV one to randomly generate the PT according to the last slice of the 2D plot

Unfortunately, I do not know why the momentum is truncated in makeDecay. I would assume due to low statistics at high momentum, but we should write to Thomas to be sure.

PR #868, opened 2025-09-25
Python scripts for genie simulation

eduard322:
Major updates for neutrino generations via `macro/makeGenieEvents.py` and genie regime in `macro/run_simScript.py`. See Changelog


anupama-reghunath:
Hello Eduard, does this address #796 ?

PR #876, opened 2025-10-14
feat(geo): Enclose target in steel cylinder

olantwin:
Sketch of cross-section (target not to scale, is in contact with steel)

target_enclosure_cross_section


olantwin:
A gap of 2 mm between target and enclosure has been added to avoid overlap.

PR #896, opened 2025-10-23
refactor: remove unused CaloDesign parameter

olantwin:
After the ECAL/HCAL removal in commit 2b10248, only the splitCal design remains supported (HcalOption=-1, EcalOption=2). The CaloDesign parameter was kept in the API but completely unused.

This commit removes CaloDesign from:
- geometry_config.py: removed default value
- run_simScript.py: removed from globalDesigns, argparse, and function call
- create_field_perturbation.py: removed from globalDesigns and function call
- TTCluster.py: removed from design2018 dict and function call


matclim (approved):

matclim:
So I ran the simulation and that works of course but I'm unable to run the event display on the built geometry, I just get a seg violation.

olantwin:
With the rebase this works for me.

PR #901, opened 2025-10-30
refactor: remove ConfigRegistry and convert to function-based API

olantwin:
ConfigRegistry was over-engineered for the simple use case of loading one configuration named "basic". It provided no real benefit over a direct function call.

Changes:
- geometry_config.py: Convert to create_config() function with explicit parameters
- ShipGeoConfig.py: Remove ConfigRegistry class and _SingletonDict (~80 lines)
- Update 11 scripts to use geometry_config.create_config() instead of ConfigRegistry.loadpy()
- Remove python/test_shipGeoConfig.py (only tested ConfigRegistry itself)

Benefits:
- Clearer API with explicit parameter signature
- Better IDE support (autocomplete, type hints, refactoring), better error messages!
- No dangerous/clumsy exec() usage
- Simpler code with same functionality


olantwin:
FYI @siilieva, IIRC you are still using ConfigRegistry in SND@(HL)LHC?

PR #910, opened 2025-11-05
Add target parameters for dedicated studies

amagnan:
- just adding parameters to be set in the run_fixedTarget command.

### DIFF
--- a/muonShieldOptimization/exitHadronAbsorber.cxx
+++ b/muonShieldOptimization/exitHadronAbsorber.cxx
@@ -237,7 +237,7 @@ void exitHadronAbsorber::ConstructGeometry()
 
    TString myname(this->GetName());
    TString shapename_prefix(myname); // Use a prefix for shapenames
-   
+
    if (fzPos>1E8){
      TGeoVolume *sensPlane = gGeoManager->MakeBox(shapename_prefix+"_box",vac,3.56*m-1.*mm,1.7*m-1.*mm,1.*mm);
      std::cout << this->GetName() << ", ConstructGeometry(): Created Box with dimensions: 3.56*m-1.*mm,1.7*m-1.*mm,1.*mm" << std::endl;
@@ -256,7 +256,7 @@ void exitHadronAbsorber::ConstructGeometry()
      std::cout << this->GetName() << ", ConstructGeometry(): Created Tube with dimensions: 12.51*cm,12.52*cm,zLoc/2." << std::endl;
      nav->cd("/target_vacuum_box_1/TargetArea_1/HeVolume_1");
      nav->GetCurrentNode()->GetVolume()->AddNode(sensPlaneCyl, 1, new TGeoTranslation(0, 0, 0));
-     AddSensitiveVolume(sensPlaneCyl); 
+     AddSensitiveVolume(sensPlaneCyl);
    }
 
 }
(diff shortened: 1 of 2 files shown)


olantwin:
Looks good to me. Could you please add an entry to the changelog?

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -233,10 +234,31 @@ void exitHadronAbsorber::ConstructGeometry()
...
+     TGeoVolume *sensPlane = gGeoManager->MakeBox(shapename_prefix+"_box",vac,3.56*m-1.*mm,1.7*m-1.*mm,1.*mm);
+     std::cout << this->GetName() << ", ConstructGeometry(): Created Box with dimensions: 3.56*m-1.*mm,1.7*m-1.*mm,1.*mm" << std::endl;
+     nav->cd("/MuonShieldArea_1/");
+     //else nav->cd("/target_vacuum_box_1");
+     sensPlane->SetLineColor(kBlue - 10);
+     //nav->GetCurrentNode()->GetVolume()->AddNode(sensPlane, fzPos<250?2:1, new TGeoTranslation(0, 0, zLoc));
+     nav->GetCurrentNode()->GetVolume()->AddNode(sensPlane, 1, new TGeoTranslation(0, 0, zLoc));
+     AddSensitiveVolume(sensPlane);
+   }
+   else {
```
olantwin: I think I'd prefer it if we explicitly whether the plane should be circular or rectangular using a setter method.

olantwin (commented):
It might be good to run `git clang-format` to format the touched lines.
Looks good otherwise.

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -233,10 +234,31 @@ void exitHadronAbsorber::ConstructGeometry()
...
    }else{zLoc = fzPos;} // use external input
-   TGeoVolume *sensPlane = gGeoManager->MakeBox("sensPlane",vac,3.56*m-1.*mm,1.7*m-1.*mm,1.*mm);
-   nav->cd("/MuonShieldArea_1/");
-   nav->GetCurrentNode()->GetVolume()->AddNode(sensPlane, 1, new TGeoTranslation(0, 0, zLoc));
-   AddSensitiveVolume(sensPlane);
+
+   TString myname(this->GetName());
+   TString shapename_prefix(myname); // Use a prefix for shapenames
+
+   if (fzPos>1E8){
```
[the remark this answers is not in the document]
THanae: I don't think this value is ever used if it is above 1e8. `zloc` is used instead and does not depend on `fzPos` if it is too large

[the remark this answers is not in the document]
olantwin: So should we just remove that branch?

muonShieldOptimization/exitHadronAbsorber.cxx
THanae: `zloc` is hard-coded only for large `fzPos`, otherwise it is equal to `fzPos`

amagnan: Yes indeed the point here is that the sensitive plane after HA is completely hardcoded. By the way, is it large enough to cover the whole available surface after the HA ? We should make sure it is. But then zLoc is fixed by the end of the HA so it's perfect.

olantwin:
Closed in favour of #914, which seems to include the same changes.

PR #917, opened 2025-11-14
Update run_fixedTarget.py

mesmith75:
Set the default to false for the hadron shield


olantwin:
Given @amagnan's and @THanae's changes in #910 and #914, I think we need to update the help message rather than the default value.
Thanks for spotting this!

PR #918, opened 2025-11-14
fix: Don't specify removed muShieldDesign, fix help message

olantwin:
@mesmith75 This should fix both issues you identified.

`muShieldDesign` was removed but missed in run_fixedTarget.py.


THanae:
> `muShieldDesign`

This is fixed in https://github.com/ShipSoft/FairShip/pull/914

olantwin:
Agreed, this should be fixed ASAP. With all the different changes in #914, I hadn't noticed that there was an urgent fix included.

PR #921, opened 2025-11-16
feat: add CITATION.cff

olantwin:
This should be useful for Zenodo etc.


olantwin:
`license` field will be added in #872

PR #922, opened 2025-11-16
Remove StopTrack from sens plane

THanae:
StopTrack leads to unwanted behaviour when the sensitive plane is not at the end of the decay volume. 
I am not sure if it is better to leave it by default, and add an option to not use, or remove it at all times


olantwin:
For the cylindrical one, it should definitely be disabled. For the one after the HA, maybe it slightly improves performance to stop the track there instead of propagating it through the air. But probably negligible compared to tracking through the HA and target esp.

Issue #806, opened 2025-07-09
RNtuple migration


olantwin:
* Being able to write/read std::vectors instead of TClonesArrays (#710 #430 )
* Changing the ShipParticle class to be independent of TParticle, including schema evolution rules. Write again data in TTree and compare with existing data.
* Convert now RNTuple-ready format to RNTuple, compare with existing TTree data

olantwin:
First simulation file has now been written directly to RNTuple.

Issue #430, opened 2023-04-17
Investigate use of STL containers instead of TClonesArrays


olantwin:
STL containers have many advantages, but the migration will be tricky. This will likely be a major breaking change, to be performed after approval.

PR #933, opened 2025-11-18
Migrate simulation to STL containers

olantwin:
Addresses #430 and next step towards #806 (first RNTuple from simulatin written today :partying_face: )


olantwin:
Tested locally, the changes give numerically completely identical results with fixed seeds.

PR #944, opened 2025-11-20
Add track ID to MCTrack

mesmith75:
Help to keep track of things
FYI @THanae


mesmith75:
This works for the parts for the target production. We can add in the rest of the detector in another MR

olantwin (dismissed):

mesmith75:
Added the changelog. I think you can squash it when you merge.

PR #958, opened 2025-12-01
Nuhist cut

eduard322:
Fixed the 50 GeV cut for the production of P-Pt histograms of neutrino. I would investigate more what exactly this script does and what was the reason for that before using it. I don't think this is an urgent thing: it does not affect the production at all.

### DIFF
--- a/macro/makeDecay.py
+++ b/macro/makeDecay.py
@@ -142,8 +142,8 @@ for n in range(nEvents):
          if par.id()<0: idhnu+=1000
          pt2=par.px()**2+par.py()**2
          ptot=ROOT.TMath.Sqrt(pt2+par.pz()**2)
-         l10ptot=min(max(ROOT.TMath.Log10(ptot),-0.3),1.69999)
-         l10pt=min(max(ROOT.TMath.Log10(ROOT.TMath.Sqrt(pt2)),-2.),0.4999)
+         l10ptot= max(ROOT.TMath.Log10(ptot),-0.3)
+         l10pt= max(ROOT.TMath.Log10(ROOT.TMath.Sqrt(pt2)),-2.)
          h[str(idhnu)].Fill(ptot,wspill)
          h[str(idhnu+100)].Fill(l10ptot,l10pt,wspill)
          h[str(idhnu+200)].Fill(l10ptot,l10pt,wspill)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -208,6 +208,7 @@ it in future.
   - Update Python code in shipDigiReco.py to use modern constructors
   - Bump ClassDef versions to 2 for schema evolution
   - Add TrackInfo to RNTuple I/O test suite
+* Fixed the weird cut on P in the script that produces P and P-Pt histograms for neutrinos
 
 ### Removed
 


antonioiuliano2:
Yes, I agree with Oliver.

I would check it, because I am worried about how the high momentum bins with low statistics would be treated by GenieGenerator during the simulation.

antonioiuliano2:
> > Still no clue from my side, unfortunately. I was not able to find any relevant note on the gevgen_fnal attempt, except that we had some issue both in the Advanced SNDLHC configuration and the SHiP configuration due to high number of volumes to process (high granularity)?. The best is just to do a new attempt and take notes on everything weird I see. I can do it next week, between the SHiP Computing Workshop sessions.
> 
> Any progress on this?

Hello, I have made some test gevgen_fnal simulations and so far I have not encountered any showstopper. What I am doing is:

- launching a dummy Particle Gun muon simulation to have a geometry file of our detector;
- converting the geometry into GDML file;
- make the splines with gmkspl with this geometry file as input;
- launching gevgen_fnal providing as input the neutrino flux histogram and the geometry file;

Then, steps are as before:
- Convert the output ghep into gst format with gntpc;
- Add the TH2D histogram to the ROOT file with the gst TTree;
- Launch the usual run_simScript.py simulation with GENIE input file

I am now going to have a look if there is something out of place, and I will then move to compare the flux histograms with and without the energy cut.

antonioiuliano2:
> How do we proceed with this @eduard322 @antonioiuliano2 ?

Even if we do not plan to use them anymore for the full production simulations, 
I would still have a look at the old and new production neutrino momentum 2D histograms with and without this cut, since they are useful info for quick comparisons of neutrino distributions.

olantwin:
> > How do we proceed with this @eduard322 @antonioiuliano2 ?
> 
> Even if we do not plan to use them anymore for the full production simulations, I would still have a look at the old and new production neutrino momentum 2D histograms with and without this cut, since they are useful info for quick comparisons of neutrino distributions.

Ok. What does this mean for the pull request?

antonioiuliano2:
> > > How do we proceed with this @eduard322 @antonioiuliano2 ?
> > 
> > 
> > Even if we do not plan to use them anymore for the full production simulations, I would still have a look at the old and new production neutrino momentum 2D histograms with and without this cut, since they are useful info for quick comparisons of neutrino distributions.
> 
> Ok. What does this mean for the pull request?

For the pull request, as pointed by @anupama-reghunath we first need to ensure the same change is applied to all related code, otherwise we leave some scripts with the cut and some without it. By looking in the scripts I see it is applied in:

- [macro/makeDecay.py](https://github.com/ShipSoft/FairShip/blob/72a34171839b8ed26b1a5c27b7a63d7332ad61ab/macro/makeDecay.py#L169)
- [muonShieldOptimization/extractNeutrinosAndUpdateWeight.py](https://github.com/ShipSoft/FairShip/blob/72a34171839b8ed26b1a5c27b7a63d7332ad61ab/muonShieldOptimization/extractNeutrinosAndUpdateWeight.py#L72)
- [muonShieldOptimization/exitHadronAbsorber.cxx](https://github.com/ShipSoft/FairShip/blob/72a34171839b8ed26b1a5c27b7a63d7332ad61ab/muonShieldOptimization/exitHadronAbsorber.cxx#L174)
- [macro/mergeMbias.py](https://github.com/ShipSoft/FairShip/blob/72a34171839b8ed26b1a5c27b7a63d7332ad61ab/macro/mergeMbias.py#L447)

They all use the exact same cut, for the same histograms.

macro/makeDecay.py
antonioiuliano2, line 97: When we remove the cut, we also need to increase the range of the histogram, otherwise all the remaining entries would simply go to overflow.

antonioiuliano2 (changes requested):
I have finally found time to do a proper review of this pull request, my apologies for the (very long) delay.
These are the points I feel now confident to objectively raise:

- The cut is strictly related to the x range of the 2D histograms->if you remove or relax the cut, you also need to change the histogram range;
- The same histogram ranges and cut are used in multiple scripts, not only this one;
- The number of bins is then used by GenieGenerator to generate the neutrinos ptlog10 from the 2D histogram. It unfortunately does not scale well with an arbitrary range (i.e. increased until maximum neutrino momentum is reached, with a lot of empty bins);

In conclusion, I would suggest to close the pull request and keep the histograms as they are. I would then focus our efforts on using directly gevgen_fnal neutinos, and using the shipgen GenieGenerator only for propagation. We can also discuss it in the next meeting, @eduard322 , @olantwin

olantwin:
Thanks @antonioiuliano2 . Could you please check with Cris about his code again? Without his permission, we can't proceed with gevgen_fnal.

PR #959, opened 2025-12-01
Layer id fix

eduard322:
Fix the outdated method of the extraction of layer_id and layer_type for MTC.


olantwin (approved):
Is GetLayerType no longer needed?

PR #998, opened 2026-01-13
Suggestion: saving hits on post-target plane

THanae:
Saving tracks reaching post-target sensitive plane (currently, if a post-target sensitive plane is added, tracks are only saved if they also reach the post-hadron absorber sensitive plane). 

This would not be used for the main production, but could be useful for side studies (?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ it in future.
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


olantwin (approved):
Looks fine. For the production, we should check what happens to performance and file size if the option is enabled.

olantwin (approved):

PR #999, opened 2026-01-13
Draft:Remove energy cut on neutrinos for fixed target simulation

THanae:
Based on discussions in software meeting, identifying places where we might want to loosen energy cuts on neutrinos.

TODO: determine if want to remove cut entirely on neutrinos, set a small cut on neutrinos or set a cut on neutrino parent


olantwin:
@antonioiuliano2 , @eduard322 , thoughts?

antonioiuliano2:
Dear all,

As we discussed at the last software meeting, we would need a feedback from the physics meeting before confirming the energy cut. Strictly from the code side, everything is clear, so I have no comment on that. 

My only doubt is that applying a much lower cut to a specific class of particles (i.e. neutrinos), may have some effect on the decay chain from the mother particles (which would be cut at much higher energy).

Best Regards,
Antonio

PR #1027, opened 2026-02-06
add TChain to input files

mesmith75:
## Checklist

- [ ] CI checks pass


shipgen/MuonBackGenerator.cxx
```diff
@@ -82,9 +94,14 @@ Bool_t MuonBackGenerator::Init(const char* fileName, const int firstEvent) {
     }
   } else {
     id = -1;
-    fTree = fInputFile->Get<TTree>("cbmsim");
+    //    fTree = fInputFile->Get<TTree>("cbmsim");
+    fTree = new TChain("cbmsim");
+    for (auto& f : fileNames) {
+      LOG(info) << "Opening input file " << f;
+      static_cast<TChain*>(fTree)->Add(f);
```
olantwin: If we also handle the case of a single file via TChain, wouldn't it be better to make `fTree` a `TChain *`? Then casts like this would be unnecessary. We should probably also rename it to avoid confusion.

PR #1030, opened 2026-02-10
feat: set UseGeneralProcess to false only when cross-sections need to be changed

THanae:
`UseGeneralProcess` is true by default (which gives a slight speedup for cases where the cross-sections are not changed), and is only set to false when cross-sections need to be changed. This is done by setting an environment variable.
Two scripts which change the cross sections are updated. 

## Checklist

- [ ] CI checks pass

### DIFF
--- a/gconfig/g4Config.C
+++ b/gconfig/g4Config.C
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // SPDX-FileCopyrightText: Copyright CERN for the benefit of the SHiP
 // Collaboration
+#include "G4UImanager.hh"
 
 // Configuration macro for Geant4 VirtualMC
 void Config() {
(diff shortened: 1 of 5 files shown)


gconfig/g4Config.C
```diff
@@ -58,4 +59,15 @@ void Config() {
   // still stupid bug in geant4_vmc
   // geant4->SetMaxNStep(10000.);  // default is 30000
   geant4->ProcessGeantMacro(configm1.Data());
+
+  const char* set_general_process_to_false =
```
olantwin: Can we use a string already here?
You might want to look whether there is a way in the C++ standard library to get environment variables.

THanae: I updated it to use a string (note that a string cannot be instantiated to NULL so now if the environment variable is not set, `set_general_process_to_false` is set to `""`)

PR #1033, opened 2026-02-11
fix: update number of muons passed to MuonBackgen

THanae:
If a too large number of events is passed to `run.Run` using the `MuonBackGen` generator, this results in a segfault. This is because only muons are considered. Suggest to update the code to run over the minimum between the number of events passed as an option, and the number of muons in `PlaneHAPoint`.

## Checklist

- [ ] CI checks pass


olantwin:
Closing in favour of #1040

PR #1034, opened 2026-02-11
Clamp random seed to PYTHIA8 maximum allowed value

olantwin:
PYTHIA8 requires Random:seed to be in range [0, 900000000] (see https://pythia.org/latest-manual/RandomNumberSeed.html). When seed is set to 0, ROOT generates a time-based seed which can exceed this limit, causing PYTHIA8 to throw an out-of-range error.

This commit adds seed clamping logic to ensure automatically-generated seeds fall within PYTHIA8's valid range whilst preserving randomness.

## Checklist

- [x] CI checks pass


olantwin:
FYI @THanae, relevant for background production

THanae (commented):
Could there also be issues with a too large seed in `muonShieldOptimization/study_muMSC.py`, `muonShieldOptimization/study_GammaConv.py` and `macro/runPythia8.py`?

PR #1047, opened 2026-02-13
Make a base class for the generators in shipgen

mesmith75:
## Checklist

- [x] CI checks pass

### DIFF
--- a/shipgen/GenieGenerator.h
+++ b/shipgen/GenieGenerator.h
@@ -5,9 +5,8 @@
 #ifndef SHIPGEN_GENIEGENERATOR_H_
 #define SHIPGEN_GENIEGENERATOR_H_ 1
 
-#include "FairGenerator.h"
 #include "FairLogger.h"  // for FairLogger, MESSAGE_ORIGIN
-#include "ShipGenerator.h"
+#include "Generator.h"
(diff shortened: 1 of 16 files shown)


shipgen/GenieGenerator.h
olantwin: I think this might need to be ClassDefOverride (maybe also the other classes)

shipgen/ShipGenerator.h
```diff
@@ -0,0 +1,40 @@
...
+class ShipGenerator : public FairGenerator {
+ public:
+  ShipGenerator() {};
+  virtual ~ShipGenerator();
+
+  virtual Bool_t Init(const char*, int) = 0;
+  virtual Bool_t Init(const char*) = 0;
+  virtual Bool_t Init(const std::vector<std::string>& inFiles,
+                      int startNumber) {
+    return Init(inFiles.at(0).c_str(), startNumber);
```
olantwin: It would be good to log a warning here if a file was dropped. Otherwise, the user might assume that all files were processed.

shipgen/GenieGenerator.h
mesmith75: Aha! that was the bit of knowledge I was missing

mesmith75:
By all means, have at it

PR #1099, opened 2026-03-12
Off-by-one in MuonBackGenerator end-of-tree check

olantwin:
When a muon is found on the last event, fn equals fNevents after the while loop increments it. The equality check incorrectly returned kFALSE despite having found a valid muon. Track whether a muon was found separately from the event counter.

Supersedes #1098 ?

## Checklist

- [ ] CI checks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ it in future.
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


olantwin:
The behaviour of the other generators is to skip empty events. I think MuonBack should also return min(`-n`, fNevents - N_{event without muons}) events with at least one muon.

PR #1140, opened 2026-04-02
Add run_fixedTarget summary

mesmith75:
Probably missing a few things, but you get the idea

## Checklist

- [x] CI checks pass

### DIFF
--- a/macro/run_fixedTarget.py
+++ b/macro/run_fixedTarget.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # SPDX-FileCopyrightText: Copyright CERN for the benefit of the SHiP Collaboration
 
+import json
 import os
 
 import geometry_config
@@ -9,7 +10,6 @@ import ROOT
 import shipRoot_conf
 import shipunit as u
 import uproot
-import json
 
 mcEngine = "TGeant4"
 simEngine = "Pythia8"
--- a/macro/run_fixedTarget.py
+++ b/macro/run_fixedTarget.py
@@ -505,7 +505,7 @@ fsr = {
     "script": "run_fixedTarget.py",
     "runNumber": args.runnr,
     "seed": seed,
-    "PoT":  args.nev,
+    "PoT": args.nev,
     "boostDiMuon": args.boostDiMuon,
     "boostFactor": args.boostFactor,
 }
--- a/macro/run_fixedTarget.py
+++ b/macro/run_fixedTarget.py
@@ -9,6 +9,7 @@ import ROOT
 import shipRoot_conf
 import shipunit as u
 import uproot
+import json
 
 mcEngine = "TGeant4"
 simEngine = "Pythia8"
@@ -511,7 +512,7 @@ fsr = {
 }
 
 with uproot.update(outFile) as _of:
-    _of["FileSummary"] = fsr
+    _of["FileSummary"] = json.dumps(fsr)
 
 fin.SetWritable(False)  # bpyass flush error
 
--- a/macro/run_fixedTarget.py
+++ b/macro/run_fixedTarget.py
@@ -505,9 +505,9 @@ fsr = {
     "script": "run_fixedTarget.py",
     "runNumber": args.runnr,
     "seed": seed,
-    "PoT":  args.nev
-    "boostDiMuon", args.boostDiMuon,
-    "boostFactor", args.boostFactor,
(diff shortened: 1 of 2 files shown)


macro/run_fixedTarget.py
```diff
@@ -500,6 +501,18 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
 rc1 = os.system("rm  " + outFile)
 rc2 = os.system("mv " + tmpFile + " " + outFile)
 print("removed out file, moved tmpFile to out file", rc1, rc2)
+fsr = {
```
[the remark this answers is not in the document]
THanae: If we want the file to be fully reproducible without looking back at the submission scripts, we might need to report on most of the arguments passed to the script. So in addition to what you already have, the following (I added some extra information on the arguments so we can decide if the list is too exhaustive -- it is likely that some of these arguments will not be changed for large central productions)
```
{
# this could be found back by looking inside the file / getting the file name, but might be nice to have in the file info in addition
"ecut": args.ecut,
# gives information about which generators/decayers are used, which in turn helps understand potential differences between files
"G4only": args.G4only,
"pythiaDecay": args.pythiaDecay,
# lets us know if s decays only to taus
"tauOnly": args.tauOnly,
# lets us know if onia production flag is set to "on"
"JpsiMainly": args.JpsiMainly,
# helps understand why some particles look to be "missing" in the file
"storeOnlyMuons": args.storeOnlyMuons,  
"skipNeutrinos": args.skipNeutrinos, 
# whether extra NTuple is created
"FourDP": args.FourDP,
# tells us if we are dealing with charm or beauty production
"charm": args.charm,
"beauty", args.beauty,
# cross-sections of ccbar or bbar over minbias
"chicc":args.chicc,
"chibb":args.chibb,
# input file for charm and beauty
"charmInputFile": args.charmInputFile,
# number of starting event (when not 0)
"nStart": args.nStart,
# information relating to shield, decay volume, and what exactly is simulated in the target production
"DecayVolumeMedium": args.DecayVolumeMedium,
"shieldName": args.shieldName,
"AddMuonShield": args.AddMuonShield,
"AddMuonShieldField": args.AddMuonShieldField,
"AddHadronAbsorberOnly": args.AddHadronAbsorberOnly,
# offsets and beam smearing/painting -- usually default values but could be changed
"x_offset": args.x_offset,
"y_offset": args.y_offset,
"z_offset": args.z_offset,
"beam_smear": args.beam_smear,
"beam_paint": args.beam_paint,
}
```
There are some arguments which I do not think are worth adding, such as the PoT to normalize over for the charm/beauty case (which should always be 4E13), the targetyaml (which should only change for specific side productions, or when the design changes), the sensitive planes added (by default it is only the post-HA plane and more planes would only be added in specific side-studies), the output file, etc.

macro/run_fixedTarget.py
```diff
@@ -500,6 +502,18 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
...
+    "runNumber": args.runnr,
+    "seed": seed,
+    "PoT": args.nev,
+    "boostDiMuon": args.boostDiMuon,
+    "boostFactor": args.boostFactor,
+}
+
+with uproot.update(outFile) as _of:
+    _of["FileSummary"] = json.dumps(fsr)
+
```
olantwin: It might be better to use ROOT, as we don't depend on uproot elsewhere in FairShip/

```python
with ROOT.TFile.Open(outFile, "UPDATE") as _of:
    _of.WriteObject(ROOT.TObjString(json.dumps(fsr)), "FileSummary")
```

macro/run_fixedTarget.py
```diff
@@ -500,6 +502,18 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
 rc1 = os.system("rm  " + outFile)
 rc2 = os.system("mv " + tmpFile + " " + outFile)
 print("removed out file, moved tmpFile to out file", rc1, rc2)
```
olantwin: This is pre-existing, but we might want to check the return codes before reopening and writing to the file.

olantwin (commented):
We should be able to automatically parse the command line and add command name as well as all arguments, or get these from argparse.

I'll take care of the commit messages when squashing. Could you add a change log entry?

mesmith75:
Yes, I think ultimately that will be the aim. For now, shall we proceed with this for the the productions?

THanae (approved):
Looks good to me as well.

PR #1153, opened 2026-04-22
feat: add angular acceptance cut for HNL and DP generators

olantwin:
Skip events where the HNL/DP trajectory angle exceeds the decay vessel acceptance. The maximum angles in x and y are derived from the vessel geometry with a 10 cm margin. Events outside acceptance trigger a retry with a new Pythia event, reusing the existing retry loop.

@amagnan @THanae I'm not sure whether we need to account for this in the statistics.

## Checklist

- [x] CI checks pass

## Summary by CodeRabbit

* **New Features**
  * Generators now enforce angular/slope acceptance and only produce candidates that can reach the decay vessel.

* **Changed**
  * Retry/end-of-run reporting now separates and reports geometric rejections (outside vessel acceptance).
  * Decay-vessel end dimensions are now read from veto configuration (may change effective vessel extents).

* **Documentation**
  * CHANGELOG updated with these Unreleased entries.

### DIFF
--- a/macro/run_simScript.py
+++ b/macro/run_simScript.py
@@ -459,11 +459,14 @@ if options.pythia8:
         P8gen.SetPaintRadius(options.PaintBeam * u.cm)  # beam painting radius
         P8gen.SetLmin((ship_geo.Chamber1.z - ship_geo.chambers.Tub1length) - ship_geo.target.z0)
         P8gen.SetLmax(ship_geo.TrackStation1.z - ship_geo.target.z0)
-        margin = 10 * u.cm
-        z_end = ship_geo.decayVolume.z0 + ship_geo.decayVolume.length
+        margin = 10 * u.cm  # covers beam smearing + non-projectivity of vessel
+        z_end = ship_geo.decayVolume.z0 + ship_geo.decayVolume.length  # vessel exit z from target
+        # xEndInner/yEndInner are full inner widths; divide by 2 for half-aperture.
+        # Assumes production vertex at target (z=0). Vessel is approximately
+        # projective, so a single angle cut at the exit (widest point) suffices.
         max_theta_x = (ship_geo.decayVolume.xEndInner / 2 + margin) / z_end
         max_theta_y = (ship_geo.decayVolume.yEndInner / 2 + margin) / z_end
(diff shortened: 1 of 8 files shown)


python/geometry_config.py
```diff
@@ -282,6 +282,9 @@ def create_config(
     c.z = 89.57 * u.m  # absolute position of spectrometer magnet
     c.decayVolume.z = c.z - 31.450 * u.m  # Relative position of decay vessel centre to spectrometer magnet
     c.decayVolume.z0 = c.decayVolume.z - c.decayVolume.length / 2.0
+    # Full inner widths at vessel exit (cm), must match veto_config_{helium,vacuums}.yaml
+    c.decayVolume.xEndInner = 400 * u.cm
+    c.decayVolume.yEndInner = 600 * u.cm
```
THanae: Could it be a problem if the veto_config changes later on? Would there be a way to read the yaml file directly?

macro/run_simScript.py
```diff
@@ -459,6 +459,14 @@
         P8gen.SetPaintRadius(options.PaintBeam * u.cm)  # beam painting radius
         P8gen.SetLmin((ship_geo.Chamber1.z - ship_geo.chambers.Tub1length) - ship_geo.target.z0)
         P8gen.SetLmax(ship_geo.TrackStation1.z - ship_geo.target.z0)
+        margin = 10 * u.cm  # covers beam smearing + non-projectivity of vessel
```
THanae, line 462: Could you explain why the smearing is accounted for (as opposed to a sharp cutoff at the values of `xEndInner` and `yEndInner`) + why exactly at 10cm (is it to cover a certain percentage)?

olantwin: Your right, we're not really accounting for the smearing here, rather the non-projectivity of the vessel. Since the vessel is not exactly a frustum, I want a bit of a margin outside (and the fiducial volume also has a margin on the inside).

python/geometry_config.py
olantwin: Good point, I'll fix this.

THanae (approved):

PR #1156, opened 2026-04-24
fix: use stepwise extrapolation in vertex finding

olantwin:
The iterative vertex finder and precision fit extrapolate tracks from the straw tracker (Z≈8400 cm) backwards to the decay vertex. 
For upstream vertices (Z≈3500–5500 cm), the single-step backward Runge-Kutta propagation throws numerical exceptions.

Use stepwise plane extrapolation (500 cm steps) in both the DOCA iteration loop and the precision fit stage of `TwoTrackVertex()`.

On an example HNL→µπ sample (2500 events, V21_3000 geometry), this recovers 68 additional vertices, improving vertexing efficiency from 66% to 85% of reconstructible events.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed upstream vertex finding — stepwise track extrapolation reduces fit failures and improves success rate for distant/long tracks.
* **Refactor**
  * Improved vertex-fitting workflow for more stable and precise vertex position and distance-of-closest-approach estimates, yielding more reliable reconstruction overall.

### DIFF
--- a/python/shipVertex.py
+++ b/python/shipVertex.py
@@ -82,6 +82,19 @@ class Task:
         self.chi2(res, self.Vy)
         return
 
+    def _stepwise_extrapolate_to_z(self, extrapolate_fn, current_z, target_z, max_step=500.0):
+        """Extrapolate in steps to avoid RK failures over large distances."""
+        dz_total = target_z - current_z
+        n_steps = max(1, math.ceil(abs(dz_total) / max_step))
+        for istep in range(n_steps):
+            step_z = current_z + dz_total * (istep + 1) / n_steps
+            plane = ROOT.genfit.DetPlane(
+                ROOT.TVector3(0, 0, step_z),
+                ROOT.TVector3(1, 0, 0),
+                ROOT.TVector3(0, 1, 0),
+            )
+            extrapolate_fn(ROOT.genfit.SharedPlanePtr(plane))
+
     def TwoTrackVertex(self) -> None:
         self.fPartArray.clear()
         fittedTracks = getattr(self.sTree, self.fitTrackLoc)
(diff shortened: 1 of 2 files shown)


olantwin:
@coderabbitai review

python/shipVertex.py
jasminwss, line 279: what is the purpose of this array?

jasminwss (approved):

python/shipVertex.py
olantwin: That seems to be just a noop. I don't see any function.

olantwin: to be removed

olantwin: https://github.com/ShipSoft/FairShip/pull/1159

wcmartylee (approved):
No complaints from me.

mborisyak (approved):
Does RK mean Runge-Kutta? I am not familiar with the inner workings of genfit, but from general theory, wouldn't this iteration have the same effect as setting tolerance to a higher value? Wouldn't it be better to use smaller step size, for example, in case of a failure, try again with a smaller step size? Also, it is somewhat bizarre that RK fails on long extrapolations since trajectories are practically linear outside the tracker (right?) and RK perfectly extrapolates these (aside from numerical errors). Am I missing something?

olantwin:
> Does RK mean Runge-Kutta? I am not familiar with the inner workings of genfit, but from general theory, wouldn't this iteration have the same effect as setting tolerance to a higher value? Wouldn't it be better to use smaller step size, for example, in case of a failure, try again with a smaller step size? Also, it is somewhat bizarre that RK fails on long extrapolations since trajectories are practically linear outside the tracker (right?) and RK perfectly extrapolates these (aside from numerical errors). Am I missing something?

An improvement would indeed doing this adaptively. Maybe this is also a bug in the GenFit implementation. For now this is a workaround, which can definitely be cleaned up further.

PR #1157, opened 2026-04-24
fix: determine track charge from bending and dual-hypothesis fitting

olantwin:
Two changes to improve charge determination in track fitting:

1. **Slope-based charge seeding**: Instead of hardcoding PDG 13 (mu-) for all tracks, use the y-slope difference between stations 1-2 and 3-4 to determine the charge sign from the bending direction. This gives the Kalman fitter a correct initial hypothesis.

2. **Dual-hypothesis fitting**: Fit each track with both charge hypotheses (PDG 13 and -13) and keep the one with better chi2/NDF. This resolves charge ambiguity for tracks where the slope-based determination is unreliable.

On an example HNL→µπ sample (2500 events, V21_3000 geometry), this reduces charge mis-ID from 48 to 19 events. Combined with ShipSoft/FairShip#1156, overall vertexing efficiency improves from 66% to 93% of reconstructible events.

## Summary by CodeRabbit

* **Improvements**
  * Track fitting now evaluates both charge-sign hypotheses for each candidate and automatically selects the best converged fit by chi‑squared per degree of freedom.
  * Removed redundant per-track consistency checks, simplifying the post-fit processing pipeline.
  * Diagnostic logging updated to reflect the new fitting flow and streamlined status information.

### DIFF
--- a/python/shipDigiReco.py
+++ b/python/shipDigiReco.py
@@ -264,6 +264,7 @@ class ShipDigiReco:
                     theTrack.checkConsistency()
                 except ROOT.genfit.Exception:
                     logger.debug("Track inconsistent after fit for hypothesis %d", try_pdg)
+                    continue
                 try:
                     fittedState = theTrack.getFittedState()
                     fittedState.getMomMag()
--- a/python/shipDigiReco.py
+++ b/python/shipDigiReco.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # SPDX-FileCopyrightText: Copyright CERN for the benefit of the SHiP Collaboration
 
+import contextlib
 import logging
 from array import array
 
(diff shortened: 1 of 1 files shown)


olantwin:
@coderabbitai review

mborisyak (approved):
Everything seems reasonable.

olantwin:
@coderabbitai review

wcmartylee (commented):
Since we’re already fitting both charges in the for loop and getting the best fit, do we still need to first determine the charge from the slopes in the beginning?

With this, theoretically, we can even try to fit other hypotheses than muons.

olantwin:
> Since we’re already fitting both charges in the for loop and getting the best fit, do we still need to first determine the charge from the slopes in the beginning?
> 
> With this, theoretically, we can even try to fit other hypotheses than muons.

The idea would be to only fit both hypotheses if we are unsure from the bending direction. This can be optimized later, I'm focussing on fixing the most obvious issues right now.

wcmartylee:
> The idea would be to only fit both hypotheses if we are unsure from the bending direction.

So simply as a safe-guard check?

olantwin:
> > The idea would be to only fit both hypotheses if we are unsure from the bending direction.
> 
> So simply as a safe-guard check?

I'm not sure what you mean.

Each method has a limited efficiency. Fitting both hypotheses adds a lot of time, but should be more reliable than just looking at the curvature from tracking.

olantwin:
@coderabbitai review

wcmartylee (approved):

PR #1181, opened 2026-05-13
Fix run_simScript.py for new production

mesmith75:
Picks up if it is a new file and sets the target location appropriately.
## Checklist

- [ ] CI checks pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved simulation target offset selection by detecting input file structure more reliably, preventing incorrect offsets.

* **Chores**
  * Added an internal utility to validate input file structure and ensure consistent branch presence across inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ShipSoft/FairShip/pull/1181)

### DIFF
--- a/python/rootUtils.py
+++ b/python/rootUtils.py
@@ -177,7 +177,7 @@ def checkForBranch(x, branchName) -> bool:
     else:
         tx = x
     if isinstance(tx, (list, tuple)):
-        hasIt = None 
+        hasIt = None
         for _f in tx:
             if _f[0:4] == "/eos":
                 f = ROOT.gSystem.Getenv("EOSSHIP") + _f
@@ -187,17 +187,11 @@ def checkForBranch(x, branchName) -> bool:
             if not test:
                 print("ERROR FileCheck: input file", f, " does not exist. Missing authentication?")
                 sys.exit(1)
-            current_has_branch = bool(
-                test.FindObjectAny("cbmsim") and test["cbmsim"].FindBranch(branchName)
-            )
+            current_has_branch = bool(test.FindObjectAny("cbmsim") and test["cbmsim"].FindBranch(branchName))
             if hasIt is None:
                 hasIt = current_has_branch
             elif current_has_branch != hasIt:
-                print(
-                    "ERROR CheckForBranch: Mixed input files (some have branch",
-                    branchName,
-                    "and some do not)."
-                )
+                print("ERROR CheckForBranch: Mixed input files (some have branch", branchName, "and some do not).")
                 sys.exit(1)
         return bool(hasIt)
     else:
--- a/python/rootUtils.py
+++ b/python/rootUtils.py
@@ -177,8 +177,7 @@ def checkForBranch(x, branchName) -> bool:
     else:
         tx = x
     if isinstance(tx, (list, tuple)):
-        # See what we are looking at and make sure all the files are of the same type
-        hasIt = False
+        hasIt = None 
         for _f in tx:
             if _f[0:4] == "/eos":
                 f = ROOT.gSystem.Getenv("EOSSHIP") + _f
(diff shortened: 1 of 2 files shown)


python/rootUtils.py
```diff
@@ -169,3 +170,34 @@ def checkFileExists(x) -> str:
...
+                print("ERROR FileCheck: input file", f, " does not exist. Missing authentication?")
+                sys.exit(1)
+            if test.FindObjectAny("cbmsim"):
+                if test["cbmsim"].FindBranch(branchName):
+                    hasIt = True
+                else:
+                    return False
+            else:
+                return False
+        return hasIt
```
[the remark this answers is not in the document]
mesmith75: fixed

python/rootUtils.py
```diff
@@ -169,3 +169,31 @@ def checkFileExists(x) -> str:
...
+        tx = [x]
+    else:
+        tx = x
+    if isinstance(tx, (list, tuple)):
+        hasIt = None
+        for _f in tx:
+            if _f[0:4] == "/eos":
+                f = ROOT.gSystem.Getenv("EOSSHIP") + _f
+            else:
+                f = _f
```
olantwin: I'd prefer we remove this special case. While it adds convenience, it makes e.g. using files from CERNBox impossible.

If on lxplus, specifying the short path will still work, as ROOT upgrades from FUSE to XRootD automatically, on all other systems we should require full paths.

python/rootUtils.py
```diff
@@ -169,3 +169,31 @@ def checkFileExists(x) -> str:
...
+            test = ROOT.TFile.Open(f)
+            if not test:
+                print("ERROR FileCheck: input file", f, " does not exist. Missing authentication?")
+                sys.exit(1)
```
olantwin: Maybe better to raise an exception here that can be caught? Depending on context, it might not be necessary to exit the programme if an input file doesn't exist.

python/rootUtils.py
```diff
@@ -169,3 +169,31 @@ def checkFileExists(x) -> str:
     else:
         print("ERROR FileCheck: File must be either a string or list of files")
         os._exit(1)
+
+
+def checkForBranch(x, branchName) -> bool:
```
olantwin: Can we use a better variable name than x here? Within the function short names are fine, it's not a complex function, but it would be good to have a self-explanatory variable name in the signature, especially in absence of a docstring.

python/rootUtils.py
```diff
@@ -169,3 +169,31 @@ def checkFileExists(x) -> str:
...
+            test = ROOT.TFile.Open(f)
+            if not test:
+                print("ERROR FileCheck: input file", f, " does not exist. Missing authentication?")
+                sys.exit(1)
+            current_has_branch = bool(test.FindObjectAny("cbmsim") and test["cbmsim"].FindBranch(branchName))
+            if hasIt is None:
+                hasIt = current_has_branch
+            elif current_has_branch != hasIt:
+                print("ERROR CheckForBranch: Mixed input files (some have branch", branchName, "and some do not).")
+                sys.exit(1)
```
olantwin: Here exiting makes more sense, but an exception would allow us to communicate more clearly what the issue is.

olantwin (changes requested):
Looks like a useful improvement! Some minor nitpicks (I know these equally apply to most of the functions in this module, I haven't gotten around to cleaning them up yet...)

python/rootUtils.py
```diff
@@ -169,3 +169,30 @@ def checkFileExists(x) -> str:
...
+
+def checkForBranch(inFile, branchName) -> bool:
+    if isinstance(inFile, str):
+        fileList = [inFile]
+    else:
+        fileList = inFile
+    if isinstance(fileList, (list, tuple)):
+        hasIt = None
+        for _f in fileList:
+            test = ROOT.TFile.Open(_f)
```
olantwin:
```suggestion
            with ROOT.TFile.Open(_f) as test:
```

This would handle any possible exit from the function so that the file is always closed. Just in case you were unaware of this pythonisation, I don't think this blocks the merge.

olantwin (dismissed):

olantwin (approved):
LGTM.

PR #1183, opened 2026-05-16
Add ParticleGunGenerator

mesmith75:
Adapted from FairBoxGenerator with some added functionality. 

- load an arbitrary set of histograms to generate variables from
- Choose to generate both +/- charges with a determined fraction
- Choose a model for the vertex and momentum generation
- Simply implement and choose a particular model for either if the existing is not what you want

## Checklist

- [ ] CI checks pass

## Summary by CodeRabbit

* **New Features**
  * New particle-gun generator with configurable kinematics, multiplicity and vertex placement.
  * Repeatable histogram inputs for 1D/2D/3D overrides with per-histogram variable mapping and list-length validation.
  * Selectable vertex smearing modes.
  * Dual-charge generation with validated charge-fraction (0.0–1.0) and configurable fraction for opposite-charge sampling.
  * Expanded CLI to control histograms, variable mapping, smearing mode, and charge behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ShipSoft/FairShip/pull/1183?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


shipgen/ParticleGunGenerator.cxx
```diff
@@ -0,0 +1,315 @@
...
+      break;
+    case 2:
+      static_cast<TH2*>(fDistHist.get())
+          ->GetRandom2(GetVar(p, fHistoVars[0]), GetVar(p, fHistoVars[1]));
+      break;
+    case 3:
+      static_cast<TH3*>(fDistHist.get())
+          ->GetRandom3(GetVar(p, fHistoVars[0]), GetVar(p, fHistoVars[1]),
+                       GetVar(p, fHistoVars[2]));
+      break;
```
THanae: Currently if I understand correctly there are 3 possible dimensions -- given the code mostly adapted for momenta. Would it be possible to have an option to also override position, potentially at the same time as momentum?

shipgen/ParticleGunGenerator.cxx
```diff
@@ -0,0 +1,315 @@
...
+      return gRandom->Uniform(centre - spread, centre + spread);
+
+    case SmearMode::kExponential:
+    default: {
+      const Double_t sign = (gRandom->Uniform() < 0.5) ? +1. : -1.;
+      return centre + sign * gRandom->Exp(spread);
+    }
+  }
+}
+
```
THanae: Similar question here: could there be a basic `SmearMomentum`, to allow for some shifts / favour high momenta? Or do we think that in such cases, a proper representation with a user-defined histogram is the best approach?

macro/run_simScript.py
```diff
@@ -60,10 +68,32 @@
...
+    default=False,
+    action="store_true",
+    help="Generate both charges of given particle ID",
+)
+pg_parser.add_argument(
+    "--chargeFraction",
+    dest="chargeFraction",
+    type=_fraction_0_1,
+    default=0.5,
+    help="Fraction of chosen charge to generate [0,1]",
```
THanae: It might be good to also add which charge corresponds to the fraction.

shipgen/ParticleGunGenerator.h
```diff
@@ -0,0 +1,171 @@
...
+  void SetVertex(Double32_t x = 0, Double32_t y = 0, Double32_t z = 0,
+                 Double32_t ex = 0, Double32_t ey = 0, Double32_t ez = 0);
+
+  void SetBoxXYZ(Double32_t x1 = 0, Double32_t y1 = 0, Double32_t x2 = 0,
+                 Double32_t y2 = 0, Double32_t z = 0);
+
+  void SetBothCharges(bool flag, Double32_t fraction = 0.5);
+
+  void SetPDGType(int pdgid) { m_pdgid = pdgid; };
+  void SetMultiplicity(int mult) { m_mult = mult; };
```
THanae, line 85: Is this multiplicity used anywhere / planned to be used anywhere?

shipgen/ParticleGunGenerator.cxx
```diff
@@ -0,0 +1,315 @@
...
+#include <TDatabasePDG.h>
+#include <TMath.h>
+#include <TNamed.h>
+#include <TParticlePDG.h>
+#include <TRandom.h>
+
+#include <cmath>   // for cos, acos
+#include <cstdio>  // for printf
+
+#include "FairLogger.h"
```
THanae, line 14: A few months ago there was a discussion about moving away from `FairLogger`, but I will let Oliver comment, given that at some point we will be moving to the new framework.

shipgen/ParticleGunGenerator.cxx
mesmith75: No, this is for any 3 variables. Unfortunately ROOT histograms stop at 3D. If you think it worthwhile we can introduce a second histogram

macro/run_simScript.py
mesmith75: It says fraction of chosen charge, i.e. the one you specify in the particle ID. I'll make the help a bit clearer

shipgen/ParticleGunGenerator.h
mesmith75: Added the option, although I am not sure anything other than 1 does much for you

shipgen/ParticleGunGenerator.cxx
mesmith75: Added arbitrary numbers of histograms

shipgen/ParticleGunGenerator.cxx
mesmith75: I'll add in something reasonable for the momenta

shipgen/ParticleGunGenerator.h
```diff
@@ -0,0 +1,201 @@
...
+enum class SmearMode { kExponential, kGaussian, kUniform };
+
+struct MomentumModelSpec {
+  int expectedPars;
+  const char* description;
+  std::function<void(ParticleGunParticle&, const std::vector<Double32_t>&)>
+      generate;
+};
+
+static const std::map<int, MomentumModelSpec> kMomentumModels = {
```
mesmith75: @THanae This is now fairly generic. If you want to add new models for the momentum generation they only need adding here. The parameters are passed as arguments to `run_simScript.py` completely generically.

shipgen/ParticleGunGenerator.cxx
olantwin: I think for FairShip I'd keep using FairLogger for consistency.. Maybe at some point I'll do a big search/replace in FairShip.

New framework uses spdlog.

olantwin:
@coderabbitai review

macro/run_simScript.py
```diff
@@ -60,16 +68,86 @@
...
 pg_parser.add_argument("--Vx", dest="Vx", default=0, type=float, help="x position of particle gun (default=0 cm)")
 pg_parser.add_argument("--Vy", dest="Vy", default=0, type=float, help="y position of particle gun (default=0 cm)")
 pg_parser.add_argument("--Vz", dest="Vz", default=0, type=float, help="z position of particle gun (default=0 cm)")
+pg_parser.add_argument(
+    "--smearMode",
+    dest="smearMode",
+    default="uniform",
+    help="Form of the vertex smearing for the particle gun",
+    type=str,
+)
```
olantwin, line 131: Could we list the allowed options as `choices=` here (so that they're also validated?)

macro/run_simScript.py
```diff
@@ -60,16 +68,86 @@
...
 pg_parser.add_argument(
     "--Dx", dest="Dx", type=float, help="size of the full uniform spread of PG xpos: (Vx - Dx/2, Vx + Dx/2)"
 )
 pg_parser.add_argument(
     "--Dy", dest="Dy", type=float, help="size of the full uniform spread of PG ypos: (Vy - Dy/2, Vy + Dy/2)"
 )
+pg_parser.add_argument(
+    "--momentumModel", dest="momentumModel", default=0, type=int, help="Select a model to generate the momentum with"
+)
```
olantwin, line 143: Can we specify the choices and validate them here using `choices`? Since this is an int, it would also be good to mention what the integers mean.

shipgen/ParticleGunGenerator.cxx
```diff
@@ -0,0 +1,364 @@
...
+
+  // Check for variable name conflicts with already-loaded histograms
+  for (const auto& name : varNames) {
+    for (const auto& entry : fHistoEntries) {
+      for (const auto& existing : entry.varNames) {
+        if (existing == name) {
+          LOG(fatal)
+              << "ParticleGunGenerator: variable \"" << name
+              << "\" is already controlled by a previously loaded histogram";
+          throw std::runtime_error(
```
olantwin, line 113: I think LOG(fatal) will automatically raise an exeption.

shipgen/ParticleGunGenerator.h
```diff
@@ -0,0 +1,214 @@
...
+      [](ParticleGunParticle& p, const std::vector<Double32_t>& pars) {
+        p.X = gRandom->Gaus(pars[0], pars[1]);
+        p.Y = gRandom->Gaus(pars[2], pars[3]);
+        p.Z = pars[4];
+      }}},
+    {"exponential",
+     {5, "Exp(X), Exp(Y), Point(Z)",
+      [](ParticleGunParticle& p, const std::vector<Double32_t>& pars) {
+        p.X = gRandom->Gaus(pars[0], pars[1]);
+        p.Y = gRandom->Gaus(pars[2], pars[3]);
```
olantwin: These are the same as in the Gaussian case. Should this be Exp?

shipgen/ParticleGunGenerator.h
```diff
@@ -0,0 +1,214 @@
...
+        p.Z = pars[4];
+      }}},
+    {"uniform",
+     {5, "Uniform(X), Uniform(Y), Point(Z)",
+      [](ParticleGunParticle& p, const std::vector<Double32_t>& pars) {
+        p.X = gRandom->Uniform(pars[0], pars[1] / 2.);
+        p.Y = gRandom->Uniform(pars[2], pars[3] / 2.);
```
olantwin: Should this be `Uniform(pars[0] - pars[1]/2., pars[0] + pars[1]/2.)` and similar for p.Y? AFAIK Uniform takes lower and upper bound?

macro/run_simScript.py
```diff
@@ -529,22 +607,44 @@
...
-            (options.Vy + options.Dy / 2) * u.cm,
-            options.Vz * u.cm,
+        myPgun.SetVertexModel(
+            options.smearMode,
+            [options.Vx * u.cm, options.Dx * u.cm, options.Vy * u.cm, options.Dy * u.cm, options.Vz * u.cm],
         )
     else:
         # point source
-        myPgun.SetXYZ(options.Vx * u.cm, options.Vy * u.cm, options.Vz * u.cm)
+        myPgun.SetVertexModel(options.smearMode, [options.Vx * u.cm, 0.0, options.Vy * u.cm, 0.0, options.Vz * u.cm])
```
olantwin, line 639: So the smearing only takes effect in the multiplePG case?

olantwin, line 639: Does the Uniform distribution behave properly when the same start and end point?

olantwin: Maybe we could skip smearing when it is zero?

macro/run_simScript.py
```diff
@@ -529,22 +607,44 @@
...
+        histo_files = options.histoFile or []
+        histo_names = options.histoName or []
+        histo_x = options.histoX or []
+        histo_y = options.histoY or []
+        histo_z = options.histoZ or []
+
+        if len(histo_files) != len(histo_names):
+            raise ValueError("Must specify one --histoName per --histoFile")
+        if len(histo_x) != len(histo_files):
+            raise ValueError("Must specify one --histoX per --histoFile")
```
olantwin, line 652: Should we check the other dimensions as well?

The interface seems very powerful, but some concrete examples would help to understand how multiple histograms would be used. Histograms would need to be sorted by dimensionality (descending)?

shipgen/ParticleGunGenerator.h
```diff
@@ -0,0 +1,214 @@
...
+
+static const std::map<int, ModelSpec> kMomentumModels = {
+    {1,
+     {6, "Gaus(Px), Gaus(Py), Landau(Pz)",
+      [](ParticleGunParticle& p, const std::vector<Double32_t>& pars) {
+        p.Px = gRandom->Gaus(pars[0], pars[1]);
+        p.Py = gRandom->Gaus(pars[2], pars[3]);
+        p.Pz = gRandom->Landau(pars[4], pars[5]);
+      }}},
+};
```
olantwin: It might be better to declare these `inline` or move them to the implementation file. Statically declared variables in headers can lead to bugs if included in multiple translation units.

shipgen/ParticleGunGenerator.cxx
```diff
@@ -0,0 +1,364 @@
...
+          << "ParticleGunGenerator: Setting to generate both charges, with "
+             "fraction of "
+          << m_pdgid << " set to " << m_chargeFraction;
+  } else
+    LOG(info) << "ParticleGunGenerator: Setting to only generate one charge "
+                 "for PDG id"
+              << m_pdgid;
+}
+
+int ParticleGunGenerator::GetPDGType() {
```
olantwin: This is a bit confusing as I'd assume a Getter to not change any member variables. Maybe add a separate method to generate the PDG in the bothCharges case?

olantwin (changes requested):
Thanks for this very useful new feature! I lost count of how many times this functionality was requested or mentioned in the past.

In addition to the inline comments:
- Could we make the getters const?
- Could you please add a changelog entry?

This generally seems nearly ready for merging.

shipgen/ParticleGunGenerator.cxx
mesmith75: It doesn't in my test

shipgen/ParticleGunGenerator.h
mesmith75: Fixed

macro/run_simScript.py
mesmith75: In principle you only need the X (i.e. a 1D histogram) - yes, you need to do it in descending dimensions. I wanted to avoid adding loads of command line options. If you prefer I can make it explicit

macro/run_simScript.py
mesmith75: That is how it was set up previously - I was surprised too.

shipgen/ParticleGunGenerator.h
mesmith75: Ah yeah

shipgen/ParticleGunGenerator.cxx
mesmith75: This doesn't change any member variables - it just returns it. I'll make it const and rename to be clear

macro/run_simScript.py
olantwin: I think it's fine, maybe we can add a few examples to the help text of the PG command. Balancing discoverability and exposing all the possibilities is tricky.

macro/run_simScript.py
olantwin: I always thought it already did, then even better that you're now adding it.
Maybe we can discuss at one of the next meetings which default behaviour would be least surprising.

shipgen/ParticleGunGenerator.cxx
olantwin: Maybe that's a change on the FairLogger side... Good to know!

shipgen/ParticleGunGenerator.cxx
olantwin: Ah yes, you're right, I misread this slightly. That repeated calls will give potentially different answers still makes a rename worth it, I think.

shipgen/ParticleGunGenerator.cxx
mesmith75: I'll check again to be sure

macro/run_simScript.py
mesmith75: Would rather not - you end up having to synchronise two bits of code that is easily forgotten if new models are added. I have added an option to print the available models, and the PG throws an exception in initialisation if you give an invalid model.

macro/run_simScript.py
mesmith75: Again, would rather not as that requires synchronisation across two bits of code

olantwin (dismissed):

olantwin:
If this is OK to merge from your side, I'll squash.

olantwin:
@coderabbitai review

olantwin (approved):

PR #1187, opened 2026-05-19
fix: guard meson-production chain export for fpbrem in DPPythia8Gener…

olantwin:
…ator

For proton-bremsstrahlung events, event[1] is the DP itself (not a beam nucleon), so unconditionally exporting event[1] and event[im] duplicated the DP and inserted the Pythia system entry (id=90). Wrap the nucleon/mother/DP export block in `if (!fpbrem && im > 0)` and add an else branch that exports only the DP as root. Also replace the hardcoded `dec_chain.begin() + 3` offset with a dynamic `prefix_len` so the daughter loop works correctly for both chain lengths.

## Checklist

- [x] CI checks pass


amagnan (approved):
looks all good to me

PR #1196, opened 2026-05-27
fix: remove unnecessary libG4clhep loads

olantwin:
## Summary
- Remove `ROOT.gSystem.Load("libG4clhep")` from `python/shipRoot_conf.py`
- Remove `ROOT.gSystem.Load("libG4clhep.so")` from `macro/eventDisplay.py`

`libG4clhep` is a Geant4-internal CLHEP wrapper that only exists when Geant4 builds its own bundled CLHEP. In conda-forge/pixi builds, Geant4 links against standalone `libCLHEP` instead, so `libG4clhep` doesn't exist. The explicit loads are unnecessary since dependent libraries (geant4vmc etc.) already resolve CLHEP symbols via the dynamic linker.

## Test plan
- [x] CI simulation test passes without the libG4clhep error

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved magnetic field initialization for simulation accuracy.
  * Updated physics simulation library loading to use more current components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ShipSoft/FairShip/pull/1196?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/macro/eventDisplay.py
+++ b/macro/eventDisplay.py
@@ -1249,7 +1249,6 @@ gEve = ROOT.gEve
 
 if hasattr(ShipGeo, "Bfield"):
     if hasattr(ShipGeo.Bfield, "fieldMap"):
-        ROOT.gSystem.Load("libG4clhep.so")
         ROOT.gSystem.Load("libgeant4vmc.so")
         import geomGeant4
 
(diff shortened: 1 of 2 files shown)


olantwin:
@coderabbitai review

PR #1213, opened 2026-06-08
style: annotate virtual overrides with override

olantwin:
## Summary

Silences 34 `-Winconsistent-missing-override` warnings emitted by rootcling/clang on the master CI build (run [27147752577](https://github.com/ShipSoft/FairShip/actions/runs/27147752577)) by adding `override` to virtual functions in 11 headers.

This is PR 1 of a small series addressing build warnings — packaged per warning category for easier review.

Files touched (one virtual function group per file):

- `Detector/DetectorPoint.h` — destructor
- `field/FairShipFields.h` — two `get(...) const` overloads
- `field/ShipFieldMaker.h` — `Construct()`
- `field/ShipFieldPar.h` — destructor
- `passive/ShipGeoCave.h` — destructor + 5 `FairGeoSet` overrides
- `passive/ShipPassiveContFact.h` — destructor + `createContainer`
- `shipgen/HNLPythia8Generator.h` — `PyTr1Rng::flat()` and `PyTr3Rng::flat()`
- `splitcal/splitcalContFact.h`, `strawtubes/strawtubesContFact.h`, `veto/vetoContFact.h` — destructor + `createContainer`
- `veto/vetoHitOnTrack.h` — `Print(...)`

Conventions used:
- Empty inline destructors (`~Foo() {}`) rewritten as `~Foo() override = default;` to match the style already in `HNLPythia8Generator.h`.
- Redundant `virtual` keywords on overrides dropped per `modernize-use-override`.

No behavioural change — `override` is a compile-time check only.

## Test plan

- [x] Build clean locally with `pixi run build` — 132/132 steps, zero `-Winconsistent-missing-override` warnings
- [x] `clang-format --dry-run --Werror` clean across the touched headers
- [ ] CI passes (build + run-fixed-target + run-tracking-benchmark + run-sim-chain on both 26.02 and 26.05 legs)

### DIFF
--- a/Detector/DetectorPoint.h
+++ b/Detector/DetectorPoint.h
@@ -14,7 +14,7 @@ namespace SHiP {
 class DetectorPoint : public FairMCPoint {
  public:
   DetectorPoint() = default;
-  ~DetectorPoint() = default;
+  ~DetectorPoint() override = default;
 
   DetectorPoint(Int_t eventID, Int_t trackID, Int_t detID, TVector3 pos,
                 TVector3 mom, Double_t tof, Double_t length, Double_t eLoss,
--- a/field/FairShipFields.h
+++ b/field/FairShipFields.h
@@ -27,9 +27,9 @@ class FairShipFields : public AbsBField {
   inline void setField(ShipCompField* gField) { gField_ = gField; }
 
   //! return value at position
-  TVector3 get(const TVector3& pos) const;
+  TVector3 get(const TVector3& pos) const override;
   void get(const double& posX, const double& posY, const double& posZ,
-           double& Bx, double& By, double& Bz) const;
+           double& Bx, double& By, double& Bz) const override;
 
  private:
   ShipCompField* gField_{nullptr};
--- a/field/ShipFieldMaker.h
+++ b/field/ShipFieldMaker.h
@@ -68,7 +68,7 @@ class ShipFieldMaker : public TG4VUserPostDetConstruction {
   };
 
   //! Set-up all local and regional fields and assign them to volumes
-  virtual void Construct();
+  void Construct() override;
 
   //! Read an input file to define fields and associated volumes
   /*!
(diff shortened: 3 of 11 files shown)


jasminwss (commented):
found no mistakes

olantwin:
> found no mistakes 

Thanks!

matclim:
Do we expect plots to be different at all? because some are.

olantwin:
> Do we expect plots to be different at all? because some are.

No, we wouldn't. Which ones?

matclim:
I caught:

* In the fixed target, e+ momentum are different and have different entry counts
* in the sim chain, IPs are slightly different between same-type version (also different entry counts)
* in the tracking benchmark, everything but the p_truth are slightly different (x resolution, momentum resolution..) but entry counts are consistent.

They are pretty close but not identical.

There are certainly more, I didn't look at all the plots, one by one

olantwin:
> I caught:
> 
> * In the fixed target, e+ momentum are different and have different entry counts
> * in the sim chain, IPs are slightly different between same-type version (also different entry counts)
> * in the tracking benchmark, everything but the p_truth are slightly different (x resolution, momentum resolution..) but entry counts are consistent.
> 
> There are certainly more, I didn't look at all the plots, one by one
> 

This PR almost certainly can't be the reason, will be fun to find out which PR introduced the differences.
Now I wish we had started using the automatic statistical comparisons...

olantwin:
@matclim To me this seems consistent with low stats and a slightly no-deterministic track-fit.

matclim:
@olantwin If we are satisfied with this at the track level I wouldn't worry, what was more interesting to me were the mismatching entry counts but perhaps this is normal?

olantwin:
> @olantwin If we are satisfied with this at the track level I wouldn't worry, what was more interesting to me were the mismatching entry counts but perhaps this is normal?

We don't run the CI with a fixed seed, so entry counts fluctuating is normal.
Contributions to improving the CI are very welcome, now that the overlap checks are faster, we could maybe add additional jobs or checks

PR #1227, opened 2026-06-12
fix(simulation): MuonBack shutdown segfault from FairRootFileSink double-free (#1226)

### LINKED ISSUE #1226: Segm. Fault in the end of run_simScript, which does not affect the physics results
Error message:
```
========================================================================
 *** Break *** segmentation violation
 Generating stack trace...
 *** Break *** segmentation violation
 Generating stack trace...
QThreadStorage: entry 1 destroyed before end of thread 0x5646ce6d00c0
QThreadStorage: entry 0 destroyed before end of thread 0x5646ce6d00c0
Hangup
```

Reproducible in container:
```
FROM fedora:41

ENV LANG=C.UTF-8
ENV LC_ALL=C.UTF-8

RUN dnf -y update && \
    dnf -y install \
      git git-lfs curl wget ca-certificates \
      cmake ninja-build make gcc gcc-c++ gcc-gfortran \
      python3 python3-pip python3-devel \
      which procps-ng findutils rsync diffutils patch \
      bzip2 bzip2-devel xz tar gzip unzip \
      libX11-devel libXext-devel libXft-devel libXinerama-devel \
      libXi-devel libXrandr-devel libXrender-devel libXcursor-devel \
      mesa-libGL-devel mesa-libGLU-devel freeglut-devel \
      openssl-devel zlib-devel \
      && dnf clean all

RUN curl -fsSL https://pixi.sh/install.sh | bash

ENV PATH="/root/.pixi/bin:${PATH}"

WORKDIR /workspace

CMD ["/bin/bash"]
```

commands sequence:
```
podman run -it --rm \
  --name fairship-test \
  fairship-fedora

git clone https://github.com/ShipSoft/FairShip

cd FairShip

pixi install

pixi run build

export QT_QPA_PLATFORM=offscreen

pixi run python macro/run_simScript.py \
  -n 100 \
  -i 100 \
  -f https://cernbox.cern.ch/remote.php/dav/public-files/vdwtXtgM5P2Z0S5/pythia8_Geant4_10.0_withCharmandBeauty0_mu.root \
  --remote-input \
  --tag muonback_io \
  --sameSeed 42 \
  --seed 42 \
  --MuonBack \
  --FollowMuon \
  --FastMuon
```

olantwin:
## Summary

- **Fix #1226**: `MuonBack` post-processing in `run_simScript.py` constructed `FairRootFileSink(fout)` with a Python-owned `TFile`. `FairRootFileSink` stores the file in `std::unique_ptr<TFile>`, so at interpreter shutdown both Python and the sink deleted the same pointer, producing the segfault printed right after `Simulation finished successfully`. Relinquishing Python ownership immediately after passing `fout` to the sink resolves the crash. The harmless `QThreadStorage` lines from Geant4's Qt teardown remain; they were a coincident symptom, not the cause.
- **MuonBackGenerator cleanup**: `MCTrack` / `vetoPoints` / `fTree` were left uninitialised by the default constructor while the destructor unconditionally `delete`d the legacy raw pointers. Initialise everything to `nullptr`, add `delete fTree` (also closes the leaked `TChain` and its remote file handles), and drop the dead `fInputFile` member + `CloseFile()` method left behind by the chain-based refactor. This is a latent bug for the `pythia8-Geant4` ntuple input path that wasn't triggered by the `cbmsim`-format reproducer.
- **CI gap fix**: the existing `ci-sim` only covers the Pythia8+EvtGen path. Add a tiny 64 KB `cbmsim` skim (Git LFS, plus the script used to produce it), a `ci-sim-muonback` pixi task, and a `run-muonback` GitHub Actions job so this code path is exercised on every PR.

## Test plan

- [x] `pixi run ci-sim-muonback` — local smoke against the committed skim
- [x] Exact `--remote-input` HTTPS command from issue #1226 (`-n 5 -i 100`, otherwise identical) — clean exit, no `*** Break ***`
- [x] `pixi run ci-sim` — regression check on the Pythia8+EvtGen path
- [x] `run-muonback` GitHub Actions job (will run on this PR)

Closes #1226

## Summary by CodeRabbit

* **Chores**
  * Added CI automation to run a muon-back smoke simulation and report its status in the overall workflow.
* **Tests**
  * Added a reproducible muon-back smoke test task, updated test data, and a small utility to produce trimmed test inputs.
* **Refactor**
  * Improved muon-back generator memory handling and simplified its public interface.
* **Bug Fixes**
  * Fixed temporary file ownership/cleanup in the simulation post-processing.

### DIFF
--- a/tests/data/skim_muonback_smoke.py
+++ b/tests/data/skim_muonback_smoke.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# SPDX-FileCopyrightText: Copyright CERN for the benefit of the SHiP Collaboration
+"""Create a tiny pythia8-Geant4 muon-back skim for the ci-sim-muonback smoke test.
+
+Reads the full muon-back background sample (via xrootd with a Kerberos token,
+or the public HTTPS link), copies the first N entries of the pythia8-Geant4
+TTree into tests/data/muonback_smoke.root, and exits. Re-run only if the
+sample format changes.
+"""
+
+import argparse
+from pathlib import Path
+
+import ROOT
+import shipRoot_conf
+
+shipRoot_conf.configure()
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--source",
+        help="Input file (HTTPS public link or root:// URL)",
+        required=True,
+    )
+    parser.add_argument(
+        "--output",
+        type=Path,
+        default=Path(__file__).with_name("muonback_smoke.root"),
+        help="Output skim file",
+    )
+    parser.add_argument(
+        "--entries",
+        type=int,
+        default=200,
+        help="Number of entries to copy from the pythia8-Geant4 tree",
+    )
+    args = parser.parse_args()
+
+    src = ROOT.TFile.Open(args.source, "READ")
+    if not src or src.IsZombie():
+        raise SystemExit(f"Failed to open source: {args.source}")
+    tree = src.Get("pythia8-Geant4") or src.Get("cbmsim")
+    if not tree:
+        raise SystemExit("Neither pythia8-Geant4 nor cbmsim tree found in source")
+
+    out = ROOT.TFile.Open(str(args.output), "RECREATE")
+    if not out or out.IsZombie():
+        raise SystemExit(f"Failed to open output: {args.output}")
+    skim = tree.CopyTree("", "", args.entries, 0)
+    skim.Write()
+    out.Close()
+    src.Close()
+    print(f"Wrote {args.entries} entries to {args.output} ({args.output.stat().st_size} B)")
+    return 0
+
+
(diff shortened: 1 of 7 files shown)


tests/data/skim_muonback_smoke.py
```diff
@@ -0,0 +1,61 @@
...
+import shipRoot_conf
+
+shipRoot_conf.configure()
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--source",
+        default="https://cernbox.cern.ch/remote.php/dav/public-files/vdwtXtgM5P2Z0S5/pythia8_Geant4_10.0_withCharmandBeauty0_mu.root",
```
kholoimov: Currently, this file is located on my CERNBOX. It's absolutely fine, but probably we should move it to some SHiP storage

olantwin: Actually, we should probably just remove it. Leftover from testing.

kholoimov (approved):

PR #1229, opened 2026-06-13
ci: scope clang-tidy to repo headers and pin its include path

olantwin:
## Summary

Cleans up the two reasons clang-tidy was producing noisy or broken PR annotations:

### 1. `HeaderFilterRegex: '.*'` is too broad

clang-tidy was reporting warnings against every header it transitively included — including everything under `.pixi/envs/*` (ROOT, Geant4VMC, conda toolchain). PR #1225's `static-analysis` check accumulated 18 annotations, *all* in headers we don't own (`vmc/TMCManagerStack.h`, `TMath.h`, `TRandom.h`).

Tighten to a negative lookahead that excludes anything under `.pixi`, `.direnv`, `build`, `cmake`:

```yaml
HeaderFilterRegex: '^(?!.*/(\.pixi|\.direnv|build|cmake)/).*\.(h|hpp|hxx)$'
```

clang-tidy's `<regex>` uses ECMAScript flavor, which supports lookahead.

### 2. `'stdarg.h'/'stddef.h' file not found`

Every translation unit emitted a `clang-diagnostic-error: 'stdarg.h' file not found` (also visible on PR #1225). Root cause: `clang-tools` (LLVM 22.1) doesn't ship its libclang builtin headers in a discoverable location, and the `compile_commands.json` records the conda-forge gcc driver, whose internal include path is normally injected implicitly. Without a compiler driver in the loop, clang-tidy couldn't find the toolchain's freestanding headers.

Fix in `pixi.toml` `[tasks.ci-clang-tidy]`: ask the gcc driver for its include directory and pass it as `-isystem`:

```toml
cmd = "bash -c 'set -o pipefail; GCC_INC=\$(x86_64-conda-linux-gnu-gcc -print-file-name=include); echo \$CPP_FILES | xargs -P \$(nproc) clang-tidy -p build/ --extra-arg-before=-isystem\$GCC_INC 2>&1 | python .github/scripts/annotate_build.py'"
```

`x86_64-conda-linux-gnu-gcc -print-file-name=include` resolves to `…/lib/gcc/x86_64-conda-linux-gnu/<ver>/include`, which is where the conda toolchain stores `stdarg.h`, `stddef.h`, etc.

## After

Running locally on splitcalCluster.cxx:

- before: 1 clang-diagnostic-error (stddef.h missing), 60 non-user-code warnings reported, 2 in-repo warnings reported
[...]

PR #1231, opened 2026-06-13
ci: drop annotations for diagnostics in third-party trees

olantwin:
## Summary

clang-tidy's `HeaderFilterRegex` (added in #1229) filters check-based warnings but **not** compiler diagnostics. Errors that clang's own frontend emits — like `use of undeclared identifier '__builtin_ia32_bsrsi'` against the conda gcc intrinsic headers (`ia32intrin.h`, `adxintrin.h`, `cmpccxaddintrin.h`, `lwpintrin.h`, `movdirintrin.h`, `cetintrin.h`) — still slip through. On the most recent master run ([27465304759](https://github.com/ShipSoft/FairShip/actions/runs/27465304759)) there were ~20 such failure annotations, all in `.pixi/envs/default/lib/gcc/.../include/`. They eat into the 50-annotation GitHub cap and obscure the real signal from in-tree code.

## Change

Filter at the annotation layer where we already have the resolved file path. In `.github/scripts/annotate_build.py`, drop any diagnostic whose file lies under `/.pixi/`, `/.direnv/`, `/build/` or `/cmake/`. clang-tidy still prints the original line to stderr for log fidelity; we just don't surface it as a PR/master annotation.

## Test

Smoke test with a synthetic mix of inputs:

```text
.pixi/.../ia32intrin.h:41:10: error: ... [clang-diagnostic-error]      → dropped
.pixi/.../TMath.h:287:20: note: ...                                    → dropped
Detector/DetectorPoint.cxx:11:45: warning: ... [perf-...]              → kept
build/foo/G__Bar.cxx:99:1: error: bogus generated-code error           → dropped
```

After the filter, `Annotation Summary` reports `Errors: 0, Warnings: 1, Notices: 0` and only the `DetectorPoint.cxx` annotation is emitted.

## Notes

This is PR 1 in a follow-up series cleaning up the lints surfaced by the now-functioning `static-analysis-push` job (per-category PRs to come for `modernize-use-equals-default`, `modernize-use-nullptr`, `performance-unnecessary-value-param`, …).

## Summary by CodeRabbit

* **Chores**
  * Build diagnostics from third-party and generated sources are now filtered out from compilation output.

### DIFF
--- a/.github/scripts/annotate_build.py
+++ b/.github/scripts/annotate_build.py
@@ -7,9 +7,10 @@
 import re
 import sys
 from pathlib import Path
+from typing import Any
 
 
-def parse_compiler_output(line):
+def parse_compiler_output(line: str) -> dict[str, Any] | None:
     """
     Parse GCC/Clang compiler output line.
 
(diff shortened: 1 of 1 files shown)


olantwin:
@coderabbitai review

PR #1241, opened 2026-06-13
perf: pass non-trivial parameters by const reference

olantwin:
## Summary

Sweeps the 41 `performance-unnecessary-value-param` findings on master run [27465304759](https://github.com/ShipSoft/FairShip/actions/runs/27465304759): `TVector3`, `TString` and `std::string` parameters that the function body only reads, changed from pass-by-value to pass-by-const-reference.

## Approach

`clang-tidy --fix --checks='-*,performance-unnecessary-value-param'` applied 24 sites in 13 translation units. The FIX-IT propagated into 12 headers (`Detector/DetectorPoint.h`, `SND/*/`*Point.h`, generator helpers in `shipgen/`, `passive/ShipMuonShield.h`, `strawtubes/strawtubesPoint.h`, `veto/veto.h`).

## Safety net

Every TObject-derived virtual is now marked `override` (PR #1213 + the `splitcalCluster` follow-up), so any signature mismatch between an override and its base would have failed the build. The build still passes 88/88 incremental, and the existing `ctest` suite (`DataClassIO`, `RNtupleIO`) is green.

## Test plan

- [x] `pixi run --locked build` clean
- [x] `pixi run --locked test` — 2/2 tests pass
- [x] `CPP_FILES=<all-touched> pixi run --locked ci-clang-tidy` reports 0 `performance-unnecessary-value-param` findings

## Risks / notes

- A few of the touched functions are virtual overrides; if any FairROOT base has a `T param` signature we now pass `const T&`, the build would fail to find the override match. None did.
- No behavioural change for callers — these parameters were already read-only inside the body; the diff just avoids a `TVector3`/`TString` copy per invocation.

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal method parameter passing across detector and geometry components for improved code efficiency.

### DIFF
--- a/Detector/DetectorPoint.cxx
+++ b/Detector/DetectorPoint.cxx
@@ -8,9 +8,10 @@
 
 // -----   Standard constructor  ----------------
 SHiP::DetectorPoint::DetectorPoint(Int_t eventID, Int_t trackID, Int_t detID,
-                                   const TVector3& pos, const TVector3& mom, Double_t tof,
-                                   Double_t length, Double_t eLoss,
-                                   Int_t pdgcode, const TVector3& Lpos, const TVector3& Lmom)
+                                   const TVector3& pos, const TVector3& mom,
+                                   Double_t tof, Double_t length,
+                                   Double_t eLoss, Int_t pdgcode,
+                                   const TVector3& Lpos, const TVector3& Lmom)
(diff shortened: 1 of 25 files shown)


olantwin:
@coderabbitai review

PR #1244, opened 2026-06-14
fix: replace atof/atoi with std::stod/std::stoi

olantwin:
## Summary

Sweeps the 38 `atof`/`atoi` sites in `field/ShipFieldMaker.cxx` (33) and `splitcal/splitcalHit.cxx` (5) that triggered `bugprone-unchecked-string-to-number-conversion` on master run [27465304759](https://github.com/ShipSoft/FairShip/actions/runs/27465304759).

## Why this matters

`atoi`/`atof` silently return `0` on parse failure. A malformed field-map config file would have produced silent zero magnetic fields instead of a startup error — a real footgun for a HEP simulation framework. The `std::stoi`/`std::stod` replacements throw `std::invalid_argument` or `std::out_of_range` with the offending substring, surfacing the parse error at simulation start.

## Transform

```
std::atof(s.c_str()) → std::stod(s)
std::atoi(s.c_str()) → std::stoi(s)
atoi(s.c_str())      → std::stoi(s)
```

`std::stoi`/`std::stod` take `const std::string&` directly, so the `.c_str()` is also dropped.

## Out of scope

The two `fscanf` calls in `shipgen/Pythia6Generator.cxx` (the remaining two `bugprone-unchecked-string-to-number-conversion` findings) are a different parser pattern and out of scope for this PR; they need a proper rewrite around `std::istringstream` and will be tracked separately.

## Test plan

- [x] `pixi run --locked build` clean (133/133, 0 warnings)
- [x] `pixi run --locked test` — 2/2 tests pass (DataClassIO, RNtupleIO)
- [x] `CPP_FILES=… pixi run --locked ci-clang-tidy` reports 0 `bugprone-unchecked-string-to-number-conversion` findings in the touched files

## Risks

A previously-tolerated malformed config file will now throw at startup. This is the intended fail-fast behaviour, but worth noting in case any in-flight branch relied on the silent-zero quirk.

## Summary by CodeRabbit

* **Refactor**
  * Improved numeric parsing robustness by replacing legacy C-style conversions with modern C++ string-to-number functions across field initialization and decoding operations.

### DIFF
--- a/field/ShipFieldMaker.cxx
+++ b/field/ShipFieldMaker.cxx
@@ -188,9 +188,9 @@ void ShipFieldMaker::defineUniform(const stringVect& inputLine) {
   if (nWords == 5) {
     const TString name(inputLine[1].c_str());
 
-    Double_t Bx = std::atof(inputLine[2].c_str());
-    Double_t By = std::atof(inputLine[3].c_str());
-    Double_t Bz = std::atof(inputLine[4].c_str());
+    Double_t Bx = std::stod(inputLine[2]);
+    Double_t By = std::stod(inputLine[3]);
+    Double_t Bz = std::stod(inputLine[4]);
     const TVector3 BVector(Bx, By, Bz);
 
     this->defineUniform(name, BVector);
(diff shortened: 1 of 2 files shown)


olantwin:
@coderabbitai review

PR #1247, opened 2026-06-14
fix: initialise variables before conditional branches

olantwin:
## Summary

Resolves the 31 *may be uninitialized* (`unbound-name`) errors from pyrefly on master run [27465304759](https://github.com/ShipSoft/FairShip/actions/runs/27465304759). Each site assigned a name only inside an `if options.X` / `if condition` branch and read it later, where pyrefly cannot statically prove the same predicate will still hold — and in some cases it genuinely cannot.

## Per-file fixes

| File | Fix |
|---|---|
| `python/ACTSReco.py` | Convert three independent `if` branches (`SiliconTarget`, `MTC`, `StrawTracker`) to `if / elif / else`, raising on unknown detector strings. Catches a real configuration bug. |
| `python/SciFiMapping.py` | Early-return when the fibre loop has no iterations (so the SiPM block doesn't dereference an unset `globfiberID`); initialise the channel-bookkeeping ints and the per-loop `alpha` to defaults. |
| `python/geomGeant4.py` | Pull `fieldsList = []` out of the `hasattr(shipGeo, "Bfield")` block so subsequent unconditional `.append` / `len(...)` calls always have a list to work with. |
| `python/hnl.py` | Make the quark-flavour cascade exhaustive with `else: raise`; catches the previously-silent case of an unknown `beta`. |
| `python/rpvsusy.py` | Initialise `lep = ""` so the f-string is defined even when no lepton was matched. |
| `macro/ShipAna.py` | Open `fgeo` unconditionally; the prior branch computed the geometry path but never opened the file. |
| `macro/ShipReco.py` | Initialise `dy = options.dy` and `fieldMaker = None` before their conditional binders. |
| `macro/makeCascade.py`, `macro/makeDecay.py`, `macro/runPythia8.py` | Initialise `idpn` / `wspill` / `sigma` to sentinel defaults matching the pre-loop state expected by downstream logic. |
| `macro/run_fixedTarget.py` | Bind `sensPlaneT = None`, then narrow with `if sensPlaneT is not None` in each later block (cleaner than re-checking the original arg flag, which pyrefly can't tie back). |
| `macro/run_simScript.py` | Pre-initialise `motherMode`, `Opt_high`, `theCouplings`, `P8gen` to `None`; replace the trailing `if "P8gen" in globals():` introspection with `if P8gen is not None:` narrowing. |

## Test plan

- [x] `pyrefly check --output-format=full-text python macro` — `may be uninitialized` count drops from 31 → 0
- [x] Total error count drops from 168 → 149 (the missing 19 includes a few downstream narrowings that the fresh `None` initialisations resolved)
- [x] No behavioural change for the cases where the original branch always fired; the cases where it didn't (`ShipAna.py`'s `fgeo`, `ACTSReco.py`'s unhandled-detector path) now fail loudly instead of obscurely

Second of three planned Pyrefly cleanups; next will tackle `None`-narrowing across the remaining ~80 `unsupported-operation` sites.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of simulation and processing scripts by ensuring proper variable initialization, reducing potential edge-case failures.
  * Enhanced error handling in detector configuration and physics generators to provide clearer feedback for unsupported configuration options.
  * Fixed control flow logic for more consistent and predictable behavior across simulation workflows.

### DIFF
--- a/macro/ShipAna.py
+++ b/macro/ShipAna.py
@@ -60,8 +60,7 @@ else:
 
 if not options.geoFile:
     options.geoFile = options.inputFile.replace("ship.", "geofile_full.").replace("_rec.", ".")
-else:
-    fgeo = ROOT.TFile(options.geoFile)
+fgeo = ROOT.TFile(options.geoFile)
 
 # new geofile, load Shipgeo dictionary written by run_simScript.py
 ShipGeo = load_from_root_file(fgeo, "ShipGeo")
--- a/macro/ShipReco.py
+++ b/macro/ShipReco.py
@@ -85,6 +85,7 @@ vertexing = not options.noVertexing
 
 
 # need to figure out which geometry was used, only needed if no geo file
+dy = options.dy
 if not options.dy:
     # try to extract from input file name
     tmp = options.inputFile.split(".")
@@ -154,6 +155,7 @@ modules = shipDet_conf.configure(run, ShipGeo)
 sGeo = fgeo["FAIRGeom"]
 import geomGeant4
 
+fieldMaker = None
 if hasattr(ShipGeo.Bfield, "fieldMap"):
     fieldMaker = geomGeant4.addVMCFields(ShipGeo, "", True, withVirtualMC=False)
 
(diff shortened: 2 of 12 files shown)


olantwin:
@coderabbitai review

PR #1255, opened 2026-06-16
perf: move parameters that are copied once into a member

olantwin:
## Summary

Sweeps the 10 `performance-unnecessary-value-param` findings of the *"passed by value and only copied once; consider moving it"* variant that PR #1241 deliberately skipped — that PR handled only the *"only used as a const reference"* subset where a const-ref signature change was the right fix.

Each touched setter assigns a `TString` or `std::string` parameter into exactly one member. Adding `std::move` at the assignment site lets the member take the parameter's storage by move-assignment instead of deep-copying it, while the by-value parameter still binds rvalues cleanly without forcing a const-ref overload.

## Sites

| File | Setter |
|---|---|
| `muonShieldOptimization/exitHadronAbsorber.h:36` | `SetVetoPointName(TString)` |
| `shipgen/FixedTargetGenerator.h:57` | `SetTarget(TString s, ...)` |
| `shipgen/Generator.h:38` | `UseExternalFile(std::string x, ...)` |
| `shipgen/Pythia8Generator.h:56` | `SetTarget(TString s, ...)` |
| `shipgen/tPythia6Generator.h:32,33` | `SetTarget(TString Type, TString Target)` |
| `veto/veto.h:64,66,67,68` | `SetVesselStructure(...)` (4 × `TString`) |

Each header gains a `#include <utility>` for `std::move`.

## Test plan

- [x] `pixi run --locked build` — clean (33/33 incremental, 0 warnings)
- [x] `pixi run --locked test` — 2/2 pass (DataClassIO, RNtupleIO)
- [x] `CPP_FILES=<all-touched-cxx> pixi run --locked ci-clang-tidy` — 0 `performance-unnecessary-value-param` findings remain

## Notes

clang-tidy's autofix doesn't propose this fix on its own (it would need to see both the parameter binding and the body assignment as a unit, which the `performance-unnecessary-value-param` check doesn't combine). Each fix here is hand-applied but mechanical.

The parameters are by-value because the existing call sites pass mixed lvalues and rvalues; switching to const-ref would force a copy on every rvalue caller anyway. Move-from-by-value is the modern idiom for "I want to take ownership but I don't want two overloads".

## Summary by CodeRabbit

* **Refactor**
  * Optimized memory handling in multiple setter methods across the generator and veto components for improved performance.
  * Enhanced internal data transfer mechanisms for target name assignments, external file path handling, and configuration parameters.
  * Changes ensure more efficient resource usage when updating system settings without affecting user-facing functionality.

### DIFF
--- a/muonShieldOptimization/exitHadronAbsorber.h
+++ b/muonShieldOptimization/exitHadronAbsorber.h
@@ -5,6 +5,8 @@
 #ifndef MUONSHIELDOPTIMIZATION_EXITHADRONABSORBER_H_
 #define MUONSHIELDOPTIMIZATION_EXITHADRONABSORBER_H_
 
+#include <utility>
+
 #include "Detector.h"
 #include "TFile.h"
 #include "TNtuple.h"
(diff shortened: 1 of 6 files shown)


olantwin:
@coderabbitai review

olantwin:
@coderabbitai rate limit

olantwin:
@coderabbitai review

PR #1256, opened 2026-06-16
stubs(ROOT): add SetMaxTheta, nrOfDP, IsEqual, mnstat

olantwin:
## Summary

Three real PyROOT/ROOT methods used in-tree but missing from the hand-curated stubs:

| Symbol | Used in | Notes |
|---|---|---|
| `Pythia8Generator.SetMaxTheta` | `macro/run_simScript.py:618` | defined on C++ base `SHiP::Generator` (`shipgen/Generator.h:51`); hoist to `Pythia8Generator` stub so `HNLPythia8Generator` / `DPPythia8Generator` inherit |
| `Pythia8Generator.nrOfDP` | `macro/run_simScript.py:986` | defined on `DPPythia8Generator` (`shipgen/DPPythia8Generator.h:88`); base stub keeps subclass call sites typed |
| `TH1.IsEqual` | `python/experimental/compare_histograms.py:17` | bin-content comparison |
| `TMinuit.mnstat` | `python/shipVertex.py:286` | fit-status accessor for migrad-convergence reporting |

## Test plan

- [x] `pyrefly check --output-format=full-text python macro` drops from 128 → 125 errors
- [x] No new findings introduced; the remaining `SetMaxTheta` site is a `NoneType` narrowing (will be resolved by PR-D2)

Surface-minimal per the stubs file's own guidance.

## Summary by CodeRabbit

* **Chores**
  * Enhanced Python type interface definitions to support histogram comparison, statistical analysis, and particle generator configuration methods, improving IDE autocompletion and error detection.

PR #1258, opened 2026-06-16
fix(pyrefly): narrow Optional and dynamic attrs across small files

olantwin:
## Summary

Per-site narrowing batch for the residual `NoneType has no attribute X` / `X is uninitialized` / `is not subscriptable` findings that the noise reduction made visible.

## Per-file fixes

| File | Fix |
|---|---|
| `python/configGenieGenerator.py:13` | `top.FindNode()` returns Optional; assert `volNuTauMudet_1` exists |
| `python/detectors/SBTDetector.py:58` | `BaseDetector` sets `self.MCdet = None` when no `mcBranchName`; guard the `push_back` chain |
| `python/experimental/compare_histograms.py:40,45` | `TClass.GetClass()` returns Optional; bind and skip-when-None instead of dereferencing |
| `macro/convertEvtCalc.py:52,117,146` | `parse_file` returned `(None, None, None)` on error and a `list` on success; collapse error branches to `return None`, check `if not parsed_data` at the caller, assert in the in-loop append |
| `macro/eventDisplay.py` | Class-level `Targetz: float = 0.0` on `DrawVetoDigi`; `assert fRun is not None` in `toggleMCTracks` and `NextEvent`; one `# pyrefly: ignore` for the idiomatic tkinter-dynamic `Checkbutton.var` attach |
| `python/pythia8darkphoton_conf.py:47` | initialise `hname = \"\"` so the `else` branch never sees an unbound name |
| `python/shipStrawTracking.py` | rebuild `hits` dict comprehension-style (was a TypedDict-mutation), guard `pos is None` after `extrapolateToPlane`, switch `max(track, key=track.get)` to a lambda |

## Test plan

- [x] \`pyrefly check --output-format=full-text python macro\` drops from 128 → 109 (19 fewer; 1 suppression)
- [x] No behavioural change for the success paths; failure modes that previously produced cryptic \`'NoneType' has no attribute X\` runtime errors now fail closer to the cause via \`assert\`

The remaining 2 sites (\`compare_histograms.py:17 TH1.IsEqual\` and \`run_simScript.py SetMaxTheta/nrOfDP\`) are stub gaps already fixed in #1256.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stronger validation for several data-loading and configuration paths, with clearer failures when required inputs or headers are missing.
  * Improved event display and track-handling behavior so controls stay in sync and invalid runtime states are handled safely.
  * Made histogram comparison and tracking routines more defensive, avoiding crashes on missing classes or unsuccessful extrapolations.
  * Fixed digitization logic to skip optional data updates when the relevant data structure is unavailable.

### DIFF
--- a/macro/convertEvtCalc.py
+++ b/macro/convertEvtCalc.py
@@ -47,7 +47,8 @@ def parse_file(infile: str):
                 current_block.append(line.strip())
 
         if current_block:
-            assert process_type is not None and sampled_points is not None  # first header sets both
+            if process_type is None or sampled_points is None:
+                raise ValueError("bin data encountered before any header")
             data = np.loadtxt(current_block)
             variables = [data[:, i] for i in range(data.shape[1])]
             parsed_data.append((process_type, sampled_points, variables))
@@ -57,7 +58,7 @@ def parse_file(infile: str):
     except FileNotFoundError:
         print(f"- convertEvtCalc - Error: The file {infile} was not found.")
         return None
-    except Exception as e:
+    except (ValueError, IndexError) as e:
         print(f"- convertEvtCalc - An error occurred: {e}")
         return None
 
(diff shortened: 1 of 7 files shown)


olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

olantwin:
@coderabbitai rate limit

jasminwss (commented):
Looks good to me. Most of the changes just check something exists before using it, to avoid error messages or vectorise.

olantwin:
@jasminwss Thanks for checking :)

PR #1259, opened 2026-06-16
fix(shipVertex): give VertexError a uniform 3-tuple return

olantwin:
## Summary

`VertexError` returned `(X, dist)` when `CovMat` was `None` and `(X, covX, dist)` otherwise. Pyrefly types the polymorphic return as `tuple[X, float] | tuple[X, TMatrixD, float]` and flagged both the 2-value and 3-value unpacks at the four call sites as `bad-unpacking`.

## Changes

- Always return three values, with `covX = None` on the no-covariance path.
- Update the four call sites in `TwoTrackVertex` (lines 138, 167, 188, 220) to bind a `_covX` slot for the early-return branch.
- Add `assert scalFac is not None` inside the CovMat branch — the existing `CovTracks` loop already relied on this implicitly when indexing `scalFac[tlist[k]]`, which caused two `None is not subscriptable` findings at lines 577/579.

## Test plan

- [x] `pyrefly check --output-format=full-text python macro` drops 128 → 122 (6 fewer in shipVertex.py)
- [x] No behavioural change for either branch; the early-return path's None slot is `_covX`-bound at all callers

The remaining `TMinuit.mnstat` missing-attribute finding is a stub gap covered by #1256.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vertex calculation handling so results are returned consistently, even when no covariance data is available.
  * Fixed vertex recomputation after extrapolation to correctly process the full set of returned values.
  * Added a safeguard when scaling track covariance inputs, helping prevent invalid calculations.

### DIFF
--- a/python/shipVertex.py
+++ b/python/shipVertex.py
@@ -135,7 +135,7 @@ class Task:
                 # ignore this for background studies
                 if PosDirCharge[t2]["charge"] == c1:
                     continue
-                newPos, doca = self.VertexError(t1, t2, PosDirCharge)
+                newPos, _covX, doca = self.VertexError(t1, t2, PosDirCharge)
                 # Extrapolate both tracks toward the initial vertex
                 # estimate in steps to avoid long backward extrapolation
                 # through the magnetic field which causes RK failures.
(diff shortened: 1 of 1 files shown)


olantwin:
@coderabbitai rate limit

PR #1260, opened 2026-06-16
fix(Pythia6Generator): replace fscanf parsing with istringstream

olantwin:
## Summary

clang-tidy's `bugprone-unchecked-string-to-number-conversion` flagged both `fscanf` calls in `Pythia6Generator::ReadEvent` (the remaining 2 sites from that category after PR #1244 swept atoi/atof). `fscanf` silently produces zeros when an input field doesn't match the format — the value just stays at its prior contents — instead of surfacing the parse error to the caller.

## Change

Rewrite with `fgets` + `std::istringstream`. Each line is read into a buffer, then `>>` extracts the typed fields one at a time. If any extraction fails, log the offending line and abort the event, matching the `std::stoi`/`std::stod` error model PR #1244 adopted.

Both the header line (`eventID`, `ntracks`) and the per-track line (15 fields) get explicit fail checks. End-of-file detection now naturally drops out of `fgets` returning `nullptr`.

## Test plan

- [x] `pixi run --locked build` — clean (133/133, 0 warnings)
- [x] `pixi run --locked test` — 2/2 pass
- [x] `CPP_FILES=shipgen/Pythia6Generator.cxx pixi run --locked ci-clang-tidy` — 0 `bugprone-unchecked-string-to-number-conversion` findings
- [ ] CI's `ci-fixed-target` job still passes (its Pythia6 input is well-formed, so behaviour is unchanged)

## Risks

Malformed Pythia6 input lines that previously fell through with zero fields will now log an error and skip the event. If any existing input file relies on the zero-fallback, this will surface as a runtime failure with a clear message — preferred to silent corruption.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced event and track input parsing with improved error detection and validation for malformed or incomplete data entries.

### DIFF
--- a/shipgen/Pythia6Generator.cxx
+++ b/shipgen/Pythia6Generator.cxx
@@ -10,6 +10,8 @@
 
 #include <cstdio>
 #include <iostream>
+#include <sstream>
+#include <string>
 
 #include "FairLogger.h"
 #include "FairPrimaryGenerator.h"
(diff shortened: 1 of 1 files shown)


olantwin:
@coderabbitai rate limit

olantwin:
@coderabbitai review

PR #1274, opened 2026-06-22
fix(shipgen): make tPythia6Generator concrete again (#1272)

### LINKED ISSUE #1272: tpythia6Generator error in makeCascade.py
Dear all,

Launching makeCascade (latest build 26.06) gives me a weird error in imported module:

  File "/afs/cern.ch/work/a/aiuliano/public/ECN3_SHIPBuild/FairShip/macro/makeCascade.py", line 126, in <module>
    tp = ROOT.tPythia6Generator()
         ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: none of the 2 overloaded methods succeeded. Full details:
  cannot instantiate abstract class 'tPythia6Generator' (from derived classes, use super() instead)
  cannot instantiate abstract class 'tPythia6Generator' (from derived classes, use super() instead)

I tried to look for a solution, but I was not able to. It seems to be related to the recent builds somehow.
Has anyone any suggestion?
Antonio

olantwin:
## Summary
- `tPythia6Generator` became abstract in 26.02 when it was reparented onto `SHiP::Generator` (which has pure-virtual file-based `Init` overloads) without stub overrides; cppyy correctly refuses to instantiate it, breaking `makeCascade.py` and the Pythia6 branch of `run_simScript.py`. Add the stub `Init(const char*)` / `Init(const char*, int)` pair that the other generators use.
- Closes #1272.

## Test plan
- [x] `pixi run build` succeeds
- [x] `python -c "import ROOT; ROOT.gSystem.Load('libShipGen'); ROOT.tPythia6Generator()"` constructs cleanly (was `TypeError: cannot instantiate abstract class`)
- [x] `makeCascade.py`-style sequence (`TDatabasePDG`, `TPythia6`, `tPythia6Generator`, `SetMom`, `SetTarget`) runs past the reported failure site
- [ ] Reporter confirmation on a real `makeCascade.py` run

## Follow-up
The sibling class `Pythia6Generator` (no `t`) had the same latent abstract-class bug but is unused (zero callsites anywhere in the tree, no substantive commit since 2008). A separate PR will remove it rather than fix it.

## Bug Fixes
* Restored Pythia6Generator instantiation from Python, which was broken following a previous refactor.
* Added file-based initialization method overloads accepting input filenames and optional event indices; methods currently unsupported but prevent crashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ it in future.
 
 ### Fixed
 
+* Restore `tPythia6Generator` instantiation from Python — broken since 26.02 by the `SHiP::Generator` base-class refactor leaving the file-based `Init` overloads pure virtual without a stub override (#1272)
+
 ### Removed
 
 ## 26.06 - 2026-06-18
--- a/shipgen/tPythia6Generator.h
+++ b/shipgen/tPythia6Generator.h
@@ -7,6 +7,7 @@
 
 #include <utility>
 
+#include "FairLogger.h"
 #include "Generator.h"
 #include "TPythia6.h"
 #include "TPythia6Calls.h"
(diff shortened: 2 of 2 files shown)


olantwin:
@coderabbitai review

antonioiuliano2 (approved):
Perfect!
Thank you very much Oliver for this fix.
I confirm it works, the code still crashes but it's for a different reason, not related to this PR. 
It looks for PDG particle -130 which does not exist, and the `if _p is not None `condition is not working as expected. I am now opening a new dedicated PR.
For this one, everything ok.

PR #1275, opened 2026-06-22
chore(shipgen): remove unused Pythia6Generator

olantwin:
## Summary
- Audit (follow-up to #1272 / #1274) found `Pythia6Generator` (no `t`) has zero callsites: no Python script constructs it, no C++ caller exists, no docs / README / CHANGELOG reference it, no CI workflow exercises it. Its `.cxx` and `.h` have only seen cosmetic refactors since the 2008 import.
- It also has two latent bugs that #1274 was about to paper over: it inherits abstract `Init` overloads from `SHiP::Generator` without overriding them, and its default constructor leaves raw `FILE*` / `Char_t*` pointers indeterminate, crashing the destructor on default-construction.
- Delete the class rather than fix it. `tPythia6Generator` is unaffected (it is actively used by `macro/makeCascade.py` and `run_simScript.py --Pythia6`). The `ROOTEGPythia6` dependency stays for `tPythia6Generator`.

## Test plan
- [x] `pixi run build` succeeds with no unresolved references
- [x] `hasattr(ROOT, 'Pythia6Generator')` returns `False` after loading `libShipGen`
- [x] `tPythia6Generator` source file is untouched (`git diff master -- shipgen/tPythia6Generator.*` is empty); independent of this PR, it still has the #1272 abstract-class bug that #1274 fixes
- [x] No straggler reference found by full-tree grep
- [x] CI passes

Independent of #1274 — neither PR blocks the other.

## Summary by CodeRabbit

* **Removed**
  * Removed the legacy text-format event reader. An alternative implementation remains available for users who require this functionality.


olantwin:
@coderabbitai review

PR #1291, opened 2026-06-30
Simplify TFile open checks and fix graceful-degradation crash

olantwin:
## Summary

Cleans up `TFile` open/zombie checks across the C++ generators and Python code, and fixes a latent crash where unreadable ROOT files were meant to be skipped.

The driving observation is that `TFile::Open` already handles zombies for you, but it does so differently in the two languages:

- **C++ `TFile::Open`** returns `nullptr` on failure (it deletes the zombie object internally). A non-null result is therefore always non-zombie, so `if (!f || f->IsZombie())` is redundant — the `IsZombie()` half can never fire.
- **PyROOT `TFile.Open` *and* the `TFile(...)` constructor** are pythonized to **raise `OSError`** on failure (see `ROOT/_pythonization/_tfile.py`). They never return `None` or a zombie, so the usual `if not f: ...` guards after them are unreachable dead code.

## Changes

**`refactor(shipgen)`** — drop the redundant `|| ->IsZombie()` from all 9 generators, keeping the live `if (!f)` null check.

**`fix`** — three Python sites intended to *degrade gracefully* on a bad file (skip + warn, return `None`, print + return) never did: because `TFile.Open` raises `OSError`, a bad file crashed the caller instead. Wrap those opens in `try/except OSError` so they behave as intended:
- `.github/scripts/render_plots.py` — skip the file and warn
- `.github/scripts/extract_physics_metrics.py` — return `None`
- `python/validationTools.py` — print a warning and return

**`refactor`** — remove the now-dead `if not f:` guards at the re-raise sites (the open already raises `OSError`), convert the remaining `ROOT.TFile(...)` constructors to `TFile.Open` for consistency, and manage `genie_interface.add_hists`' files with a combined `with` statement (which also closes the input file if opening the output file fails).

## Testing

- `prek` (ruff, ruff-format, mypy, clang-format, cpplint, codespell, …) passes on all changed files.
- All 9 changed C++ files pass `-fsyntax-only` with their real compile flags from `compile_commands.json`.
- Runtime checks: `extract_metrics_from_file(<missing>)` now returns `None` instead of crashing; `add_hists` closes the input file when the output open fails, still raises `FileNotFoundError` on a missing histogram, and copies the histogram on the happy path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated several analysis, validation, and generator workflows to handle file-open failures more consistently.
  * Some file checks now rely on standard open errors, which may change when invalid inputs are reported during processing.

* **Refactor**
  * Simplified ROOT file opening and cleanup logic across scripts and generators.

### DIFF
--- a/.github/scripts/extract_physics_metrics.py
+++ b/.github/scripts/extract_physics_metrics.py
@@ -196,10 +196,12 @@ def extract_metrics_from_file(file_path: Path, config: dict[str, Any]) -> dict[s
         2. Add a new key to the metrics dict (e.g., "graphs": {})
         3. Add an isinstance check in the loop below
     """
-    with ROOT.TFile.Open(str(file_path)) as root_file:
-        if not root_file or root_file.IsZombie():
-            return None
+    try:
+        root_file = ROOT.TFile.Open(str(file_path))
+    except OSError:
+        return None
 
+    with root_file:
         metrics = {
             "file_size": file_path.stat().st_size,
             "trees": {},
--- a/.github/scripts/render_plots.py
+++ b/.github/scripts/render_plots.py
@@ -110,8 +110,9 @@ def main() -> int:
     all_rendered: list[dict[str, str]] = []
 
     for input_path in args.inputs:
-        root_file = ROOT.TFile.Open(input_path)
-        if not root_file or root_file.IsZombie():
+        try:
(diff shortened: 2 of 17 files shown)


olantwin:
@coderabbitai review

PR #1301, opened 2026-06-30
Freeze Eigen in Renovate

olantwin:
Disables Renovate updates for `eigen`, which is pinned to `3.4.*` because acts-ship requires Eigen 3.4.0 EXACT; without this Renovate would open a build-breaking 5.x bump.


olantwin:
@coderabbitai review

PR #1310, opened 2026-07-02
Fix correctness bugs in shipgen event generators

olantwin:
Fixes bug-category findings from a full-tree code review, scoped to `shipgen/`. Commits are grouped by risk/kind.

## Crash / high severity
- **MuDISGenerator**: dereferenced the material pointer even when the sampled point was outside the geometry (`FindNode` failed / `IsOutside`) → segfault. Now guards on the node/material and resamples, matching `GenieGenerator`.
- **ParticleGunGenerator**: the `uniform` vertex model set X to the width (`pars[1]`) instead of the central value (`pars[0]`) in the non-positive-width fallback, unlike the Y branch and gaussian model.

## Physics behaviour (reviewed with maintainer)
- **Pythia8Generator**: the primary interaction-point sampling was gated on `targetName != ""`, but the modern `SetTargetCoordinates()` path (used by run_simScript charmonly) leaves `targetName` empty. `zinter` stayed 0, so all charm-hadron vertices were placed at z≈0 instead of distributed through the target with interaction-length weighting. Now gated on `targetFromGeometry || targetName != ""`. Also guards `ReadEvent` against a null input tree (the standalone Init branch has no event path).

## Medium
- **MuonBackGenerator**: the downscaled-dimuon recovery loop keyed the muon map with the loop counter instead of the actual vetoPoint index (`list.at(i)`).
- **TTreeGenerator**: primaries were stored with `E = 0` (passed `e=0.`; FairPrimaryGenerator only recomputes when `e<0`). Now passes `-9e9`.
- **tPythia6Generator**: mother index read PYJETS `V(I,4)` (production time) instead of `K(I,3)` (parent line).
- **HNL/DPPythia8Generator**: `fUseRandom1/3` were uninitialized; a null RNG engine could be handed to `setRndmEnginePtr`. Defaulted to TRandom3 in-class.
- **CosmicsGenerator.h**: `fRandomEngine` raw pointer was indeterminate yet deleted in the destructor; defaulted to `nullptr`.

## Low
- **FixedTargetGenerator**: `pot` counter was `Int_t` incremented by 0.5 (truncated to 0); made `Double_t`. Unknown `Option` now aborts `Init()` instead of continuing half-configured.
- **CosmicsGenerator**: `!nInside % 10000` parsed as `(!nInside) % 10000`; fixed the progress modulo.

## Verification
Built cleanly with pixi; `ci-sim`, `ci-reco`, and `ci-fixed-target` all complete successfully (the last exercises FixedTargetGenerator; ci-sim exercises HNLPythia8Generator).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event generation reliability by fixing several edge cases in particle, cosmic, muon, and fixed-target workflows.
  * Corrected progress reporting so milestone messages appear at the expected intervals.
  * Added safer handling for invalid settings and missing inputs to prevent crashes or silent bad results.
  * Fixed track and vertex bookkeeping so generated events use the correct source values and energy handling.

### DIFF
--- a/shipgen/DPPythia8Generator.h
+++ b/shipgen/DPPythia8Generator.h
@@ -94,11 +94,11 @@ class DPPythia8Generator : public SHiP::Generator {
 
  protected:
   // Bool_t fHadDecay;    //select hadronic decay
-  Double_t fMom;       // proton energy
-  Int_t fDP;           // DP ID
-  Int_t fId;           // target type
-  Bool_t fUseRandom1;  // flag to use TRandom1
-  Bool_t fUseRandom3;  // flag to use TRandom3 (default)
+  Double_t fMom;               // proton energy
+  Int_t fDP;                   // DP ID
+  Int_t fId;                   // target type
+  Bool_t fUseRandom1{kFALSE};  // flag to use TRandom1
+  Bool_t fUseRandom3{kTRUE};   // flag to use TRandom3 (default)
   Bool_t
       fpbrem;  // flag to do proton bremstrahlung production (default is false)
   TH2F* fpbremPDF;   // pointer to TH2 containing PDF(p,theta) to have a dark
--- a/shipgen/CosmicsGenerator.cxx
+++ b/shipgen/CosmicsGenerator.cxx
@@ -186,7 +186,7 @@ Bool_t CosmicsGenerator::ReadEvent(FairPrimaryGenerator* cpg) {
   cpg->AddTrack(
       PID, px, py, pz, x, y, z, -1, true, TMath::Sqrt(P * P + mass * mass), 0,
       weight);  // -1 = Mother ID, true = tracking, SQRT(x) = Energy, 0 = t
-  if (!nInside % 10000) {
+  if (nInside % 10000 == 0) {
     cout << nInside / 10000 << "10k events have been simulated" << endl;
   }
   return kTRUE;
--- a/shipgen/CosmicsGenerator.h
+++ b/shipgen/CosmicsGenerator.h
@@ -78,7 +78,7 @@ class CosmicsGenerator : public SHiP::Generator {
   int n_EVENTS;
 
  private:
-  Co3Rng* fRandomEngine;  //!
+  Co3Rng* fRandomEngine = nullptr;  //!
 
  protected:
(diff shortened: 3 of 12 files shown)


olantwin:
@coderabbitai review

shipgen/DPPythia8Generator.h
```diff
@@ -94,11 +94,11 @@ class DPPythia8Generator : public SHiP::Generator {
...
   // Bool_t fHadDecay;    //select hadronic decay
-  Double_t fMom;       // proton energy
-  Int_t fDP;           // DP ID
-  Int_t fId;           // target type
-  Bool_t fUseRandom1;  // flag to use TRandom1
-  Bool_t fUseRandom3;  // flag to use TRandom3 (default)
+  Double_t fMom;               // proton energy
+  Int_t fDP;                   // DP ID
+  Int_t fId;                   // target type
+  Bool_t fUseRandom1{kFALSE};  // flag to use TRandom1
```
THanae, line 100: Would there be reasons to use TRandom1 over TRandom3? From documentation I also get the impression that `TRandomRanluxpp` would be faster / more precise than `TRandom1` (both are based on the RANLUX algorithm). `TRandomMixMax` (different algorithm from the ones above) also seems faster / more precise -- timings are [here](https://root.cern.ch/doc/v632/classTRandom.html) for reference.

olantwin: I'm not sure we have large enough stats for signal that the issues with MT would be visible. I also don't like us setting a different random engine just in the generator, we'd probably want to have a way to compare random engines globally.

Iff we want to have a different random engine here, indeed TRandomMixMax or TRandomRanluxpp seem for all purposes better than TRandom1.

olantwin: Note that for aegir,we need to use a counter-based RNG if we want reproducibility, e.g. Philox.

olantwin: How shall we proceed here? Remove the TRandom1 option? Add another? Leave as is pending a larger cleanup of RNG options?

THanae: I think your comment about not liking setting different random engines in different places is valid. I would lean towards your last suggestion (leaving this for an overall cleanup of RNG options)

PR #1316, opened 2026-07-02
Fix correctness bugs in macro/, muonDIS/, muonShieldOptimization/

olantwin:
Fixes bug-category findings from a full-tree code review, scoped to the driver macros and muon-DIS / muon-shield tooling. Commits grouped by subsystem/kind. Note: several of these files were refactored since the review, so each finding was re-verified against current `master`.

## Dead-on-arrival crashes
- **muonDIS scripts** (`makeMuonDIS`, `make_nTuple_SBT`, `make_nTuple_Tr`, `add_muonresponse`): `TClonesArray[i] = obj` raises `RuntimeError` in ROOT ≥ 6.32, so all four crashed on the first event. Replaced with `ConstructedAt(i).__assign__(obj)` (point objects) / `ConstructedAt(i)` + `ResizeTo` + `__assign__` (TVectorD), the idiom verified against the built ROOT.
- **Import-time crashes**: `mergeMbias` (`hadronAbsorber.length`), `dumpEvent` (`Config.loadpy`), `getInteractionAndRadiationLength` (removed `Geniegen.MeanMaterialBudget` member → free `shipgen::MeanMaterialBudget`, param array sized 10).
- **`run_simScript`**: `-f none` set inputFile to None then iterated it (TypeError); `--SusyBench` lacked `type=int`; three error paths exited 0.
- **`exitHadronAbsorber`**: uninitialized `EMax`/`fUniqueID`/`fUseCaveCoordinates` read in stepping code (UB) — added in-class initializers (`EMax = 0.` stops no track).

## Logic / correctness
- **`make_nTuple_SBT`**: the final SBT-hit count was written into the UBT column (`-2`) instead of the SBT column (`-3`).
- **`make_nTuple_Tr`** (reviewed): the strawtube loop lacked the per-muon `track_id == muon_` filter its UBT loop and the SBT twin have, cross-contaminating multi-muon events.
- **`makeMuonDIS` DIS cross-section (reviewed — please sanity-check the weights)**: the cross-section was stored after the muon vector was copied to output (previous event's value), and `update_file()` overwrote all entries with the last (neutron) converged value. Now stores it before the copy and applies the converged proton/neutron cross-section per entry based on its target flag.
- **`ShipReco`** `-n` treated as end index; **`ShipAna`** per-event veto decisions moved out of the per-candidate loop; **`eventDisplay`** `-i type=int` + HNL end-vertex found-flag; **`makeCascade`** interpolation-anchor indentation; **`makeDecay`** EOS prefix + print format; **`runPythia8`** empty-histogram guards.
- **muonShieldOptimization**: `run_reco` `-g` + argv tokenisation; `compactingBackgroundProduction` hadd list / CLI wiring / snapshot-iterate; `muDIS_mergeFiles` double-fill; `makeMuonEM` swap-consumes-vector; `run_prod`/`runCharmHadProd` integer/range bugs; `ana_ShipMuon` discarded `origin()` / `.Get()` on filename strings; `extractNeutrinosAndUpdateWeight` duplicated PDG 4232 → 4322.

## Not changed
- The MuDIS `CrossSection` post-processing finding no longer applies (that code was refactored away).
- The `add_muonresponse` in-place-rewrite metadata-loss finding was left for a separate change per maintainer preference.

## Verification
Built cleanly (the `exitHadronAbsorber` C++ change compiles); the full `ci-sim` → `ci-reco` → `ci-ana` → `ci-fixed-target` chain completes with no errors (exercises run_simScript, ShipReco, ShipAna). The muon-DIS / muon-shield scripts are not covered by CI (they need dedicated input data) — they are AST-clean, pass ruff/mypy, and use the ROOT-verified `ConstructedAt` idiom.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected event-count handling for runs beginning at a nonzero event.
  * Prevented duplicate veto results and empty-data calculation errors.
  * Fixed detector-track displays, particle ancestry reporting, material boundaries, output paths, run ranges, and event-processing edge cases.
  * Improved validation and diagnostics for missing geometry, inputs, and results.

* **New Features**
  * Added material-budget calculation support.
  * Improved detector-hit and ROOT event-data handling.
  * Added support for safer input discovery and clearer processing feedback.


olantwin:
@coderabbitai review

macro/mergeMbias.py
```diff
@@ -17,7 +17,7 @@
 rnr = ROOT.TRandom()
 eospath = ROOT.gSystem.Getenv("EOSSHIP") + "/eos/experiment/ship/data/"
 ship_geo = geometry_config.create_config(Yheight=10.0)
-endOfHadronAbsorber = (ship_geo["hadronAbsorber"].z + ship_geo["hadronAbsorber"].length / 2.0) / 100.0
+endOfHadronAbsorber = ship_geo["hadronAbsorber"].z / 100.0
```
THanae: I am not sure I understand this. Wouldn't this then correspond to the start of the hadron absorber?

olantwin: No, it would correspond to the centre. But yes, you're correct, this does not make sense. I'll fix it.

olantwin: Good catch — fixed. The `.length` term had been dropped because `hadronAbsorber.length` was lost when the legacy `ShipGeo.py` was removed, so the line was crashing at import; I "fixed" it by removing the half-length, which (as you noted) left it at the centre. I've now restored `hadronAbsorber.length = 3.0 m` in `geometry_config.py` and put back the `z + length/2` formula, so it points at the downstream face of the absorber again.

olantwin: Maybe we should take this from the muon shield parameters though. Not sure, @Gfrisella , do you have an opinion?

I don't like that the two lengths could get out of sync.

olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

muonShieldOptimization/run_prod.py
```diff
@@ -233,7 +233,7 @@ def copy2EOS() -> None:
         for run in range(0, 10):
             prefix = "muon" + str(prod + run)
             if prod in [610, 620] and run == 0:
-                prefix = "muon" + str(int(prod / 100))
+                prefix = "muon" + str(int(prod / 10))
```
THanae, line 236: As far as I understand, this part of the code is legacy that was used in past production. Might be better to remove it altogether (or to move it to a `legacy` folder) instead of correcting it? Even if we do decide to keep it, it might be good to add a comment saying this has been updated from a previous value -- that way, we have a clearer trail of what has been done.

olantwin: Good point, this also applies to the comment to run_reco, most likely it would be better to remove or hide the old code.

olantwin: Kept track of in #1382

muonShieldOptimization/run_reco.py
```diff
@@ -132,13 +133,15 @@ def executeSimple(prefixes: list[str], reset=False) -> None:
...
                 with contextlib.suppress(Exception):
                     os.system("rm logRec")
                 if reset:
-                    os.system("python " + cmd + " -n 9999999 -f " + inputfile + " --saveDisk >> logRec &")
+                    os.system(
+                        "python " + cmd + " -n 9999999 -f " + inputfile + " -g " + geofile + " --saveDisk >> logRec &"
+                    )
                 else:
-                    os.system("python " + cmd + " -n 9999999 -f " + inputfile + " >> logRec &")
+                    os.system("python " + cmd + " -n 9999999 -f " + inputfile + " -g " + geofile + " >> logRec &")
```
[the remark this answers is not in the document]
olantwin: Kept track of in https://github.com/ShipSoft/FairShip/issues/1382

olantwin:
@coderabbitai review

macro/makeDecay.py
```diff
@@ -62,10 +62,7 @@
 FIN = fname + ".root"
 tmp = os.path.abspath(FIN).split("/")
 FOUT = "Decay-" + tmp[len(tmp) - 1]
-if FIN.find("eos") < 0:
-    fin = ROOT.TFile(FIN)
-else:
-    fin = ROOT.TFile.Open(ROOT.gSystem.Getenv("EOSSHIP") + FIN)
+fin = ROOT.TFile.Open(FIN)
```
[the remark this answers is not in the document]
olantwin: Full paths should be specified for EOS, we should not train users to rely on hacks like this.

olantwin:
@coderabbitai review

muonDIS/makeMuonDIS.py
```diff
@@ -82,7 +82,10 @@ def update_file(filename: str, final_xsec) -> None:
 
     for i, event in enumerate(original_tree):
         mu = event.InMuon[0]
-        mu[10] = final_xsec[int(first_mu_event + i / args.nDIS)]
+        proton_xsec, neutron_xsec = final_xsec[int(first_mu_event + i / args.nDIS)]
+        # mu[9] is the isProton flag: apply the converged cross-section of the
+        # matching target rather than the last (neutron) value to every entry.
+        mu[10] = proton_xsec if int(mu[9]) == 1 else neutron_xsec
```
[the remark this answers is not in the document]
olantwin: This should now be addressed.

PR #1321, opened 2026-07-03
perf(field): flatten magnetic field-map storage

olantwin:
Addresses the HIGH-severity field-map storage finding.

`ShipBFieldMap` stored the grid as `vector<vector<Float_t>>` — one heap-allocated
3-float vector per node, giving a 3–5× memory overhead and a pointer chase on
every corner access. This switches storage to `vector<array<Float_t,3>>`: one
contiguous block, no per-node allocation, same `[bin][axis]` indexing (populate
and read sites compile unchanged). `ShipBFieldMap` is never persisted, so there is
no on-disk compatibility impact.

Also moves the per-call interpolation scratch out of member variables into
`Field()` locals (the evaluator is now reentrant) and fetches each of the eight
corners once via a shared `triLinearInterp` helper instead of `BInterCalc`
re-reading them per axis.

Verification: field values are **bit-identical** before/after over a grid of
interpolation and quadrant-symmetry sample points; full build and `ctest` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved magnetic field interpolation for smoother, more consistent values across the 3D field map.
  * Applied symmetry and scaling consistently when computing field components.

* **Refactor**
  * Simplified the magnetic field calculation path by using a single trilinear interpolation approach for components.
  * Updated field-map storage to improve lookup efficiency and cache performance.

### DIFF
--- a/field/ShipBFieldMap.cxx
+++ b/field/ShipBFieldMap.cxx
@@ -187,30 +187,49 @@ void ShipBFieldMap::Field(const Double_t* position, Double_t* B) {
   Int_t iY1(iY + 1);
   Int_t iZ1(iZ + 1);
 
-  binA_ = this->getMapBin(iX, iY, iZ);
-  binB_ = this->getMapBin(iX1, iY, iZ);
-  binC_ = this->getMapBin(iX, iY1, iZ);
-  binD_ = this->getMapBin(iX1, iY1, iZ);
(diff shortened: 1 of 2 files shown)


olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

PR #16, opened 2016-11-24
Adding rpv susy generator

kap01:
I have made the required changes to be able to add the rpv susy generation to FairShip.
This includes modifying pythia8_conf.py, moving some common utility functions to both
HNL and SUSY generators into pythia8_conf_utils.py, introducing the script rpvsusy.py which
calculates widths and BFs as well as a set of .dat files which store the BFs for various neutralino mases. This is not the most flexible approach as any additional model would effectively have to
be also added into pythia8_conf as a separate configureXYZ function, 
making this script extremely long. 

BTW no need to pick up the changes to ShipAna.py not sure why that one looks so different to the master but please ignore this as i do not need to change anything to it

PR #1324, opened 2026-07-03
refactor(SND): migrate MC point classes to SHiP::DetectorPoint

olantwin:
Removes the duplicated "FairMCPoint + fPdgCode" boilerplate in four SND MC-point
classes (review finding #16, Part 4) by inheriting the existing
`SHiP::DetectorPoint` base — the same base already used by `vetoPoint`,
`strawtubesPoint`, `splitcalPoint`, `TimeDetPoint`, and `UpstreamTaggerPoint`.

`TTPoint`, `TargetPoint`, `MTCDetPoint`, and `SiliconTargetPoint` now:
- inherit `SHiP::DetectorPoint` (so `fPdgCode`, `PdgCode()`, and `Print` come
  from the base);
- keep a thin 8-arg forwarding constructor, so the `AddHit` call sites, the
  `LinkDef`s, and the IO tests are **unchanged** and the stored eventID stays 0
  as before;
- keep their detID-decoder helpers (`GetLayer`/`GetLayerType`,
  `GetPlane`/`GetColumn`/`GetRow`/`GetStrip`/`GetModule`).

Net ~196 lines removed. `ClassDef` versions are bumped.

Verification: full build plus the `DataClassIO` and `RNtupleIO` round-trip tests,
which construct and read back all four classes (checking `PdgCode()` and the
decoders).

Note on I/O: the migrated points now also carry the base's `fLpos`/`fLmom`
(defaulted to the entrance pos/mom), so new files are slightly larger. `fPdgCode`
moves into the `DetectorPoint` base; reading **legacy** files relies on ROOT's
by-name schema evolution (member-moved-to-base) — worth a read-back check against
an existing legacy file with these branches if one is available.

## Summary by CodeRabbit

* **Refactor**
  * Updated several detector point types to use a shared point base class for more consistent behavior.
  * Simplified point class interfaces by removing now-unneeded custom methods and members.
  * Registered updated classes with the runtime metadata system for better compatibility.

### DIFF
--- a/SND/EmulsionTarget/TTPoint.cxx
+++ b/SND/EmulsionTarget/TTPoint.cxx
@@ -4,34 +4,4 @@
 
 #include "TTPoint.h"
 
-#include <iostream>
-using std::cout;
-using std::endl;
-
-// -----   Default constructor   -------------------------------------------
-TTPoint::TTPoint() : FairMCPoint() {}
-// -------------------------------------------------------------------------
-
-TTPoint::TTPoint(Int_t trackID, Int_t detID, const TVector3& pos,
-                 const TVector3& mom, Double_t tof, Double_t length,
-                 Double_t eLoss, Int_t pdgcode)
-    : FairMCPoint(trackID, detID, pos, mom, tof, length, eLoss),
-      fPdgCode(pdgcode) {}
-
-// -------------------------------------------------------------------------
-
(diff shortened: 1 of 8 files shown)


olantwin:
@coderabbitai review

antonioiuliano2 (approved):
Ok, the detector point classes appear a lot clearer now.

PR #1329, opened 2026-07-03
ci: seed the CI simulation for reproducible metrics

olantwin:
The first physics-metrics comparisons showed multi-sigma differences on a tooling-only PR because ci-sim ran with a time-based seed. Seeding makes the reference comparison sensitive to real regressions; the first post-merge master run will store a seeded reference.


olantwin:
Verified: ran the workflow twice on the same commit and compared the metrics artifacts — 0 differences across all 32 histograms, fits and tree entries in both configurations. The seeded chain is fully reproducible.

PR #1335, opened 2026-07-10
refactor: drop redundant ClassImp macros

olantwin:
## Summary

Removes all `ClassImp` usages in the tree. In modern ROOT the rootcling-generated dictionary provides the class registration, so `ClassImp` only records cosmetic impl-file metadata. Every affected class already has a `ClassDefOverride` in its header **and** a `LinkDef` dictionary entry, making `ClassImp` redundant.

An inventory found exactly **5** usages:

- `Detector/DetectorPoint.cxx` — a real source file with out-of-line methods; only the trailing `ClassImp` line is dropped.
- `MTCDetPoint`, `TargetPoint`, `TTPoint`, `SiliconTargetPoint` — header-only point classes whose `.cxx` files existed *solely* to host `ClassImp`. These stubs are deleted, and their headers moved to `EXTRA_DICT_HEADERS` in the three affected `CMakeLists.txt` (needed because `ship_add_library` derives dictionary headers from `SOURCES`).

## Verification

- The four affected libraries (`Detector`, `MTCDetector`, `EmulsionTarget`, `SiliconTarget`) build and their dictionaries generate cleanly.
- Loading the libraries in PyROOT confirms every class is still registered with its correct schema version — `SHiP::DetectorPoint` v1, `MTCDetPoint`/`TargetPoint`/`TTPoint` v4, `SiliconTargetPoint` v3 — so the change is behaviour-neutral (schema evolution via the `+` LinkDef entries is unaffected).

### DIFF
--- a/Detector/DetectorPoint.cxx
+++ b/Detector/DetectorPoint.cxx
@@ -42,5 +42,3 @@ void SHiP::DetectorPoint::Print() const {
 void SHiP::DetectorPoint::extraPrintInfo() const {
   LOG(info) << "Nothing to see here";
 }
-
-ClassImp(SHiP::DetectorPoint)
--- a/SND/EmulsionTarget/CMakeLists.txt
+++ b/SND/EmulsionTarget/CMakeLists.txt
@@ -3,8 +3,10 @@
 
 ship_add_library(
   NAME EmulsionTarget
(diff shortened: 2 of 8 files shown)


olantwin:
@eduard322 You can follow this pattern for the new class you added.

PR #1340, opened 2026-07-14
Converter of neutrino flux to the aegir-genie format

antonioiuliano2:
Dear all,

This script reads the files produced in FairShip fixed Target productions and makes a neutrino flux file usable by 
[aegir-genie](https://github.com/ShipSoft/aegir-genie/).

It uses as reference muonShieldOptimization/extractNeutrinosAndUpdateWeight.py from FairShip and scripts/make_flux_ntuple.py from aegir-genie.

Example of usage:
`python convert_nu_flux_ntuple.py -f /eos/usual_eos_ship_path/pythia8_Geant4_eCut_1.0_PoT_20000000_CERN_j98.root -p 20000000 --nocharm`.

I have added the converter here instead of aegir-genie scripts since it needs the FairShip classes to function, while it does not need any input library from aegir-genie.

Best Regards,
Antonio
## Checklist

- [ ] CI checks pass

## Summary by CodeRabbit

- **New Features**
  - Added a command-line tool to convert FairShip neutrino simulation files into GENIE-compatible ROOT output.
  - Supports optional charm and tau filtering, metadata generation, and preservation of per-neutrino weights.

- **Bug Fixes**
  - Removed a duplicate particle identifier from muon-shield optimization handling.
  - Corrected neutrino flux weighting to use stored track weights instead of a constant value.

- **Documentation**
  - Updated the changelog with recent simulation, tracking, validation, geometry, and configuration improvements.


macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,131 @@
...
+neutrinos = [-12, 12, -14, 14, -16, 16]
+
+cm2mm = 10.
+nentries = 0
+
+
+f = ROOT.TFile.Open(os.environ["EOSSHIP"] + inputfilename, "READ")  
+print("opened file ", inputfilename)
+sTree = f.Get("cbmsim")
+for n in range(sTree.GetEntries()):
```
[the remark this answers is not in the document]
antonioiuliano2: EOSSHIP is used in almost all FairShip scripts, so I will keep it.
I shall, however, add a check if cbmsim TTree is retrieved successfully.

olantwin:
@antonioiuliano2 Could you please add a stub for the two missing classes?

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+        entry["t"] = v.GetTime()  # ns
+        entry["px"] = t.GetPx()  # GeV
+        entry["py"] = t.GetPy()
+        entry["pz"] = t.GetPz()
+        entry["weight"] = 1.0
+        entry["parent_pdg"] = mompdg
+        entry["parent_px"] = momt.GetPx()
+        entry["parent_py"] = momt.GetPy()
+        entry["parent_pz"] = momt.GetPz()
+        entry["process_id"] = momt.GetProcID()
```
olantwin: Here we take the parent's process ID. This was probably my mistake in aegir-genie. I guess we want the process ID for the neutrino?

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+for n in range(sTree.GetEntries()):
+    sTree.GetEntry(n)
+    header = sTree.MCEventHeader
+    for v in sTree.PlaneHAPoint:
+        nu = v.GetTrackID()
+        t = sTree.MCTrack[nu]
+        pdg = t.GetPdgCode()
+        if abs(pdg) not in neutrinos:
+            continue  # story only neutrinos
+        momt = sTree.MCTrack[t.GetMotherId()]
```
olantwin: We need to be careful here: if the mother ID is -1 (i.e. not saved in FairRoot's convention), we select the last MCTrack due to python's negative indexing semantics.

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+emax = 0.0
+
+charmExtern = [4332, 4232, 4132, 4232, 4122, 431, 411, 421, 15]  # exclude charmed hadrons and tau leptons
+neutrinos = [-12, 12, -14, 14, -16, 16]
+
+cm2mm = 10.0
+nentries = 0
+
+
+f = ROOT.TFile.Open(os.environ["EOSSHIP"] + inputfilename, "READ")
```
olantwin: I don't like how this assumes that the file will be on EOS. This will break for any file that's not on EOS.

It would be better to accept filenames that include env vars, and then expand any env vars (similar to how we do it elsewhere) before opening the file.

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+        writer.Fill(entry)
+del writer
+
+meta = ROOT.RNTupleModel.Create()
+meta.MakeField["std::int32_t"]("schema_version")
+meta.MakeField["double"]("pot")
+meta.MakeField["double"]("max_energy")
+meta.MakeField["std::string"]("description")
+meta.MakeField["std::string"]("software")
+f = ROOT.TFile.Open(out, "UPDATE")
```
olantwin, line 147: f isn't closed when it is reassigned here.

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+model.MakeField["std::int32_t"]("parent_pdg")
+model.MakeField["std::int32_t"]("process_id")
+model.MakeField["std::int64_t"]("origin_run")
+model.MakeField["std::int64_t"]("origin_event")
+
+writer = ROOT.RNTupleWriter.Recreate(model, "nu_flux", out)
+entry = writer.CreateEntry()
+emax = 0.0
+
+charmExtern = [4332, 4232, 4132, 4232, 4122, 431, 411, 421, 15]  # exclude charmed hadrons and tau leptons
```
olantwin: 4232 is duplicated

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+    "--nocharm",
+    dest="noCharm",
+    help="If True, do not store neutrinos from charmed hadrons and tau leptons",
+    action="store_true",
+    default=False,
+)
+
+options = parser.parse_args()
+
+inputfilename = str(options.inputfilename)
```
olantwin: No need to cast

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+if options.out:
+    out = str(options.out)
+else:
+    out = "nu_flux_" + os.path.basename(inputfilename)
+pot_number = int(options.pot_number)
```
olantwin: No need to cast, this argument is defined as type int

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,149 @@
...
+
+noCharm = bool(options.noCharm)
```
olantwin: no need to cast, this is defined using store_true

olantwin (changes requested):
Some nitpicks and a major issue with the filenames.

macro/convert_nu_flux_ntuple.py
antonioiuliano2: I thought of that too, but it would not be informative, right? 
Neutrino process ID is always 4(decay) by physics (except of course when neutrinos are the primary particles simulated, in that case is 0(primary).
We are more interested in which process generated the neutrino parent.

macro/convert_nu_flux_ntuple.py
antonioiuliano2: True, I guess we need to fill the mother information with dummy values in this case.

macro/convert_nu_flux_ntuple.py
antonioiuliano2: Well spotted! It is the same in muonShieldOptimization/extractNeutrinosAndUpdateWeight.py, so I am noting to fix this there too.

macro/convert_nu_flux_ntuple.py
antonioiuliano2: I do not know if this is the solution you were proposing, but I have now copied the check from python/rootUtils.py:

https://github.com/ShipSoft/FairShip/blob/6b32c15be9855c2525b7ed4b4ef21bdb91f5bc2f/python/rootUtils.py#L16-L17

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,165 @@
...
+emax = 0.0
+
+charmExtern = [4332, 4232, 4132, 4122, 431, 411, 421, 15]  # exclude charmed hadrons and tau leptons
+neutrinos = [12, 14, 16]
+
+cm2mm = 10.0
+nentries = 0
+
+if (inputfilename[0:4] == "/eos"):
+    inputfilename = os.environ["EOSSHIP"] + inputfilename
```
olantwin: No. This is banned. I've cleaned it up in many places. Please remove it. 

This breaks for files on cernbox and teaches users bad habits. 
If users are on lxplus, paths to eos are automatically upgraded to the xrootd protocol, everywhere else users should give valid remote paths.

olantwin (changes requested):

macro/convert_nu_flux_ntuple.py
antonioiuliano2: Added a commit with the same change in muonShieldOptimization/extractNeutrinosAndUpdateWeight.py and muonShieldOptimization/extractMuonsAndUpdateWeight.py

olantwin:
Could you please rebase to resolve the merge conflict?

Then I'll give it a final review, but I think it should be good for merge.

macro/convert_nu_flux_ntuple.py
```diff
@@ -0,0 +1,162 @@
...
+        emax = max(emax, e)
+        entry["pdg"] = pdg
+        entry["vx"] = v.GetX() * cm2mm  # mm
+        entry["vy"] = v.GetY() * cm2mm  # mm
+        entry["vz"] = v.GetZ() * cm2mm  # mm
+        entry["t"] = v.GetTime()  # ns
+        entry["px"] = t.GetPx()  # GeV
+        entry["py"] = t.GetPy()
+        entry["pz"] = t.GetPz()
+        entry["weight"] = 1.0
```
THanae, line 128: Will weight always be 1 or could there be instances in the future when it isn't? If the latter, would it be safer to use `t.GetWeight`?

olantwin: Support for weights is now being added in Aegir, so we'll probably also want to load them here.

olantwin:
@antonioiuliano2 Could you please resolve the merge conflicts?

antonioiuliano2:
My apologies, I noticed that this PR is still open.
I have resumed by resolving the CHANGELOG conflicts.
I will now have a look at the remaning coderabbit issue.

CHANGELOG.md
```diff
@@ -151,6 +154,8 @@ random sequence.
 
 ### Fixed
 
+* Remove duplicated 4232 pdg code in extractNeutrinosAndUpdateWeight and extractMuonsAndUpdateWeight
```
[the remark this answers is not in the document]
antonioiuliano2: Good Point! I shall also do the same as extractNeutrinosAndUpdateWeight and replace the duplicated 4232 with 4322, as done in commit [00096c7](https://github.com/ShipSoft/FairShip/commit/00096c7a1197dd0b45f982e099c37051e5f30375)

PR #1345, opened 2026-07-20
Introducing regression tests

kholoimov:
This PR introduces a CTest-based regression testing framework for FairShip.

  Main features:

  - Adds eight regression tests covering simulation, event generation, muon-background workflows, reconstruction, particle-gun output, and tracking performance.
  - Compares command output against committed reference files.
  - Filters known nondeterministic output using configurable skip patterns.
  - Supports dependencies and shared working directories between related tests.
  - Runs independent tests in parallel.
  - Adds commands to run the regression suite and regenerate reference outputs.
  - Expands testing documentation and clarifies existing C++/RNTuple test descriptions.

  Run the tests with:

`  pixi run test-fairship`

  Regenerate references with:

`  pixi run regenerate-fairship-references`

## Summary by CodeRabbit

* **New Features**
  * Added dependency-aware C++ and simulation regression testing.
  * Added commands to run test suites and regenerate reference outputs.
  * Added options to retain test output for inspection.
  * Simulation generators now support consistent, configurable random seeds for reproducible results.

* **Documentation**
  * Added guides for running, configuring, and extending tests, including network requirements.

* **Bug Fixes**
  * Improved regression-output validation and corrected RNTuple-related messaging.


kholoimov:
Going to apply coderabbit comments

tests/fairship_tests/scripts/reconstruction.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# SPDX-FileCopyrightText: Copyright CERN for the benefit of the SHiP Collaboration
+
+set -euo pipefail
+
+python "$FAIRSHIP_ROOT/macro/ShipReco.py" \
```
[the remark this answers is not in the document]
kholoimov: It should not be needed as the same environment variable should be propagated from the simulation test, which is in the dependency list of the reconstruction test.

kholoimov:
@coderabbitai review

olantwin:
@kholoimov Could you please get the CI to pass? I think most jobs fail because the lockfile needs updating due to the added env.

olantwin:
No merge commit please.

kholoimov:
It's empty. Have no idea why github decided that I need to update my branch

kholoimov:
@olantwin I believe now it's failing not because of my changes. I can fix it in the PR, but probably will be better to have it separately:
```

[vkholoim@hl01 FairShip]$ pixi install --locked
Error:   × failed to fetch geant4-vmc-6.8-hb0f4dca_6.conda
  ├─▶ failed to interact with the package cache layer.
  ╰─▶ HTTP status client error (404 Not Found) for url (https://prefix.dev/ship/linux-64/geant4-vmc-6.8-hb0f4dca_6.conda)
```

olantwin:
> @olantwin I believe now it's failing not because of my changes. I can fix it in the PR, but probably will be better to have it separately:
> 
> ```
> 
> [vkholoim@hl01 FairShip]$ pixi install --locked
> Error:   × failed to fetch geant4-vmc-6.8-hb0f4dca_6.conda
>   ├─▶ failed to interact with the package cache layer.
>   ╰─▶ HTTP status client error (404 Not Found) for url (https://prefix.dev/ship/linux-64/geant4-vmc-6.8-hb0f4dca_6.conda)
> ```

This is indeed my fault due to geant4-vmc moving to conda-forge.

olantwin:
#1348 should fix the geant4-vmc issue.

olantwin:
Rebasing now should work.

kholoimov:
I faced the problem that even with Seed propagation inside Pythia8 and EvtGen generators, I see different output on different machines, while the output on the same machine stays the same.

kholoimov:
> I faced the problem that even with Seed propagation inside Pythia8 and EvtGen generators, I see different output on different machines, while the output on the same machine stays the same.

Looks like the issue is that my second machine utilises an AMD processor. Going to validate tests on lxplus, and if everything goes well, this is ready to be merged

kholoimov:
@coderabbitai review

olantwin:
The seeding code should be moved to SHiP::Generator to reduce duplication (and you missed some generators!).

kholoimov:
> The seeding code should be moved to SHiP::Generator to reduce duplication (and you missed some generators!).

Fixed. There are probably still some non-deterministic parts of the code, but I believe we'll address them later, once we have more tests covering those areas.

kholoimov:
Also, this implementation is technically independent of FairShip, so once we converge on a final version, I can easily port it to the new framework as well.

olantwin:
Could you please resolve the merge conflict and request a coderabbit review. I think this is ready for merge?

olantwin:
@coderabbitai resume

kholoimov:
not ready for now, I have to regenerate references at the lxplus. for now I accidentaly pushed the references from my extra cluster, which is Intel, meanwhile lxplus is AMD by default

shipgen/tPythia6Generator.cxx
```diff
@@ -25,7 +25,7 @@ tPythia6Generator::tPythia6Generator() {
 
 // -----   Default constructor   -------------------------------------------
 Bool_t tPythia6Generator::Init() {
-  fPythia->SetMRPY(1, gRandom->GetSeed());
+  fPythia->SetMRPY(1, GetSeed());
```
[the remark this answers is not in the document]
kholoimov: Applied

kholoimov:
@coderabbitai resume

kholoimov:
@olantwin it's ready now

olantwin:
@coderabbitai full review

tests/fairship_tests/scripts/muonback-io.sh
```diff
@@ -0,0 +1,22 @@
...
+export QT_QPA_PLATFORM=offscreen
+
+python "$FAIRSHIP_ROOT/macro/run_simScript.py" \
+    -n 100 \
+    -i 100 \
+    -f https://cernbox.cern.ch/remote.php/dav/public-files/vdwtXtgM5P2Z0S5/pythia8_Geant4_10.0_withCharmandBeauty0_mu.root \
+    --remote-input \
+    --tag muonback_io \
+    --sameSeed 42 \
+    --seed 42 \
```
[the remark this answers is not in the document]
kholoimov: It's intended to use a remote file as input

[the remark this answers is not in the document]
kholoimov: I don’t think this check is needed at this stage. The remote input is intentionally used directly so the test does not need to download and store a local copy, even temporarily for checksum validation. If immutable input becomes a requirement, we can move the file to a versioned artifact location instead.

shipgen/TEvtGenDecayer.cxx
```diff
@@ -64,12 +65,13 @@ void TEvtGenDecayer::Init() {
...
+    const UInt_t seed = fSeed != 0 ? fSeed : gRandom->GetSeed();
+    const UInt_t pythiaSeed = SHiP::NormalizePythiaSeed(seed);
 
     // Also seed Pythia8 with the same seed for consistency
     TPythia8* pythia8 = TPythia8::Instance();
     if (pythia8) {
-      pythia8->Pythia8()->readString(Form("Random:seed = %u", seed));
+      pythia8->Pythia8()->readString(Form("Random:seed = %u", pythiaSeed));
       pythia8->Pythia8()->readString("Random:setSeed = on");
       pythia8->Pythia8()->init();  // Re-initialize with new seed
```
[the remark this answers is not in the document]
kholoimov: The installed ROOT 6.40.02 TPythia8Decayer owns a private TPythia8 instance and exposes neither that instance nor a seed setter. Its Init() method is empty. Therefore, moving the current code and re-calling fPythia8Decayer->Init() would still not seed the Pythia instance used by Decay().

  A real fix requires a small FairShip-owned decayer implementation or an upstream ROOT API change. The current code still seeds the unrelated TPythia8::Instance() singleton.

kholoimov:
@olantwin i regenerated and validated references yesterday. it is ready. 

the biggest problem for now: implementation and tests are controlled by CPU manufacturers. Results are still different between AMD and Intel CPUs

For now references are generated for AMD CPU, as this is what used by default on lxplus.

olantwin:
If we run the tests on the self-hosted runners (which are currently idle), we might be able to have reproducible results. Unfortunately GHA runners run on a mix of intel and AMD.

olantwin:
Alternatively, we'd need to compare numbers with some tolerance for numerical effects (which Geant4 unfortunately tends to amplify beyond the usual floating-point inaccuracies...)

kholoimov:
> Alternatively, we'd need to compare numbers with some tolerance for numerical effects (which Geant4 unfortunately tends to amplify beyond the usual floating-point inaccuracies...)

this is an example of the difference for the tracking benchmark. The difference is stable and stays the same for different Intel machines.

```
+++ tracking-benchmark (current output)
@@ -527,14 +527,14 @@
   Efficiency:       0.9980 +/- 0.0015
   Clone rate:       0.0000 +/- 0.0005
   Ghost rate:       0.0020 +/- 0.0015
-  dp/p sigma:       0.012636 +/- 0.000369
-  dx RMS:           0.2014 +/- 0.0045 cm
+  dp/p sigma:       0.012378 +/- 0.000359
+  dx RMS:           0.1736 +/- 0.0039 cm
   dy RMS:           0.0051 +/- 0.0001 cm
-  dtx RMS:          0.000224 +/- 0.000005
-  dty RMS:          0.000060 +/- 0.000001
-  Hit resid RMS:    0.036651 +/- 0.000197 cm
-  Hit pull mean:    -0.0551 +/- 0.0091
-  Hit pull sigma:   1.1407 +/- 0.0071
+  dtx RMS:          0.000247 +/- 0.000006
+  dty RMS:          0.000070 +/- 0.000002
+  Hit resid RMS:    0.036586 +/- 0.000196 cm
+  Hit pull mean:    -0.0560 +/- 0.0091
+  Hit pull sigma:   1.1401 +/- 0.0070
   Charge-ID eff:    1.0000 +/- 0.0005
   Correct/wrong:    998/0
 ==================================
```

kholoimov:
> Alternatively, we'd need to compare numbers with some tolerance for numerical effects (which Geant4 unfortunately tends to amplify beyond the usual floating-point inaccuracies...)

For pythia8 generator test it even changes the number of tracks generated in an event:

```
+++ pythia8-summary (current output)
@@ -768,18 +768,18 @@
   Countable branches           : 5

 [MCTrack Validation]
-  Total tracks                 : 14868
+  Total tracks                 : 14086
   Primary tracks               : 100
-  Secondary tracks             : 14768
-  Tracks per event             : 148.7
-  Track momentum               : 3.828 +- 29.68  [min=0.05956, max=400]
-  Track energy                 : 3.922 +- 29.67  [min=0.1, max=400]
-  Track start Z                : 1.029e+04 +- 2593  [min=0, max=1.605e+04]
-  Top PDG counts               : 22:4494, 11:2713, -11:2187, 2112:1305, 2212:1201, 111:820, 211:703, -211:698, -13:118, 9900015:100
+  Secondary tracks             : 13986
+  Tracks per event             : 140.9
+  Track momentum               : 4 +- 30.48  [min=0.05971, max=400]
+  Track energy                 : 4.103 +- 30.47  [min=0.1, max=400]
+  Track start Z                : 1.033e+04 +- 2701  [min=0, max=1.605e+04]
+  Top PDG counts               : 22:3930, 11:2439, -11:1861, 2212:1365, 2112:1327, 111:858, 211:807, -211:755, -13:119, 9900015:100
```

tests/fairship_tests/harness.py
```diff
@@ -0,0 +1,324 @@
...
+    expression = ".*".join(re.escape(part) for part in pattern.split("*"))
+    return re.compile(expression)
+
+
+def _patterns_for(test_name: str) -> list[re.Pattern[str]]:
+    known_tests = set(test_names())
+    patterns: list[str] = []
+    current_selector: str | None = None
+    for line_number, raw_line in enumerate(SKIP_PATTERNS.read_text(encoding="utf-8").splitlines(), start=1):
+        line = raw_line.strip()
```
olantwin, line 167: This should strip leading whitespace...

tests/fairship_tests/harness.py
```diff
@@ -0,0 +1,324 @@
...
+        stdout = result.stdout
+    except subprocess.TimeoutExpired as error:
+        returncode = 124
+        stdout = error.stdout or ""
+        if isinstance(stdout, bytes):
+            stdout = stdout.decode(errors="replace")
+        stdout += f"\nTest timed out after {TEST_TIMEOUT_SECONDS} seconds\n"
+    patterns = _patterns_for(test_name)
+    lines = stdout.splitlines(keepends=True)
+    output = "".join(line for line in lines if not any(pattern.fullmatch(line.rstrip("\r\n")) for pattern in patterns))
```
olantwin: ...while this only strips the carriage return, so that leading whitespace is intact. If I read this correctly, indented lines won't match for any patterns anchored to the start of the line.

pixi.toml
```diff
@@ -29,8 +29,6 @@ numpy = "*"
 scipy = "*"
 matplotlib = "*"
 pandas = "*"
-pytest = "*"
-pytest-xdist = "*"
```
olantwin, line 33: Please drop this change, we now have some first pytest unit tests.

shipgen/TEvtGenDecayer.cxx
```diff
@@ -64,12 +65,13 @@ void TEvtGenDecayer::Init() {
...
       LOG(debug) << "TEvtGenDecayer: Pythia8 RNG seeded with " << seed;
```
olantwin: This should report the new pythiaSeed instead of the unnormalised seed

olantwin:
I see several potential ways forward for the architecture-dependence:
- use self-hosted runners we control (to be verified)
- limit ourselves to things that are deterministic (e.g. analysis of known samples) and give up on event gen and simulation
- if we have only 2 versions (AMD and Intel), we could potentially just store both. If we have more than that, this approach doesn't scale.

tests/fairship_tests/harness.py
```diff
@@ -0,0 +1,324 @@
...
+        location = f"{TEST_CASES} test #{index}"
+        if isinstance(configured_test, dict) and set(configured_test) == {"name"}:
+            name = configured_test["name"]
+            dependencies = []
+        elif isinstance(configured_test, dict) and set(configured_test) == {"name", "depends_on"}:
+            name = configured_test["name"]
+            dependencies = configured_test["depends_on"]
+        else:
+            raise RuntimeError(
+                f"{location} must be a test name or a mapping containing 'name' and optional 'depends_on'"
```
olantwin: This is not quite clear, if I read things correctly, `{"name"}` is fine, but `"name"` is not, which is how the message could be understood. I suggest to rephrase.

gconfig/DecayConfigTEvtGen.C
```diff
@@ -9,6 +9,14 @@ void DecayConfig() {
   // Create the custom EvtGen decayer
   TEvtGenDecayer* decayer = new TEvtGenDecayer();
 
+  // run_simScript exports the resolved command-line seed before VMC loads this
+  // macro. Pass it explicitly so initialization cannot advance the ROOT RNG
+  // and silently change the EvtGen/Pythia8-decayer seed.
+  TString RandomSeed = gSystem->Getenv("FAIRSHIP_RANDOM_SEED");
+  if (!RandomSeed.IsNull()) {
+    decayer->SetSeed(static_cast<UInt_t>(RandomSeed.Atoll()));
```
olantwin: Is this the right size? I'm concerned that ll and uint_t have different numerical limits.

olantwin (changes requested):

kholoimov:
> * if we have only 2 versions (AMD and Intel), we could potentially just store both. If we have more than that, this approach doesn't scale.

For now, this is what I see on the machines I have access to. The result will also probably differ on ARM machines.

tests/fairship_tests/harness.py
kholoimov: Ah yes. Thanks for noticing!

gconfig/DecayConfigTEvtGen.C
kholoimov: in `run_simScript` seed is generated within limits 

```
  seed = options.theSeed
  if seed == 0:
      ROOT.gRandom.SetSeed(0)
      seed = ROOT.gRandom.GetSeed()

  if seed > 900000000:
      seed = seed % 900000000
```

this is in range of both long long and uint_t

olantwin: Ok, fine.

By the way, I noticed that you use UInt_t here, please use std::uint32_t instead.

olantwin: Also, since we are <900000000, why not use Atoi?

olantwin:
For now, we don't have ARM runners, so that's fine. But we can't choose between AMD and Intel.

gconfig/DecayConfigTEvtGen.C
kholoimov: done

PR #1350, opened 2026-07-25
perf: replace ShipStack track maps with dense vectors

olantwin:
## Motivation

Investigating memory growth in `run_fixedTarget.py` with high `--kaon-pion-splits` levels showed that RSS ratchets with the largest event seen (no true leak), at ~300 B per stack particle. About a third of that, and a surprisingly large share of CPU time, came from `ShipStack`'s two per-event `std::map`s (`fStoreMap`, `fIndexMap`): with per-step splitting, events reach millions of particles, so the maps cost millions of O(log n) lookups and ~96 B/particle of node churn that fragments the allocator arenas.

## Change

Both maps are keyed by track indices, which are contiguous in `[0, fNParticles)`, so they are replaced by vectors indexed by particle id:

- `fStoreMap` → `std::vector<Bool_t> fStoreFlags`
- `fIndexMap` → `std::vector<Int_t>` (value `-2` = track not stored, unchanged semantics)
- the former `fIndexMap[-1] = -1` sentinel for primary mothers becomes an explicit `< 0` branch at the lookup sites
- `ISTLPointContainer::UpdatePointTrackIndices` takes `const std::vector<Int_t>&`; the sole implementer (`SHiP::Detector<PointType>`) does a bounds-checked direct index
- the "particle not found in storage map" fatal is dropped (impossible by construction); the out-of-range fatals in `UpdateTrackIndex` are kept

## Results

400-event benchmark (`run_fixedTarget.py -n 400 --kaon-pion-splits 100 --multiple-kpi-splits -rs 1234`):

- CPU time: 478 s → 249 s (~1.9×)
- RSS at the largest event (3.35M particles): 2558 MB → 2278 MB (~84 B/particle)

## Verification

- Simulation output is identical: a 50-event before/after comparison (same seed) matches event by event in all MCTrack PDG codes, mother IDs, momenta and weights, and all PlaneHAPoint track IDs
- ctest (DataClassIO, RNtupleIO), `ci-fixed-target` and `ci-sim` pass

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved track selection and index remapping performance for high-multiplicity events.
  * Reduced CPU usage and peak memory consumption during particle splitting and track processing.
* **Bug Fixes**
  * Improved handling of track and detector-hit index updates, including unmapped and out-of-range track identifiers.
* **Documentation**
  * Added changelog documentation describing the performance and memory improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ it in future.
 
 ### Changed
 
+* Use dense vectors instead of `std::map` for ShipStack track selection and index remapping, roughly halving CPU time and reducing peak memory for high-multiplicity events (e.g. kaon/pion splitting)
+
 ### Fixed
 
 ### Removed
--- a/Detector/Detector.h
+++ b/Detector/Detector.h
@@ -57,13 +57,11 @@ class Detector : public FairDetector, public ISTLPointContainer {
 
   TClonesArray* GetCollection(Int_t iColl) const override { return nullptr; }
 
-  void UpdatePointTrackIndices(
-      const std::map<Int_t, Int_t>& indexMap) override {
+  void UpdatePointTrackIndices(const std::vector<Int_t>& indexMap) override {
     for (auto& point : *fDetPoints) {
       Int_t oldTrackID = point.GetTrackID();
-      auto iter = indexMap.find(oldTrackID);
-      if (iter != indexMap.end()) {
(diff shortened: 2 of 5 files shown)


olantwin:
FYI @THanae, this might speed up splitting quite a bit for large events/events with many splits.

olantwin:
@coderabbitai review

PR #1351, opened 2026-07-25
perf: per-track MC point counts and early-exit mother flagging in ShipStack

olantwin:
Stacked on #1350 (base branch `shipstack-dense-index-maps`; retarget to `main` once #1350 merges).

Two further reductions of per-particle cost in the `ShipStack` track-selection path, motivated by the same kaon/pion-splitting profiling as #1350 (events with millions of stack particles):

1. **Count MC points per track instead of per (track, detector).** `fPointsMap` was keyed by (track index, detector id), but the only consumer, `SelectTracks`, sums over all detectors per track for the `fMinPoints` cut — the per-detector breakdown is never read anywhere (no accessor exists in ShipStack or FairGenericStack). The map is replaced by a vector indexed by track, removing 21 map probes per stack particle. The `AddPoint(DetectorId)` detector-facing interface is unchanged (10 call sites untouched).
2. **Stop the mother-flagging walk at already-flagged ancestors.** Mothers always have smaller indices than daughters and every walk runs to the root, so an already-flagged mother implies its whole chain is flagged; breaking there avoids re-walking shared ancestor chains.

## Results

400-event benchmark (`run_fixedTarget.py -n 400 --kaon-pion-splits 100 --multiple-kpi-splits -rs 1234`), on top of #1350:

- CPU time: 249 s → 195 s (478 s before #1350)
- RSS unchanged (the point-count map held only ~10³ entries per event; this is a CPU fix)

## Verification

- Simulation output identical to the pre-#1350 baseline: 50-event comparison (same seed) matches event by event in all MCTrack PDG codes, mother IDs, momenta and weights, and all PlaneHAPoint track IDs
- ctest (DataClassIO, RNtupleIO) and `ci-fixed-target` pass

## Summary by CodeRabbit

* **Performance**
  * Improved track selection efficiency for high-multiplicity events by counting MC points per track across detectors.
  * Reduced CPU usage during mother-track flag propagation by stopping at already-flagged ancestors.
* **Documentation**
  * Updated the unreleased changelog with these performance improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ it in future.
 * Use dense vectors instead of `std::map` for ShipStack track selection and index remapping, roughly halving CPU time and reducing peak memory for high-multiplicity events (e.g. kaon/pion splitting)
 * Deduplicate the charm/beauty over min-bias cross-section scaling shared by `makeDecay` and `run_fixedTarget` into `python/heavyFlavourScaling.py`
 * Make `--chicc` and `--chibb` mutually exclusive in `makeDecay` and `run_fixedTarget`, and raise a clear error when the override does not match the run type (e.g. `--chicc` for a beauty run) instead of silently ignoring it
+* Count ShipStack MC points per track instead of per (track, detector) and stop the mother-flagging walk at already-flagged ancestors, further reducing track-selection CPU time for high-multiplicity events
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


olantwin:
@coderabbitai review

THanae (dismissed):

olantwin:
@coderabbitai review

PR #1341, opened 2026-07-14
Update Charm and Beauty over MBias cross sections ration in run_fixedTarget

antonioiuliano2:
PR related to issue #1265.

Basically, when we added the target composition in makeDecay.py, we neglected to modify run_fixedTarget.py.
It follows the same computation to normalize the simulation to the number of protons on target.
Therefore, the same scaling to target material (defaulting to tungsten) was applied here.

All the changes are applied in the python script, without affecting either methods or members of the FixedTargetGenerator generator class.

Best Regards,
Antonio

## Checklist

- [ ] Changelog updated (if applicable)
- [ ] Commit message follows [conventional commits](https://www.conventionalcommits.org/)
- [ ] CI checks pass

## Summary by CodeRabbit

* **New Features**
  * Added target composition presets (tungsten/molybdenum) plus a command-line option to override the target mass number.
  * Charm and beauty cross sections for fixed-target runs are now derived from the selected target and scaling model.
  * Charm/beauty cross-section CLI overrides are now optional (only applied when provided) and the derived configuration is reported during runtime.

* **Bug Fixes**
  * Updated charm/beauty cross-section handling for fixed-target runs over minimum-bias events.

* **Documentation / Chores**
  * Updated the changelog entry under “Unreleased” for the fixed-target update.
  * Extended type stubs for the fixed-target generator’s cross-section setters.

PR #1357, opened 2026-07-29
refactor: share heavy-flavour cross-section scaling between macros

olantwin:
Follow-up to #1341.

The target-composition scaling of the charm/beauty over min-bias cross-section ratios (`chicc`/`chibb`) was duplicated verbatim between `makeDecay.py` and `run_fixedTarget.py`, so a future physics update (reference cross sections or the `A`-scaling exponents) would have to be applied in two places and could silently drift.

This extracts the constants and the derivation into a new shared module `python/heavyFlavourScaling.py` (`derive_cross_sections` + `format_summary`) and rewires both macros to use it. A small pytest (`tests/test_heavyFlavourScaling.py`) covers the presets, overrides and the `A > 0` guard.

No behaviour change: the default (W) and Mo presets, explicit `--chicc`/`--chibb`/`-A` overrides, and the printed summary are all preserved.

A separate follow-up will make `--chicc`/`--chibb` mutually exclusive and reject the wrong one for the run type.

## Checklist

- [ ] CI checks pass

## Summary by CodeRabbit

- **New Features**
  - Added consistent charm and beauty cross-section scaling across decay and fixed-target workflows.
  - Supports target presets, explicit target mass values, and manual cross-section overrides.
  - Improved summary reporting, including clearer handling when an explicit target mass is provided.

- **Bug Fixes**
  - Added validation for missing or invalid target-mass inputs.

- **Tests**
  - Added coverage for presets, scaling, overrides, and input validation.

### DIFF
--- a/python/heavyFlavourScaling.py
+++ b/python/heavyFlavourScaling.py
@@ -22,13 +22,17 @@ TARGET_A = {"W": 184.0, "Mo": 98.0}
 CrossSections = namedtuple("CrossSections", ["chicc", "chibb", "A", "scale"])
 
 
-def derive_cross_sections(target_composition, A=None, chicc=None, chibb=None):
+def derive_cross_sections(target_composition=None, A=None, chicc=None, chibb=None):
     """Return default-scaled chicc/chibb, honouring explicit overrides.
 
     ``A`` overrides the ``target_composition`` preset; ``chicc``/``chibb``
-    override the target-derived value when not ``None``.
+    override the target-derived value when not ``None``. Supply either
+    ``target_composition`` (a known preset) or an explicit ``A``.
     """
-    A = A if A is not None else TARGET_A[target_composition]
+    if A is None:
+        if target_composition is None:
+            raise ValueError(f"Provide either target_composition (one of {sorted(TARGET_A)}) or an explicit A.")
+        A = TARGET_A[target_composition]
     if A <= 0:
         raise ValueError(f"Invalid target mass number A={A}. Must be > 0.")
     scale = (A / A_REF) ** (heavyflavour_Ascale - mbias_Ascale)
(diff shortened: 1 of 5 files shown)


python/heavyFlavourScaling.py
```diff
@@ -0,0 +1,46 @@
...
+
+A_REF = 98.0
+CHICC_REF = 1.7e-3  # prob to produce primary ccbar pair/pot on Mo
+CHIBB_REF = 1.6e-7  # prob to produce primary bbbar pair/pot on Mo
+TARGET_A = {"W": 184.0, "Mo": 98.0}
+
+CrossSections = namedtuple("CrossSections", ["chicc", "chibb", "A", "scale"])
+
+
+def derive_cross_sections(target_composition, A=None, chicc=None, chibb=None):
```
antonioiuliano2: Since A overrides target_composition, should not the latter be an optional argument too?
Example (hypotetical simulation in tantalum):
`derive_cross_sections(A=181)`
As the code is now, it would return error due to target_composition argument not provided, even if in this case it would not be used at all.

tests/test_heavyFlavourScaling.py
```diff
@@ -0,0 +1,44 @@
...
+def test_w_preset_scales_up():
+    cs = derive_cross_sections("W")
+    assert cs.A == 184.0
+    assert cs.scale == pytest.approx((184.0 / 98.0) ** 0.29)
+    assert cs.chicc == pytest.approx(CHICC_REF * cs.scale)
+    assert cs.chibb == pytest.approx(CHIBB_REF * cs.scale)
+
+
+def test_explicit_A_overrides_preset():
+    cs = derive_cross_sections("W", A=98.0)
```
antonioiuliano2, line 38: See my comment above, the function is perfectly fine, but the "W" argument here is misleading.

tests/test_heavyFlavourScaling.py
```diff
@@ -0,0 +1,44 @@
...
+    assert cs.A == 98.0
+    assert cs.scale == pytest.approx(1.0)
+
+
+def test_explicit_ratios_override_derived():
+    cs = derive_cross_sections("W", chicc=5e-3, chibb=2e-7)
```
antonioiuliano2: Same as above

antonioiuliano2 (dismissed):
Thanks for this PR, 
it removes unneeded code duplication which we now from past experiences it easily leads to divergences in the future.
I just have a small request concerning the target_composition required argument.

olantwin:
@antonioiuliano2 Thanks for your review, I've now made the composition optional if A is given (and check that at least one of those is given), and have adjusted the console output to not give the wrong material when the target composition is overridden.

tests/test_heavyFlavourScaling.py
olantwin: Now that the composition is optional, the A-only path is tested in `test_A_without_target_composition`; this test deliberately passes both to check that an explicit `A` takes precedence over the `"W"` preset — I've added a docstring to make that intent explicit.

THanae (approved):

PR #1358, opened 2026-07-29
fix: reject mismatched chicc/chibb overrides in charm/beauty macros

olantwin:
Follow-up to #1341. **Stacked on #1357** (based on that branch); rebase onto `main` once #1357 merges.

A run is either charm or beauty, so only one of the ccbar/bbbar over min-bias cross-section ratios is ever meaningful. Previously the wrong override was silently ignored:

- in `run_fixedTarget.py`, passing `--chicc` for a beauty run had no effect because `FixedTargetGenerator::InitForCharmOrBeauty` unconditionally sets `chicc = chibb` for beauty;
- in `makeDecay.py`, the `setByHand` flag caused a user-supplied `--chicc` to be used even on a beauty file.

This makes `--chicc`/`--chibb` mutually exclusive (argparse) and adds `check_run_type_override` to the shared `heavyFlavourScaling` module, which raises a clear error when the supplied override does not match the run type:

- `run_fixedTarget.py`: checked against the resolved `--charm`/`--beauty` flag;
- `makeDecay.py`: checked after the input file's charm-vs-beauty type is known (`M > 5`), replacing the silent `setByHand` swap.

Unit tests cover the accepted and rejected combinations.

## Checklist

- [ ] CI checks pass

### DIFF
--- a/macro/makeDecay.py
+++ b/macro/makeDecay.py
@@ -8,7 +8,11 @@ import os
 
 import ROOT
 import rootUtils as ut
-from heavyFlavourScaling import derive_cross_sections, format_summary
+from heavyFlavourScaling import (
+    check_run_type_override,
+    derive_cross_sections,
+    format_summary,
+)
 
 ROOT.gROOT.LoadMacro("$VMCWORKDIR/gconfig/basiclibs.C")
 ROOT.basiclibs()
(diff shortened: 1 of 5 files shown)


antonioiuliano2 (dismissed):
Great!
This shall avoid any unwanted overlap between the charm and beauty run options.

macro/makeDecay.py
```diff
@@ -128,7 +133,9 @@
     rc = sTree.GetEvent(n)
     # check if we deal with charm or beauty:
     if n == 0:
-        if not setByHand and sTree.M > 5:
+        is_beauty = sTree.M > 5
+        check_run_type_override(is_beauty, args.chicc, args.chibb)
```
THanae, line 137: Should it be `check_run_type_override(is_beauty, chicc, chibb)` instead?

THanae: Ah no never mind, one can also access with `args`

THanae (dismissed):

THanae (approved):

Issue #1360, opened 2026-07-29
run_fixedTarget: relative --TARGET_YAML path fails (cwd changes before config load)


olantwin:
`macro/run_fixedTarget.py` changes into its run output directory (`<host>_run_fixedTarget_<r>/`) before `geometry_config.create_config()` opens the target YAML, so a relative `--TARGET_YAML` path that is valid from the invocation directory fails:

```
python macro/run_fixedTarget.py -n 10 -e 10.0 -r 1 --TARGET_YAML geometry/target_config.yaml
...
FileNotFoundError: [Errno 2] No such file or directory: 'geometry/target_config.yaml'
```

`run_simScript.py --target-yaml` accepts the same relative path, so the inconsistency is surprising; an absolute path works. Suggest resolving the path at argument-parsing time (e.g. `os.path.abspath` / `type=lambda p: Path(p).resolve()`).

Issue #1359, opened 2026-07-29
FixedTargetGenerator: TGeo fallback target-extent extraction uses wrong node and array size


olantwin:
The fallback path in `FixedTargetGenerator::Init()` that derives the target z extent from the geometry (used only when `SetTargetCoordinates()` was not called) has two bugs:

https://github.com/ShipSoft/FairShip/blob/0979da44a59dfd86ea1a0a3fe4c09279b86edde4/shipgen/FixedTargetGenerator.cxx#L330-L346

1. The end position is computed from the *first* node's shape instead of the last one:
   ```cpp
   nav->cd(targetName + "/" + last->GetName());
   sha = static_cast<TGeoBBox*>(first->GetVolume()->GetShape());  // should be last
   ```
   The local-to-master conversion uses the correct node, but `dz` comes from the wrong shape, so `endZ` is wrong whenever the first and last daughters of `HeVolume` have different half-lengths.

2. `nodes->GetSize()` returns the allocated capacity of the `TObjArray`, not the number of entries, so `nodes->At(nodes->GetSize() - 1)` can return `nullptr` (subsequent dereference crashes) or a stale entry. Should be `GetEntriesFast() - 1` (or `GetLast()`).

Not triggered by the standard chains (`run_fixedTarget.py`/`run_simScript.py` always pass target coordinates from python), but the fallback silently miscomputes the material-budget range when used.

PR #1361, opened 2026-07-29
feat(target): implement 2026 BDF target design

### LINKED ISSUE #1356: Implement new target design
Available here in CAD: [ ST1A07710_01 AB.02 ENC3 BDF TARGET + PLUG IN TABLE](https://plm.cern.ch/PLM/?StartItem=CAD:7B9A7B66C61F44AC9FD2FB2F628901BF).

olantwin:
## Summary

Implements the 2026 BDF/SHiP target design, extracted from CATIA model `ST1A07710_01_AB.02`, as a selectable target design. The legacy design remains the default and is byte-identical to before, so the two can be compared directly.

- `ShipTargetStation` gains a design version (`SetDesign`), driven by a new `version` key in the target YAML:
  - **version 1** (default): legacy 19 Ta-clad disk target, unchanged
  - **version 2**: the 2026 BDF target:
    - 33 solid tungsten disks (30 to 455 mm thick, Ø250 mm; larger Ø314 mm rear block), separated by 4–5 mm He slits, total length 146 cm
    - steel core (the two clamp halves modelled as one axisymmetric composite) with stepped bore (r 125 → 157 mm) and outer profile (r 207 → 190 mm), including the serpentine He cooling grooves (61° arcs r 125→153 mm, staggered between the upper and lower half, plus the rear groove r 157→182 mm)
    - jacket tube (r 217–225 mm) with front and rear flanges; He annulus between core and jacket
    - domed rear endcap (8 mm shell: cylinder + three cone segments + apex disc)
- New `geometry/target_config_2026.yaml`; select with `--target-yaml geometry/target_config_2026.yaml`
- Downstream elements (proximity shielding, hadron absorber, muon shield) are positioned from the nominal legacy target length so they stay identical for both designs

## Design

Cross-sections of the implemented (simplified) geometry; z = 0 at the front face of the first disk, dimensions in mm.

### x–y section (z ≈ 420 mm)
x-y section

### z–x section (horizontal plane, no grooves)
z-x section

### z–y section (vertical plane, staggered cooling grooves)
z-y section

Simplifications relative to the CAD model: bolt lugs, key-ways, pipe stubs, fasteners and survey hardware omitted; the two core halves are merged into one axisymmetric piece; the endcap dome is approximated by three cones; the flange/jacket boundary is placed at the end of the core front step (55.2 mm instead of the CAD's stepped flange bore at 64.2 mm).

## Verification

- **Backward compatibility**: `getGeoInformation --level 2` dump of the default (legacy) geometry is identical before/after this change; `ci-fixed-target` and `ctest` pass.
- **New design**: overlap check clean at 0.1 µm in the target region; all 33 disk z-positions and radii verified against the CAD solids; material probes confirm groove staggering, He annulus, flange bore and endcap; material maps rendered from the geofile match the drawings above; `run_fixedTarget.py` works with the new YAML (target extent 0–146 cm).
- Top-level detector positions (muon shield etc.) are identical between the two designs.

Pre-existing issues found while testing: #1359, #1360.

The drawings are served from the assets-only branch `target-2026-assets`, which is not meant to be merged.

## Review fixes

Follow-up commits addressing code-review findings (no functional change to either design):

- **Single source for the shielding reference length**: the fixed length that keeps the proximity shielding independent of the actual target length was hardcoded twice (`158.64 cm` in the Python config, `1586.4 mm` in `ShipTargetStation`). It now lives once as `target.length_fixed` in the geometry config and is passed to `ShipTargetStation` via `SetShieldingReferenceLength`; the C++ side aborts if it is left unset. `hadronAbsorber.z` placement is unchanged.
- **Reject unknown target designs**: the Design 2 branch was a bare `else`, so any `fDesign` other than 1 silently built the 2026 target. It is now guarded by an explicit `fDesign == 2` check with a fatal error naming the unrecognized value.
- **Type stubs**: added the missing `ShipTargetStation` setters (`SetDesign`, `SetLastDiskDiameter`, `SetShieldingReferenceLength`) to the ROOT stub, clearing the `pyrefly` missing-attribute errors.

## Summary by CodeRabbit

* **New Features**
  * Added support for the 2026 BDF/SHiP target design, including tungsten disks, helium cooling, and an updated steel enclosure.
  * Added selectable target configurations, with the legacy design remaining the default.
  * Added support for larger rear-disk dimensions in the 2026 design.

* **Improvements**
  * Maintained consistent downstream geometry alignment when switching target designs.
  * Preserved the nominal legacy target length for direct downstream comparisons.
  * Documented how to enable the 2026 target configuration.


olantwin:
The performance impact of the cone-approximated back of the vessel needs to be looked at. If it's prohibitive, we might want to simplify further.

olantwin:
In ægir, no negative impact on simulation performance is measurable (in some cases even positive impact, albeit not with Geant4 settings that are relevant for us), so there is no need to simplify or remove the end-cap.

Geometries with poly-cone end-cap and without an end-cap were compared, using:
- a particle gun right upstream of the end-cap
- particle gun at start of target
- fixed-target

amagnan:
Just one question: is it normal that we have no material Inconel on target front face ? They will start directly with Tungsten ??

amagnan (commented):
Everything looks ok to me! I had one worry about having a gap between target end and proximity shielding (about 12cm I think?) to keep the same z=0 as before, but, it is only 12cm so I guess it is OK ? Ideally we would not have that and have target vessel end touching the shielding right ?
Maybe one question then: if one decides to close that gap, and shift target by 12cm, z=0 would be moved but all the rest will stay the same in the geometry (just not in z values, but we should not rely on any z values hardcoded anywhere, everything should be with respect to the front of the target.....)

olantwin:
1. You're right, I should also add the front of the vessel.
2. For now I didn't want to move anything else, as I did not have time to check whether the positioning had changed in the integration drawings. Could someone please have a look and prepare a follow-up PR for the shielding and positioning relative to it?

olantwin:
Added the upstream closure of the vessel (commit f9015e051), extracted from the same CATIA model:

- **beam window**: dished membrane closing the bore behind the front flange nose; simplified to a flat disc (r = 141 mm) preserving the 8 mm of steel on the beam axis (z −33.5…−25.5 mm)
- **nose ring** r 141–195 mm (z −35.8…−23.8); the flange proper now starts at the cover-plate step (z = −23.8 mm)
- **cover plate** `ST2514472`: 600 × 650 × 20 mm with a stepped central bore (r 196/226 mm), asymmetric about the beam axis (+250 mm above, −400 mm below); modelled as two rings inside the He container plus a rectangular composite outside it
- He container extended upstream to z = −37.8 mm

The drawings in the PR description have been updated to show the closure (the z–y view shows the asymmetric plate). Verified: overlap check clean at 0.1 µm, material probes across window/nose/rings/plate all correct, and the legacy-design geometry dump is still identical to `main`.

olantwin:
@coderabbitai rate limit

olantwin:
@coderabbitai review

olantwin:
@THanae

PR #1355, opened 2026-07-28
Adding a pvc vessel to contain helium

rethanSHiP:
## Checklist

- Modified the veto/veto.cxx file to include per block 1 TropezoidHollow(), 1 Tropezoid() made out of 1 mm thick pvc. 

## Summary by CodeRabbit

* **New Features**
  * Added helium-balloon assemblies for both detector blocks.
  * Added configurable helium-balloon thickness and material settings.
  * Added PVC shielding with block-specific side walls and front/rear lids.
  * Added support for selecting the 2026 BDF target design and comparing it with the legacy target.
  * Added hidden decay-medium volumes within the detector assembly.

* **Bug Fixes**
  * Updated fiducial-volume and wall-distance checks to recognize revised decay-medium geometry names.

* **Documentation**
  * Updated the changelog with target-design and configuration details.

PR #1376, opened 2026-08-17
Nest decay vessel instead of fragile dimension matching

olantwin:
Reviewing @rethanSHiP's PR #1355, I noticed that we had some extremely fragile placement logic, which is also not ideal for navigation. This PR now addresses this by nesting, which makes overlaps impossible and hides complexity from Geant4, the event display etc.

Commits 1 and 2 refactor the placement, commit 3 only renames some of the touched variables. I would recommend reviewing each commit separately.

Positioning has been checked with the geo-information CI job and by comparing the node of random points before and after. No geometry changes were found apart from points which were on the boundary within numerical uncertainty.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Updated veto geometry with a continuous decay-medium volume, nested helium liner, and inner veto-wall components.
  - Preserved existing geometry dimensions, materials, rib layout, and scintillator placement.

- **Bug Fixes**
  - Improved fiducial-volume detection to recognize all decay-medium volumes, including consolidated geometry.

- **Documentation**
  - Added an unreleased changelog entry describing the updated decay-medium structure.

### DIFF
--- a/veto/veto.h
+++ b/veto/veto.h
@@ -159,12 +159,10 @@ class veto : public SHiP::Detector<vetoPoint> {
                                  Int_t color, TGeoMedium* material,
                                  Bool_t sens);
 
-  /** Adds a custom block of OuterWall+InnerWall+LiSc+Support ribs for a given
+  /** Adds a custom block of OuterWall+LiSc+Support ribs for a given
    * distance along z. Ensures consistency in implementation throughout the z.
    */
-  void AddBlock(TGeoVolumeAssembly* tInnerWall,
-                TGeoVolumeAssembly* tdecay_medium,
-                TGeoVolumeAssembly* tOuterWall, TGeoVolumeAssembly* tLongitRib,
+  void AddBlock(TGeoVolumeAssembly* tOuterWall, TGeoVolumeAssembly* tLongitRib,
                 TGeoVolumeAssembly* tVerticalRib, TGeoVolumeAssembly* ttLiSc,
                 int blockNr, int nx, int ny, double z1, double z2,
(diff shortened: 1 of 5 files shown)


olantwin:
NB: This should also improve performance by a decent amount for simulations with activity in the vessel.

olantwin:
@jasminwss, may I ask you for a review on this?

veto/veto.h
```diff
@@ -143,9 +143,11 @@ class veto : public SHiP::Detector<vetoPoint> {
    * dimensions of wX_start * wY_start and end cross-section dimensions of
    * wX_end *wY_end
    */
-  TGeoVolume* GeoTrapezoid(const TString& xname, Double_t wz, Double_t wX_start,
-                           Double_t wX_end, Double_t wY_start, Double_t wY_end,
-                           Int_t color, TGeoMedium* material, Bool_t sens);
+  TGeoVolume* make_geo_trapezoid(const TString& xname, Double_t wz,
```
jasminwss, line 146: maybe we could add the same parameter names as in veto.cxx (not Double_t wz but Double_t z_thick, …)

olantwin: I agree that the naming needs to be improved, but this PR keeps the old names by design to keep the lines changed to a minimum.

PRs for updating the names would be very welcome!

jasminwss (commented):
I would also change the parameter names in veto.h in make_geo_trapezoid so that they are equal to the ones used in veto.cxx. otherwise no comments

jasminwss (approved):
fine by me

olantwin:
Thanks for the review!

PR #1389, opened 2026-08-28
Fix splitting clone loss due to low momentum cut

olantwin:

Some more details at the computing meeting.

Events can still get very large. The per-step clone buffer is now capped (default 25000 clones, tunable via `--max-split-buffer`); once the cap is hit, further per-step splitting is skipped for that track, which reduces the statistical boost but conserves weight.

Another issue at high momenta remains, to be fixed in a separate PR.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable intermediate kaon/pion splitting before forced decays.
  * Added a buffered split-clone limit, defaulting to 25,000 records.
  * Added validation to ensure the buffer limit accommodates configured split counts.

* **Bug Fixes**
  * Prevented clone loss when tracks stop before taking a step or are skipped.
  * Improved repeated splitting and event-boundary buffer handling.
  * Added warnings for clones remaining buffered at event completion.
  * Reserved capacity for endpoint clones when enforcing buffer limits.

* **Documentation**
  * Updated changelog entries and command-line guidance for splitting controls.


matclim:
Just for my understanding, why does the buffer exist at all? `PostTrack()` already pushes continuation tracks directly via `stack->PushTrack()` in the `!isNaturalDecay && !isParticleDestroyed` branch, and that path apparently works. If decay clones could be pushed the same way, the buffer and this entire class of bug disappear.

macro/run_fixedTarget.py
```diff
@@ -90,7 +90,13 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
...
 )
 ap.add_argument(
-    "--multiple-kpi-splits", action="store_true", help="split kaons and pions multiple times along the track path"
+    "--multiple-kpi-splits",
+    action="store_true",
+    help="split kaons and pions multiple times along the track path. "
+    "WARNING: computationally unbounded — every step spawns N clones and clone "
+    "daughters are split recursively, which exhausts memory within a few "
+    "events now that buffered clones are reliably tracked; needs a cascade "
+    "guard before production use",
```
THanae: I would suggest that we investigate that before merging the pull request?

olantwin:
> Just for my understanding, why does the buffer exist at all? `PostTrack()` already pushes continuation tracks directly via `stack->PushTrack()` in the `!isNaturalDecay && !isParticleDestroyed` branch, and that path apparently works. If decay clones could be pushed the same way, the buffer and this entire class of bug disappear.

Researching this a bit, it seems like this can only be done during a step, not after it is complete (PostTrack), as any stack pushes would be discarded *before* the next PreTrack.
Even for the step clones which are created during a step, they'd only get popped during the next step, so we'd need a fallback for the last step.
While I agree that the buffer seems inelegant, I don't see a better way.

matclim:
Would it be possible to create the clones directly during the step rather than buffering them?

As I read it, both paths currently `buffer: ProcessHits` does `fSecondaryBuffer.push_back(clone)` for the per-step clones, and `PostTrack` does the same for the decay endpoint clones. The only `PushTrack` for clones is the flush in `PreTrack`.

For the per-step clones, could the push_back be replaced with an immediate `stack->PushTrack(1, trueParentId, track_pid, mom.Px(), ..., kPNoProcess, ntr, cloneWeight, 999)` plus `fCloneTracks.insert(ntr)`? They're created mid-step, so the popper should pick them up without any deferral.

For the endpoint clones, the same `kPDecay` detection `PostTrack `does is available one callback earlier, on the step where the decay happens. If they were pushed from there too, `fSecondaryBuffer` would have no writers left and the flush in `PreTrack`, the `BeginEvent` warning and the member itself could all go.

The reason I think this works: `TG4SteppingAction::UserSteppingAction` resurrects a dying track for one extra popper step when the VMC stack has pending pushes, and it does that check after calling the user stepping function — so a push from `ProcessHits` is caught even on the parent's final step. The geant4_vmc history entry for that change says it was added exactly because user tracks generated at the last step of the parent weren't being popped. `PostTrack` runs after that check, which would make it the actual root cause here, with the energy cut just being the most common way of exposing it.

The two caveats I can see are:

`ProcessHits` only fires in sensitive volumes while `PostTrack` fires wherever a track dies, and `ConstructGeometry` only registers `/target_vacuum_box_1` when `!fSplitOnce.` So this would need the registration extended to split-once mode, at the cost of a stepping callback in the target region where there currently is none. Worth checking whether that's affordable maybe?
The endpoint-clone code would need to live outside the `!fSplitOnce` guard in `ProcessHits`, since `PostTrack` currently produces those in both modes.

olantwin:
> Would it be possible to create the clones directly during the step rather than buffering them?
> 
> As I read it, both paths currently `buffer: ProcessHits` does `fSecondaryBuffer.push_back(clone)` for the per-step clones, and `PostTrack` does the same for the decay endpoint clones. The only `PushTrack` for clones is the flush in `PreTrack`.
> 
> For the per-step clones, could the push_back be replaced with an immediate `stack->PushTrack(1, trueParentId, track_pid, mom.Px(), ..., kPNoProcess, ntr, cloneWeight, 999)` plus `fCloneTracks.insert(ntr)`? They're created mid-step, so the popper should pick them up without any deferral.
> 
> For the endpoint clones, the same `kPDecay` detection `PostTrack `does is available one callback earlier, on the step where the decay happens. If they were pushed from there too, `fSecondaryBuffer` would have no writers left and the flush in `PreTrack`, the `BeginEvent` warning and the member itself could all go.
> 
> The reason I think this works: `TG4SteppingAction::UserSteppingAction` resurrects a dying track for one extra popper step when the VMC stack has pending pushes, and it does that check after calling the user stepping function — so a push from `ProcessHits` is caught even on the parent's final step. The geant4_vmc history entry for that change says it was added exactly because user tracks generated at the last step of the parent weren't being popped. `PostTrack` runs after that check, which would make it the actual root cause here, with the energy cut just being the most common way of exposing it.
> 
> The two caveats I can see are:
> 
> `ProcessHits` only fires in sensitive volumes while `PostTrack` fires wherever a track dies, and `ConstructGeometry` only registers `/target_vacuum_box_1` when `!fSplitOnce.` So this would need the registration extended to split-once mode, at the cost of a stepping callback in the target region where there currently is none. Worth checking whether that's affordable maybe? The endpoint-clone code would need to live outside the `!fSplitOnce` guard in `ProcessHits`, since `PostTrack` currently produces those in both modes.

From my reading of the code I don't think this is possible, but I'd happily be mistaken. Could you maybe flesh this out as a PR so that we can test this approach?

olantwin:
NB @THanae: This now includes a cap of 10000 on the stack. We should test whether this is ever reached and if not, what margin we have..

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -40,6 +41,13 @@ constexpr Double_t cm = 1;         // cm
 constexpr Double_t m = 100 * cm;   //  m
 constexpr Double_t mm = 0.1 * cm;  //  mm
 
+// Safety valve for the intermediate (per-step) splitting: the clone buffer
+// grows by fIntermediateNsplits on every qualifying step and is only drained in
+// PreTrack(), so a pathological split count could make it grow without bound
+// within a single track. ~10k TrackBuffer records is about 1.2 MB, far above
+// what any sane configuration reaches.
+constexpr std::size_t kMaxSecondaryBuffer = 10000;
```
THanae: Do you think it might be useful to have this as an (expert-only) argument, for cases where more splitting (and accordingly more memory) might be required? It might not be a super common requirement so unclear to me if we want that.

olantwin: Yes, we should probably expose that as an argument, with appropriate warnings.

olantwin: Done.

muonShieldOptimization/exitHadronAbsorber.h
```diff
@@ -44,6 +45,10 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
   // void FinishEvent();
 
   void SetNSplits(Int_t n) { fNsplits = n; }
+  void SetIntermediateNSplits(Int_t n) { fIntermediateNsplits = n; }
+  void SetMaxSplitBuffer(Int_t n) {
+    fMaxSplitBuffer = static_cast<std::size_t>(n);
```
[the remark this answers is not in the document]
olantwin: This should now be addressed by requiring a positive argument on the command-line.

olantwin:
@coderabbitai full review

olantwin:
@matclim @THanae I'd like to merge this PR and #1389 before the CM, ideally today so that there's time to run new simulations and tests on the output. I'd appreciate it, if you could review.

THanae:
> @matclim @THanae I'd like to merge this PR and #1389 before the CM, ideally today so that there's time to run new simulations and tests on the output. I'd appreciate it, if you could review.

I'm happy with this PR (and tests so far on thousands of longer subjobs show no issue). As discussed privately, I would also like a cap on the total stack size, but I can add this later in another PR.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ it in future.
 ### Fixed
 
 * `veto` now registers the configured `sensitiveMed` instead of a hardcoded medium name; previously any other value resolved to a null `TGeoMedium`
+* Kaon/pion splitting no longer silently loses the whole clone set when the first track after the decay falls below the energy cut (the buffered clones were discarded by the stack popper's reset before being tracked). This removed ~17% of muons from charged-kaon decay in flight (−2.8% of the total muon rate) in split productions; pions were unaffected because their soft decay muons imply equally soft, unobservable clones. The clone buffer is now flushed only into tracks that survive the cut, and a warning is emitted if an event ends with clones still buffered. To avoid memory issues, a specific number is set for the splits per GEANT4 step, and the per-step clone buffer is capped: once the cap is reached, further per-step splitting is skipped for that track, which reduces the statistical boost but conserves weight. The cap defaults to 25000 buffered clones, is configurable with `--max-split-buffer`, and is a hard bound: per-step splitting stops early enough to leave room for the clones added when the parent finally decays
```
THanae: Maybe just one comment, this feels quite wordy / lengthy. If there is an easy way for you to shorten it while keeping the relevant information that might be helpful for someone having a quick look. Otherwise no worries

macro/run_fixedTarget.py
```diff
@@ -87,10 +87,28 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
...
+    "in order to boost the number of muons stemming from forced decays. "
+    "EXPERTS ONLY: risk of very large memory usage if the number is set without proper testing",
+)
+ap.add_argument(
+    "--max-split-buffer",
+    type=int,
+    default=25000,
+    help="EXPERTS ONLY: maximum number of split clones buffered per track before further per-step splitting is "
+    "skipped. Memory safety valve for --multiple-kpi-splits; lowering it reduces the statistical boost but "
+    "conserves weight",
```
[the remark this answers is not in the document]
olantwin: This was intentionally changed. Did I miss updating this anywhere but in the PR message?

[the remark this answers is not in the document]
THanae: In the AI description at the top of this PR I think

CHANGELOG.md
olantwin: I'll fix this.

olantwin: Done

THanae (dismissed):

olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -238,6 +238,18 @@ void exitHadronAbsorber::Initialize() {
...
+    // the previous event, so its clones could not be handed to the stack
+    // popper and their weight is lost.
+    Double_t lostWeight = 0;
+    for (const auto& trk : fSecondaryBuffer) {
+      lostWeight += trk.weight;
+    }
+    LOG(warning) << "exitHadronAbsorber: discarding " << fSecondaryBuffer.size()
+                 << " buffered split clones (summed weight " << lostWeight
+                 << ") left over from the previous event";
+  }
```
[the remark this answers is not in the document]
olantwin: This is part of #1398

olantwin:
@matclim @THanae This should now ready for a final review. Question is whether we first want to merge #1403 in and clean up the commit history again.

THanae (dismissed):

matclim:
A full review from me is upcoming, will attempt to do it before the software meeting on Friday.

Excellent job both of you in any case!

olantwin:
Thanks @matclim

matclim:
The energy cut now should be correctly ahead of the flush if I'm reading this right, but is it the only path that kills a track before its first step? fSkipNeutrinos calls StopTrack() further down in PreTrack, after the buffer has already been pushed and cleared:


if ((fMom.E() - fMom.M()) < EMax) { StopTrack(); return; }   // buffer kept
if (!fSecondaryBuffer.empty()) { ...PushTrack...; clear(); } // flushed here
...
if (fSkipNeutrinos && (idabs == 12 || idabs == 14 || idabs == 16)) StopTrack();

So if the first track after a splitting decay is an above-threshold neutrino and `--skip-neutrinos` is active, the kaon/pion clones get pushed into a carrier that never steps, the stack popper never runs, and the clone set is dropped at the next track's Reset(), the same muon loss fixed here, via a different route. Since K -> mu nu puts a neutrino in the stack right beside the muon, that carrier isn't an unlikely draw.

It also evades the new warning: the buffer was cleared in step 3, so BeginEvent (or FinishEvent) sees nothing pending and the loss stays silent.

Am I getting this right?

olantwin:
I fear you might be right, investigating...

olantwin:
@matclim Should be addressed now, checking for an alive track as well, just to be on the safe side.

matclim:
I'll have a look a bit later, thanks for the quick reaction!

PR #1281, opened 2026-06-24
Updated Muon DIS simulation chain

amagnan:
## Checklist

- [x] Changelog updated (if applicable)
- [ ] Commit message follows [conventional commits](https://www.conventionalcommits.org/)
- [ ] CI checks pass

Additional code added to muonDIS folder, including the top macro to execute to make rootfile with all DIS information: prepareEvents.py. TO DO: The output tree will then have to be propagated again through a DISGenerator to remake independent events and process daughters trhough Geant4 again. 

## Summary by CodeRabbit

* **New Features**
  * Added a Muon DIS simulation workflow that derives material-dependent paths/vertices and generates DIS final states, writing dedicated ROOT outputs with full track/hit and weight information.
  * Introduced new Muon DIS processing, geometry-path, and event-branching components, and updated the generator to read DIS events from input ROOT files.
* **Documentation**
  * Expanded build-from-source and environment setup guidance.
  * Added a dedicated Muon DIS README and updated usage instructions.
* **Bug Fixes**
  * Improved robustness in muon/track ancestry handling to avoid non-terminating chains.
  * Enhanced particle-gun theta handling with validation and safe defaults.
* **Build / Packaging**
  * Added and integrated the Muon DIS build targets/libraries.
* **Removed**
  * Removed legacy Muon DIS helper scripts and generators in favor of the new workflow.

PR #1396, opened 2026-09-02
feat: refactor muonDIS production with c++ classes and per material type

amagnan:
## Checklist

- [ x] Changelog updated (if applicable)
- [ x] Commit message follows [conventional commits](https://www.conventionalcommits.org/)

Clean-up of PR #1281 
New classes are all in newMuonDIS folder, including:
-  the python script to generate the DIS events - prepareEvents.py 
- and the FairGenerator to process them again in FairShip - NewDISGenerator.h/.cxx 
- also with the updated run_simScript_newDIS.py.

Still TODO:
- add B field for the building of the muon path, and selecting a DIS vertex along that path.
- add back the initial muon hits in UBT/SBT/SST that happened before the DIS vertex, in the run_simScript.
- add a filter post-prepareEvents.py with configurable parameters, to avoid processing all events in Geant4, the NewDISGenerator step is way too time consuming at the moment.
- optimise further the NewDISGenerator step: e.g. skip propagating particles in the cavern at this stage, except muons for potential bouncing.
- add a validation script to make validation plots.

## Summary by CodeRabbit

* **New Features**
  * Added muon deep-inelastic scattering simulation across detector materials.
  * Added muon path tracking through detector geometry, including material and volume classification.
  * Added support for generating and reading DIS events in SHiP simulations.
  * Added command-line tools for preparing DIS events and running simulations.
* **Documentation**
  * Added setup, configuration, input/output, and usage guidance.
* **Build**
  * Integrated muon DIS functionality into the project build.
* **Bug Fixes**
  * Improved handling of missing data and invalid geometry during processing.


newMuonDIS/DISparticle.h
```diff
@@ -0,0 +1,34 @@
+#ifndef SHIPMuDIS_DISparticle_H_
+#define SHIPMuDIS_DISparticle_H_
+
+#include "Rtypes.h"   // for Double_t, Int_t, Double32_t, etc
+#include "TObject.h"  // for TObject
```
olantwin: No need for these, there is no reason for this to inherit from TObject. The dictionary is only needed for the std::vector DISparticle.

Given recent experiences in data-model, I'd recommend versioning the class in the LinkDef.

newMuonDIS/MuGeoProcessor.h
```diff
@@ -0,0 +1,93 @@
...
+#include "MuonPath.h"
+#include "TGeoManager.h"
+#include "TGeoMaterial.h"
+#include "TGeoNode.h"
+#include "TGeoShape.h"
+#include "TGeoVolume.h"
+#include "TMath.h"
+#include "TVector3.h"
+
+using namespace ShipMuDIS;
```
olantwin: Either use `using with specific symbols, or work within the namespace, to which this class should probably also belong.

newMuonDIS/MuDISDefs.cxx
```diff
@@ -0,0 +1,2 @@
+#include "MuDISDefs.h"
```
olantwin, line 1: Why is this stub needed?

newMuonDIS/NewMuDISGenerator.h
```diff
@@ -0,0 +1,53 @@
...
+#include "Generator.h"
+#include "TF1.h"
+#include "TROOT.h"
+#include "TChain.h"
+#include "TVector3.h"
+#include "vector"
+
+#include "MuDISDefs.h"
+
+using namespace ShipMuDIS;
```
olantwin: See above. By the way, SHiP::muon_dis or something like that would be better.

newMuonDIS/NewMuDISGenerator.h
```diff
@@ -0,0 +1,53 @@
...
+  Bool_t ReadEvent(FairPrimaryGenerator*) override;
+  Bool_t Init(const char*, int) override;
+  Bool_t Init(const char*) override;
+  Bool_t Init(const std::vector<std::string>&, int) override;
+  Bool_t Init(const std::vector<std::string>&) override;
+  Int_t GetNevents();
+  void SetNevents();
+
+ protected:
+  FairLogger* fLogger;  //!   don't make it persistent, magic ROOT command
```
olantwin: This class won't be persisted anyway, so no need for ROOT I/O comments

newMuonDIS/prepareEvents.py
```diff
@@ -0,0 +1,113 @@
...
+## Load library
+#cppyy.load_library("/cvmfs/ship.cern.ch/26.05/sw/slc9_x86-64/pythia6/latest/lib/libPythia6.so")
+## Include header
+#cppyy.include("/cvmfs/ship.cern.ch/26.05/sw/slc9_x86-64/ROOTEGPythia6/latest/include/TPythia6.h")
+## Create object
+#myPythia = cppyy.gbl.TPythia6()
+#print("Pythia6 created")
+"""
+End of pythia block to avoid aliBuild error
+"""
```
olantwin, line 18: Better to do this programatically, e.g. trying to load the library.

newMuonDIS/prepareEvents.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
+"""Script to collect muons hitting either the UBT extended plane, or the SBT, including soft interaction products, and DIS events, to a ROOT file."""
+
+"""Created May 2026 A.-M. Magnan"""
+
+"""
```
olantwin, line 4: SPDX header missing. Module level comment can stay, author/date notice isn't needed, it's in the git history.

newMuonDIS/run_simScript_newDIS.py
olantwin, line 1: Definitely no. run_simScript.py should be updated, not a copy made.

olantwin (changes requested):
Thanks for the cleanup!

By no means a full review, just a first few comments on things that stood out on a first skim.

General comments:
- I'd prefer we replace the old muon DIS completely.
- Since we're introducing new classes, functionality etc., we should take care to give things sensible names
- The python script should go into macro
- At least the generator can go into shipgen
- The coderabbit comments and lint issues generally look all worth fixing.

olantwin:
For the TODO items, do you want to address them later as separate PRs or as part of this PR?

newMuonDIS/MuDISProcessor.cxx
```diff
@@ -0,0 +1,490 @@
...
+  fPythia->Pylist(1);
+
+  double lastxs = 0;
+  for (int ia(0); ia < fnDIS; ++ia) {
+    LOG(debug) << " ---- Processing DIS event " << ia;
+    //@FIXME AMM - half-way through, we change to neutron target with 50-50 :
+    //---> update to real material ??
+    if (ia == static_cast<int>(fnDIS / 2)) {
+      fPythia->Initialize("FIXT", tType.c_str(), "n0", P);  // target = "n0"
+      isProton = false;
```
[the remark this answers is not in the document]
amagnan: At the moment we do not see any difference between p and n, so this is kept like this for now and to be improved once we actually have different events for p and n...

newMuonDIS/MuDISDefs.cxx
amagnan: So the include is put within the library... At least I encountered the issue and could not fix it otherwise....

amagnan:
Definitely later PR for the TODOs. Also, I separated so it was clear what needs reviewing (including the new MuonDISGenerator). Otherwise the previous PR was exactly replacing the older code, but, it became a mess each time you want to remerge with master given files are deleted / added. So, if that's OK, I'd like to have this validated separately, then, I can do another PR to delete old / replace by new very easily. Including the duplicate of the run_simScript which was just there temporarily for review and for completeness. 
I will implement the review by coderabbit, most indeed sounds sensible.

newMuonDIS/MuDISDefs.cxx
olantwin: You can use this approach:
https://github.com/ShipSoft/FairShip/blob/853cf4a887d9afe52269737fc6e4333704d46229/shipdata/CMakeLists.txt#L9

olantwin:
> Definitely later PR for the TODOs. 

Ok!

> Also, I separated so it was clear what needs reviewing (including the new MuonDISGenerator). Otherwise the previous PR was exactly replacing the older code, but, it became a mess each time you want to remerge

The problem were the merges, not replacing the files As a rule of thumb, in ShipSoft repos, never merge, only rebase. While there were some minor merge conflicts, those can be resolved fairly easily.

> with master given files are deleted / added. So, if that's OK, I'd like to have this validated separately, then, I can do another PR to delete old / replace by new very easily.

Ok, we can review things separately, but then they should be moved to sensible directories.

> Including the duplicate of the run_simScript which was just there temporarily for review and for completeness.

For run_simScript.py, I'll need the diff to review.

> I will implement the review by coderabbit, most indeed sounds sensible.

Thanks! And please run `pixi run lint`

olantwin:
> 🤖 [Coding task started](https://app.coderabbit.ai/code/tasks/112f9d5b-7074-4595-b84e-9ecd7164cd4d?source=coding_agent_github_pr_comment) for 10 unresolved review comments.

@amagnan Did you trigger this?

amagnan:
I did: I thought I would try and then review the commit ?? Should I not ? But not sure how it works....

olantwin:
I've never tried, but it seems to not have worked anyway (I think an issue with the environment?). I was just confused.

Copying the review feedback and using claude code, opencode or codex locally worked quite well in my experience.

PR #1398, opened 2026-09-04
Fix residual kaon/pion splitting clone loss

olantwin:

Stacked on #1389 — please review/merge that one first; the base of this PR is `fix-splitting-clone-loss`.

#1389 removes the dominant clone-loss channel, but instrumented runs afterwards still showed whole clone sets going missing for roughly 6 % of splitting decays. Two independent channels are responsible, and this PR closes both.

### 1. The energy cut was applied to the clones themselves

A clone is the parent kaon or pion re-injected at its decay point so the decay can be re-sampled. Its own `PreTrack` applied the `EMax` kinetic-energy cut a second time, although the parent had already passed that cut at the start of its own track. A parent that dropped below the cut before decaying therefore had every clone stopped before `ForceDecayTime(0)` could fire — while an unsplit run keeps such a decay and applies the cut to the decay muon instead, which for K → μν can be much harder than the parent.

Instrumented runs reproduce this deterministically: of 40 splitting decays in a 10k-event run, the 3 with a sub-threshold parent lost every clone and all 37 others were fine.

### 2. The clone buffer could be stranded until the end of the event

Clones are buffered at the decay and handed to the VMC stack in a later track's `PreTrack`, because `TG4StackPopper` only drains the stack during a step. If every track after the decay was stopped before its first step — by the energy cut, or by `SkipNeutrinos` — the buffer survived to the end of the event and was dropped in `ShipStack::Reset`.

A decay always puts its daughters on the stack, so a track always follows it. The first such track is now designated as the carrier and exempted from both stops so that it does step. This costs at most one sub-threshold track transported per splitting decay and cannot change the output: `ProcessHits` only scores crossings above the same cut, and no secondary of a sub-threshold track exceeds it. The designation is shared between instances because `run_fixedTarget` can add up to three sensitive planes and modules are polled in `AddModule` order, so a sibling plane would otherwise stop the carrier before the splitting instance sees it.

A third hypothesis — that a flushed buffer can go unpopped when the carrier takes a zero-length step — was checked and **refuted**: of 14 zero-length π⁰/η/Σ⁰ carriers in the instrumented logs, 12 delivered all 100 clones, and the 2 that did not are exactly the sub-threshold-parent events above. `TG4SteppingAction`'s exclusive-step rescue covers that case, so no retry machinery is needed.

### 3. The loss is now visible in production logs

The per-event warning could only see the buffer left by the *previous* event, so a clone set stranded in the last event of a run went unreported, and nothing summarised the total. `FinishRun` now inspects the buffer once more and reports how many times the splitting fired, how many clones it produced, and how many were never tracked. A production which reports a non-zero loss undercounts the corresponding decay channel.

## Verification

Runs use `run_fixedTarget` with `-e 5`, `--AddHadronAbsorberOnly`, and a scaled K± lifetime to turn decay-in-flight from a rare process into a common one (kaon gun for the loss counting, Pythia for the weighted comparison).

**Loss regression**

| | splitting decays | clone sets lost |
|---|---|---|
| before | ~36 600 | **544** |
| after | 36 622 | **0** |

The "before" figure counts only the stranded-buffer channel, since the sub-threshold-clone channel is silent without the fix, so it is a lower bound.

**Closure at the scoring plane**, split100 vs unsplit (18k vs 90k events): all muons **1.0096 ± 0.0247**, muons from charged-kaon decay **1.0132 ± 0.0248**, with the momentum spectrum agreeing bin by bin up to 160 GeV/c.

Two things worth flagging honestly:

- This configuration cannot resolve the *rate* impact of the fix (a paired comparison of the split runs before/after gives −6.3 % ± 3.5 %, i.e. 1.8σ). The per-event weighted sum is dominated by rare weight-≈1 decays, so the ×100 boost buys tail coverage rather than rate precision. The case for the fix rests on the direct clone-set accounting; the closure test's value here is confirming that the new carrier exemption introduces no bias.
- With `--multiple-kpi-splits` and the buffer cap from #1389, a 300-event run performs 297 137 splits producing 623 674 clones and loses none, so the mode is exercised too — but no physics closure is claimed for it.

Each commit builds on its own. `ci-fixed-target` is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved kaon and pion splitting so buffered clone tracks are reliably processed after splitting.
  * Prevented clone and carrier tracks from being incorrectly stopped by energy or neutrino-selection cuts.
  * Added safeguards for buffered clones remaining at event or run completion.

* **Configuration**
  * Increased the maximum supported split-buffer capacity and documented its hard limit.

* **Monitoring**
  * Added reporting for split decays, buffered clones, lost clone events, untracked clone tracks, and untracked clone weight.

### DIFF
--- a/muonShieldOptimization/exitHadronAbsorber.h
+++ b/muonShieldOptimization/exitHadronAbsorber.h
@@ -72,6 +72,10 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
   // Ask for the next track to carry the clones just added to the buffer.
   static void RequestCloneCarrier() { fgCarrierTrackID = kCarrierRequested; }
 
+  // Drop the clones still buffered at the end of an event or run, keeping
+  // track of how much weight was never simulated.
+  void DiscardBufferedClones();
+
   Int_t fUniqueID = 0;
   Bool_t fOnlyMuons;         //! flag if only muons should be stored
   Bool_t fSkipNeutrinos;     //! flag if neutrinos should be ignored
@@ -108,6 +112,12 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
   std::set<Int_t> fCloneTracks;
   std::set<Int_t> fContinuationTracks;
   std::set<Int_t> fDecayedParentIDs;
+
+  Int_t fSplitDecays = 0;          //! decays replaced by clones
+  Int_t fClonesBuffered = 0;       //! clones created for those decays
+  Int_t fLostBufferEvents = 0;     //! events ending with unflushed clones
+  Int_t fLostCloneTracks = 0;      //! clones which were never tracked
+  Double_t fLostCloneWeight = 0.;  //! summed weight of those clones
 
   TFile* fout;               //!
   TClonesArray* fElectrons;  //!
--- a/muonShieldOptimization/exitHadronAbsorber.h
+++ b/muonShieldOptimization/exitHadronAbsorber.h
@@ -59,6 +59,19 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
   inline void SetUseCaveCoordinates() { fUseCaveCoordinates = kTRUE; }
 
  private:
+  // Hand the buffered clones of a splitting decay to the stack popper. They
+  // can only be popped during a step, so the first track after the decay is
+  // designated as their carrier and must not be stopped before it steps.
+  // The designation is shared by all instances: run_fixedTarget can add
+  // several sensitive planes and whichever is polled first would otherwise
+  // stop the carrier before the splitting instance ever sees it.
+  static constexpr Int_t kNoCarrier = -1;
+  static constexpr Int_t kCarrierRequested = -2;
+  static Int_t fgCarrierTrackID;  //!
+
+  // Ask for the next track to carry the clones just added to the buffer.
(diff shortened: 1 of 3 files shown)


olantwin:
@coderabbitai review

olantwin:
@coderabbitai review

muonShieldOptimization/exitHadronAbsorber.h
```diff
@@ -59,6 +59,23 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
...
 
  private:
+  // Hand the buffered clones of a splitting decay to the stack popper. They
+  // can only be popped during a step, so the first track after the decay is
+  // designated as their carrier and must not be stopped before it steps.
+  // The designation is shared by all instances: run_fixedTarget can add
+  // several sensitive planes and whichever is polled first would otherwise
+  // stop the carrier before the splitting instance ever sees it.
+  static constexpr Int_t kNoCarrier = -1;
+  static constexpr Int_t kCarrierRequested = -2;
```
THanae, line 69: I am not sure I understand the `-2`. Is it just because no "real" TrackID will have that number?

olantwin: Yes, and because we need a different value from -1.

olantwin:
@coderabbitai review

PR #1403, opened 2026-09-08
feat(splitting): Add a cap on the size of the stack

THanae:
Added a cap on the stack size to avoid memory issues for large number of intermediate splits

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a `--max-event-size` command-line option, defaulting to 5,000,000.
  - Event processing now enforces a hard limit on event size during kaon and pion splitting.
  - The configured limit must be positive and at least as large as the split count.
  - Clear warnings indicate when event-size or buffering limits prevent further splitting.

- **Documentation**
  - Documented the new command-line option in the unreleased changelog.

### DIFF
--- a/macro/run_fixedTarget.py
+++ b/macro/run_fixedTarget.py
@@ -106,6 +106,15 @@ ap.add_argument(
     "conserves weight",
 )
 ap.add_argument(
+    "--max-event-size",
+    type=int,
+    default=5_000_000,
+    help="EXPERTS ONLY: maximum size of the event (accounting for an extra safety factor  before further"
+         " per-step splitting is skipped. Memory safety valve for --multiple-kpi-splits;"
+         "lowering it reduces the statistical boost but conserves weight",
+)
+
+ap.add_argument(
     "--multiple-kpi-splits",
     action="store_true",
     help="split kaons and pions multiple times along the track path",
@@ -247,11 +256,19 @@ if args.multiple_kpi_splits and (args.kaon_pion_splits <= 0 or args.intermediate
     ap.error("--multiple-kpi-splits requires --kaon-pion-splits > 0 and --intermediate-kaon-pion-splits > 1")
 if args.max_split_buffer < 1:
     ap.error("--max-split-buffer must be >= 1")
+if args.max_event_size < 1:
+    ap.error("--max-event-size must be >= 1")
+
 if args.max_split_buffer < args.kaon_pion_splits:
     ap.error(
         "--max-split-buffer must be >= --kaon-pion-splits: the clones buffered when the parent "
         "decays have to fit under the cap"
     )
+if args.max_event_size < args.kaon_pion_splits:
+    ap.error(
+        "--max-event-size must be >= --kaon-pion-splits: the event size has to fit under the cap"
+    )
+
 
 
 if args.G4only:
@@ -418,6 +435,7 @@ if args.AddMuonShield or args.AddHadronAbsorberOnly:
 sensPlaneHA = ROOT.exitHadronAbsorber()
 sensPlaneHA.SetNSplits(args.kaon_pion_splits)
 sensPlaneHA.SetMaxSplitBuffer(args.max_split_buffer)
+sensPlaneHA.SetMaxEventSize(args.max_event_size)
 if args.multiple_kpi_splits:
     sensPlaneHA.SetSplitMultipleTimes()
     sensPlaneHA.SetIntermediateNSplits(args.intermediate_kaon_pion_splits)
--- a/muonShieldOptimization/exitHadronAbsorber.cxx
+++ b/muonShieldOptimization/exitHadronAbsorber.cxx
@@ -79,6 +79,16 @@ void exitHadronAbsorber::SetMaxSplitBuffer(Int_t n) {
   fMaxSplitBuffer = static_cast<std::size_t>(n);
 }
 
+void exitHadronAbsorber::SetMaxEventSize(Int_t n) {
+  if (n < 1) {
+    LOG(error) << "exitHadronAbsorber: max event size must be >= 1, ignoring "
+               << n;
+    return;
+  }
+  fMaxEventSize = static_cast<std::size_t>(n);
+}
+
+
 Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
   /** This method is called from the MC stepping */
   TString volName = gMC->CurrentVolName();
@@ -162,10 +172,18 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
           // when the parent finally decays, so the buffer never exceeds
           // fMaxSplitBuffer. All terms are on the left to keep the unsigned
           // arithmetic from wrapping.
-          if (fSecondaryBuffer.size() +
-                  static_cast<std::size_t>(fIntermediateNsplits) +
-                  static_cast<std::size_t>(fNsplits) >
-              fMaxSplitBuffer) {
+          std::size_t secondaryBufferSize = fSecondaryBuffer.size();
+          std::size_t event_size = gMC->GetStack()->GetNtrack();
+
+          const std::size_t pending = secondaryBufferSize +
+                             static_cast<std::size_t>(fIntermediateNsplits) +
+                             static_cast<std::size_t>(fNsplits);
+          const bool bufferFull = pending > fMaxSplitBuffer;
+
+          // multiplying pending tracks by factor to account for showers
+          const int32_t shower_safety_factor = 500;
+          const bool eventFull  = event_size + pending * shower_safety_factor > fMaxEventSize;
+          if (bufferFull || eventFull) {
             // Skip the split for this step instead of truncating the buffer.
             // fCurrentSurvivalFactor is the weight ledger: leaving it untouched
             // means the weight we did not split off is still carried by the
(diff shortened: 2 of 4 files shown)


olantwin:
@THanae This will need rebasing, dropping all the conflicting commits, as I rewrite the base branch.

THanae:
> @THanae This will need rebasing, dropping all the conflicting commits, as I rewrite the base branch.

All done :)

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -162,10 +172,12 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
...
           // arithmetic from wrapping.
-          if (fSecondaryBuffer.size() +
-                  static_cast<std::size_t>(fIntermediateNsplits) +
+          std::size_t secondaryBufferSize = fSecondaryBuffer.size();
+          std::size_t stack_size = gMC->GetStack()->GetNtrack();
+          if ((secondaryBufferSize + static_cast<std::size_t>(fIntermediateNsplits) +
                   static_cast<std::size_t>(fNsplits) >
-              fMaxSplitBuffer) {
+              fMaxSplitBuffer) || (stack_size + secondaryBufferSize *
+              static_cast<std::size_t>(fIntermediateNsplits) > fMaxStackSize)) {
```
olantwin: This is hard to follow and I think also wrong (intermediate splits contribute only secondaryBufferSize tracks to the stack).

Maybe split this into
```cpp
const std::size_t pending = secondaryBufferSize +
                             static_cast<std::size_t>(fIntermediateNsplits) +
                             static_cast<std::size_t>(fNsplits);
 const bool bufferFull = pending > fMaxSplitBuffer;
 const bool stackFull  = stackCount + pending > fMaxStackSize;
 if (bufferFull || stackFull) { ... }
```

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -175,11 +187,13 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
...
                               "reached "
-                           << fMaxSplitBuffer
+                           << secondaryBufferSize
+                           << " entries and stack size reached "
+                           << stack_size
                            << " entries; skipping further per-step splitting "
                               "for this track. Consider lowering "
                               "--intermediate-kaon-pion-splits or raising "
-                              "--max-split-buffer.";
+                              "--max-split-buffer or --max-stack-size.";
```
olantwin: Now that two different conditions can cause this warning, we should probably give a separate warning depending on what happened.

muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -162,10 +172,12 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
           // when the parent finally decays, so the buffer never exceeds
           // fMaxSplitBuffer. All terms are on the left to keep the unsigned
           // arithmetic from wrapping.
-          if (fSecondaryBuffer.size() +
-                  static_cast<std::size_t>(fIntermediateNsplits) +
+          std::size_t secondaryBufferSize = fSecondaryBuffer.size();
+          std::size_t stack_size = gMC->GetStack()->GetNtrack();
```
olantwin: This is not actually the stack size, but the total number of particles in the event. This is arguable an error in FairRoot's naming, which we copied.

This is the right thing to limit though, the stack itself is internal and just a pointer into the array of particles in the event. So this don't-call-it-stack is the main driver of memory (and means that draining the stack to Geant4 earlier won't solve memory issues unless we really split things into separate events (CC @matclim because that was his idea).

So the naming/documentation/comments everywhere are potentially confusing. Not sure what a better name is though.

macro/run_fixedTarget.py
```diff
@@ -105,6 +105,15 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
     "skipped. Memory safety valve for --multiple-kpi-splits; lowering it reduces the statistical boost but "
     "conserves weight",
 )
+ap.add_argument(
+    "--max-stack-size",
+    type=int,
+    default=5000000,
+    help="EXPERTS ONLY: maximum size of the stack (per event) before further per-step splitting is "
```
olantwin: This should be rephrased, maybe call it event size?

macro/run_fixedTarget.py
```diff
@@ -247,13 +256,15 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
...
 if args.max_split_buffer < 1:
     ap.error("--max-split-buffer must be >= 1")
-if args.max_split_buffer < args.kaon_pion_splits:
+if args.max_stack_size < 1:
+    ap.error("--max-stack-size must be >= 1")
+
+if args.max_split_buffer < args.kaon_pion_splits or args.max_stack_size < args.kaon_pion_splits:
     ap.error(
-        "--max-split-buffer must be >= --kaon-pion-splits: the clones buffered when the parent "
+        "--max-split-buffer and --max-stack-size must be >= --kaon-pion-splits: the clones buffered when the parent "
```
olantwin: The part after the colon is not correct if the max-stack-size is the issue.

macro/run_fixedTarget.py
```diff
@@ -247,13 +256,15 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
     ap.error("--multiple-kpi-splits requires --kaon-pion-splits > 0 and --intermediate-kaon-pion-splits > 1")
 if args.max_split_buffer < 1:
     ap.error("--max-split-buffer must be >= 1")
-if args.max_split_buffer < args.kaon_pion_splits:
+if args.max_stack_size < 1:
+    ap.error("--max-stack-size must be >= 1")
+
+if args.max_split_buffer < args.kaon_pion_splits or args.max_stack_size < args.kaon_pion_splits:
```
olantwin: Since the reasons and error messages for both are different, it might be worth checking these separately.

olantwin (changes requested):

muonShieldOptimization/exitHadronAbsorber.cxx
THanae: I agree that it should be separated, but the solution above has a tendency to lead to larger memory usage. I was using a multiplication because the increment in number of tracks is tends to be much larger depending on the number of intermediate splittings, meaning that a cut like the one suggested above would still lead to cases with 7M tracks before it gets stopped, as opposed to 5M or closer.

olantwin: By what mechanism? For memory, it shouldn't matter, unless I'm missing something.

THanae: I assume that the split particles can shower, etc such that even if an intermediate split is 100 (I am purposefully testing with large numbers to see what "breaks"), many more particles can come out of it.

THanae: But I agree with you, this multiplication approach is unclear

olantwin: In that case, we should probably document it as an empirical safety factor and give it a different name.

THanae: Agreed, doing that now (checking a few intermediate splitting numbers to make sure the safety factor makes sense) :)

THanae: I added a safety factor fo 500 to avoid memory jumps. On the tests I have run with the (default) intermediate splitting of 2, this does not seem to change things, but this helps stabilise the memory for higher intermediate splitting factors. Note that this means that the buffer size never reaches 25000 as a result. I think it would still be worth keeping it for cases where the `max-event-size` is changed manually

muonShieldOptimization/exitHadronAbsorber.cxx
THanae: I have now separated them in the code

muonShieldOptimization/exitHadronAbsorber.cxx
THanae: For now, I changed `stack_size` to `event_size`

macro/run_fixedTarget.py
THanae: Done!

macro/run_fixedTarget.py
THanae: You're right, I separated the warnings and made the second one clearer

macro/run_fixedTarget.py
THanae: I have now separated them

macro/run_fixedTarget.py
```diff
@@ -105,6 +105,15 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
     "skipped. Memory safety valve for --multiple-kpi-splits; lowering it reduces the statistical boost but "
     "conserves weight",
 )
+ap.add_argument(
+    "--max-event-size",
+    type=int,
+    default=5000000,
+    help="EXPERTS ONLY: maximum size of the event before further per-step splitting is "
+    "skipped. Memory safety valve for --multiple-kpi-splits; lowering it reduces the statistical boost but "
```
olantwin: Would be good to document what unit this is (i.e. tracks/particles), as this could be interpreted as a number of e.g. bytes as well.

macro/run_fixedTarget.py
```diff
@@ -247,11 +256,19 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
...
 if args.max_split_buffer < args.kaon_pion_splits:
     ap.error(
-        "--max-split-buffer must be >= --kaon-pion-splits: the clones buffered when the parent "
+        "--max-split-buffer and must be >= --kaon-pion-splits: the clones buffered when the parent "
         "decays have to fit under the cap"
     )
+if args.max_event_size < args.kaon_pion_splits:
+    ap.error(
+        "--max-event-size must be >= --kaon-pion-splits: the event size has to fit under the cap"
+    )
```
olantwin, line 270: Isn't it the other way around: The number of splits * safety factor have to fit into the event size?

macro/run_fixedTarget.py
```diff
@@ -247,11 +256,19 @@ def get_work_dir(run_number, tag: str | None = None) -> str:
     ap.error("--multiple-kpi-splits requires --kaon-pion-splits > 0 and --intermediate-kaon-pion-splits > 1")
 if args.max_split_buffer < 1:
     ap.error("--max-split-buffer must be >= 1")
+if args.max_event_size < 1:
+    ap.error("--max-event-size must be >= 1")
+
 if args.max_split_buffer < args.kaon_pion_splits:
     ap.error(
-        "--max-split-buffer must be >= --kaon-pion-splits: the clones buffered when the parent "
+        "--max-split-buffer and must be >= --kaon-pion-splits: the clones buffered when the parent "
```
olantwin: The and now needs to be removed.

muonShieldOptimization/exitHadronAbsorber.h
```diff
@@ -83,6 +84,7 @@ class exitHadronAbsorber : public SHiP::Detector<vetoPoint> {
   // enough to leave room for the fNsplits endpoint clones PostTrack() appends,
   // and Initialize() rejects a cap that fNsplits alone would exceed.
   std::size_t fMaxSplitBuffer = 25000;
+  std::size_t fMaxEventSize = 5000000;
```
olantwin, line 87: Using `'` as separator between thousands would make this more readable.

olantwin (changes requested):

macro/run_fixedTarget.py
THanae: The event size plus the number of splits * safety factor have to fit into the maximum event size. I wasn't so sure if we want to mention the safety factor in `run_fixedTarget` though, what do you think?

macro/run_fixedTarget.py
THanae: Thanks, done!

PR #1404, opened 2026-09-08
feat: Added option to use EvtGen decayer for heavy hadrons in run_simScript

THanae:
Updated `g4Config.C` to allow EvtGen decay of heavy hadrons when using `run_simScript`. I also updated the `CMakeList` in order to get the necessary library. 

Open to other suggestions for addressing the problem, this is mainly a suggestion.

Note also that this could be adapted to other runfiles if we think this could be useful

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for using EvtGen to decay heavy hadrons in simulations.
  * EvtGen decays are automatically configured for selected charm, beauty, quarkonium, and tau particles.
  * Simulation setup now applies the appropriate external decayer selection automatically.

* **Bug Fixes**
  * Improved simulation startup reliability when using EvtGen by preventing related decayer symbol lookup errors.

* **Documentation**
  * Updated the changelog to document EvtGen support for heavy-hadron decays.


gconfig/g4Config.C
```diff
@@ -80,4 +80,42 @@ void Config() {
...
+      511, -511, 521, -521, 531, -531, 541, -541, 513, -513, 523, -523, 533, -533,
+      // ccbar and bbbar
+      443, 553, 10441, 20443, 445, 10551, 20553, 555,
+      // charm baryons
+      4122, -4122, 4222, -4222, 4212, -4212, 4112, -4112, 4232, -4232, 4132, -4132, 4332, -4332,
+      // beauty baryons
+      5122, -5122, 5112, -5112, 5212, -5212, 5222, -5222, 5132, -5132, 5232, -5232, 5332, -5332,
+      // taus
+      15, -15
+    };
```
olantwin, line 106: For this we can probably disable clang-format for the block.

macro/run_simScript.py
```diff
@@ -915,7 +915,7 @@ def useInputFile(paths):
 # -----Configure external decayer globally------------------------------------
 # Override any previous SetPythiaDecayer calls if EvtGenDecayer is requested
 if options.evtgen_decayer:
-    run.SetPythiaDecayer("DecayConfigTEvtGen.C")
```
olantwin, line 918: We should delete the now unused config.

shipgen/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ ship_add_library(
           ParticleGunGenerator.cxx
   LINKDEF LinkDef.h
   DEPENDENCIES Base ShipData veto EvtGen::EvtGen EvtGen::EvtGenExternal
-               FairLogger::FairLogger ROOT::EGPythia6
+               FairLogger::FairLogger ROOT::EGPythia6 ROOT::EGPythia8
```
olantwin, line 14: Why is this change needed? This seems correct, but unrelated.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ it in future.
 
 * He Balloon added with configurable thickness and material.
 * 2026 BDF target design (33 pure tungsten disks with a larger rear block, steel core with serpentine He cooling grooves, jacket tube, flanges, upstream beam window and cover plate, and domed rear endcap), extracted from CATIA model ST1A07710_01_AB.02. Select with `--target-yaml geometry/target_config_2026.yaml`; the legacy design remains the default. Downstream elements are positioned using the nominal legacy target length so both designs can be compared directly.
+* Added option to use EvtGen decayer for heavy hadrons in run_simScript
```
olantwin: Maybe in parentheses "expanded from previous support for J/psi)" or similar

Also, when removing the DecayConfigTEvtGen.C, we should note the removal here.

olantwin (changes requested):

shipgen/CMakeLists.txt
THanae: If I do not make this change, I get `python: symbol lookup error: sw/slc9_x86-64/FairShip/evtgen_heavy_hadrons-local1/lib/libShipGen.so.26.07.103: undefined symbol: _ZN15TPythia8DecayerC1Ev`

CHANGELOG.md
THanae: Updated the changelog to clarify and account for the removal of the file

macro/run_simScript.py
THanae: Done :)

gconfig/g4Config.C
```diff
@@ -80,4 +80,42 @@ void Config() {
...
+
+    TEvtGenDecayer* decayer = new TEvtGenDecayer();
+    decayer->SetEvtGenDecayFile((TString(gSystem->Getenv("EVTGENDATA")) + "/DECAY.DEC").Data());
+    decayer->SetEvtGenParticleFile((TString(gSystem->Getenv("EVTGENDATA")) + "/evt.pdl").Data());
```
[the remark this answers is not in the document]
THanae: I would prefer to let it fail rather than let it continue if it cannot find those files. What do you think @olantwin ?

gconfig/g4Config.C
THanae: I added `// clang-format off` and `// clang-format on` around the block, is this what you meant?

shipgen/CMakeLists.txt
olantwin: Then this was probably a pre-existing bug, which nobody noticed because nobody was using EvtGen...
Can we make this a separate commit, and add a changelog entry?

gconfig/g4Config.C
olantwin: Yes, exactly. Seems to have worked, but clang-format isn't happy yet with the rest.

gconfig/g4Config.C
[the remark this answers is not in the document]
olantwin: I agree that it should fail, but it would be good to have a better message than a cryptic path error.

Unless we know that EVTGENDATA is set in aliBuild and pixi envs, I think it's worth having a useful error.
Alternatively, we can ensure that it's in the env as part of our activation script or the evtgen package.

gconfig/g4Config.C
```diff
@@ -80,4 +80,44 @@ void Config() {
...
+      // beauty baryons
+      5122, -5122, 5112, -5112, 5212, -5212, 5222, -5222, 5132, -5132, 5232, -5232, 5332, -5332,
+      // taus
+      15, -15
+    };
+    // clang-format on
+
+    TEvtGenDecayer* decayer = new TEvtGenDecayer();
+    decayer->SetEvtGenDecayFile((TString(gSystem->Getenv("EVTGENDATA")) + "/DECAY.DEC").Data());
+    decayer->SetEvtGenParticleFile((TString(gSystem->Getenv("EVTGENDATA")) + "/evt.pdl").Data());
```
olantwin: I'd probably prefer using std::string here instead of TString.

THanae: I've now changed it to std::string

shipgen/CMakeLists.txt
THanae: I've now changed the PR to have two separate commits :)

olantwin: Thanks!

olantwin:
Ok, now that you've mastered the commit scopes, I'd suggest you phrase both commits as imperative and shorten the title.

E.g. "feat: add option to use EvtGen for heavy hadrons" and "fix: link EGPythia8 to avoid error when using EvtGen". .You can give more info in the commit message body, if needed.

gconfig/g4Config.C
```diff
@@ -80,4 +80,54 @@ void Config() {
...
+      // beauty baryons
+      5122, -5122, 5112, -5112, 5212, -5212, 5222, -5222, 5132, -5132, 5232, -5232, 5332, -5332,
+      // taus
+      15, -15
+    };
+    // clang-format on
+
+
+    const char* evtgendata = gSystem->Getenv("EVTGENDATA");
+    if (!evtgendata) {
```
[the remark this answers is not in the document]
olantwin: @THanae The proposed fix looks ok. We can also use std::getenv, which has the same return value (nullptr if not found/unset, empty const char array if empty).

gconfig/g4Config.C
olantwin: Thanks!

gconfig/g4Config.C
```diff
@@ -80,4 +80,52 @@ void Config() {
...
+    decayer->SetEvtGenDecayFile(
+        (std::string(evtgendata) + "/DECAY.DEC").c_str());
+    decayer->SetEvtGenParticleFile(
+        (std::string(evtgendata) + "/evt.pdl").c_str());
+
+    std::string heavyHadronsString = "";
+    for (int pdg : heavyHadronsPDG) {
+      decayer->AddEvtGenParticle(pdg);
+      TParticlePDG* p = TDatabasePDG::Instance()->GetParticle(pdg);
+      if (p) heavyHadronsString += std::string(p->GetName()) + " ";
```
olantwin, line 123: I think we should use `G4ParticleTable::FindParticle(pdg)->GetParticleName()` here, I don't think ROOT and Geant4 always agree on the particle names.

macro/run_simScript.py
```diff
@@ -915,7 +915,7 @@ def useInputFile(paths):
...
+    os.environ["EXTDECAYER_HEAVY_HADRONS"] = "1"
     print("Using TEvtGenDecayer for J/psi and quarkonium decays with EvtGen")
```
olantwin: This message needs updating.

macro/run_simScript.py
```diff
@@ -915,7 +915,7 @@ def useInputFile(paths):
 # -----Configure external decayer globally------------------------------------
 # Override any previous SetPythiaDecayer calls if EvtGenDecayer is requested
 if options.evtgen_decayer:
-    run.SetPythiaDecayer("DecayConfigTEvtGen.C")
+    os.environ["EXTDECAYER_HEAVY_HADRONS"] = "1"
```
olantwin, line 917: The value this environment variable should take should probably be documented. This will also be survive the process and will be visible to subprocesses, so behaviour of subsequent run_simScript.py runs will be different.

macro/run_simScript.py
```diff
@@ -915,7 +915,7 @@ def useInputFile(paths):
 # -----Configure external decayer globally------------------------------------
 # Override any previous SetPythiaDecayer calls if EvtGenDecayer is requested
```
olantwin: This comment is no longer accurate. If any of the other config files are selected (e.g. DecayConfigPy8, DecayConfigNuAge), EvtGen will not be used.

gconfig/g4Config.C
```diff
@@ -80,4 +80,52 @@ void Config() {
...
+    }
+
+    decayer->Init();
+    geant4->SetExternalDecayer(decayer);
+    G4UImanager::GetUIpointer()->ApplyCommand(
+        "/mcPhysics/setExtDecayerSelection " + heavyHadronsString);
```
olantwin, line 129: We should remove this command in `gconfig/g4config.in:34` which is overridden by this line. Otherwise it could cause confusion.

olantwin, line 129: We might want to check the return code on this to make sure the command was applied successfully (not sure whether Geant4 would warn if e.g. a particle isn't known, but this might catch some category of misconfiguration)

olantwin: Maybe do this first, to see whether the ROOT and Geant4 names agree.

olantwin (changes requested):

gconfig/g4Config.C
THanae: The return value of this is 0

THanae: I have now removed the comment in `g4config.in`

macro/run_simScript.py
THanae: I have removed the comment

macro/run_simScript.py
THanae: I have now updated the message

macro/run_simScript.py
THanae: I have now added an `else` after the `if` statement to reset the environment variable to 0, what do you think?

gconfig/g4Config.C
THanae: Given the return value below, do you think it would still be necessary?

olantwin: Hm, maybe VMC uses ROOT names? Could you compare the two sets of names?

THanae: Comparing the two sets of names (` std::cout << "Name1: " << p->GetName() << " , Name2: " << particle->GetParticleName() << std::endl;`) leads to a segfault, but also the strings seem different:
```
[INFO] TEvtGenDecayer: Added PDG 411 to EvtGen particle list (Total: 1)
Name1: D+ , Name2: D+
[INFO] TEvtGenDecayer: Added PDG -411 to EvtGen particle list (Total: 2)
Name1: D- , Name2: D-
[INFO] TEvtGenDecayer: Added PDG 421 to EvtGen particle list (Total: 3)
Name1: D0 , Name2: D0
[INFO] TEvtGenDecayer: Added PDG -421 to EvtGen particle list (Total: 4)
Name1: D0_bar , Name2: anti_D0
[INFO] TEvtGenDecayer: Added PDG 431 to EvtGen particle list (Total: 5)
Name1: D_s+ , Name2: Ds+
[INFO] TEvtGenDecayer: Added PDG -431 to EvtGen particle list (Total: 6)
Name1: D_s- , Name2: Ds-
[INFO] TEvtGenDecayer: Added PDG 413 to EvtGen particle list (Total: 7)
Name1: D*+ , Name2:  *** Break *** segmentation violation
```

olantwin: Probably an undefined/`nullptr` `particle`?

THanae: Agreed, but the name is found with ROOT (`name1` is what is already in the code, `name2` is with geant4)

THanae: To summarise offline discussion: the code is now using geant4 particle names, and dropping the cases where the PDG ID is not found

THanae:
> Ok, now that you've mastered the commit scopes, I'd suggest you phrase both commits as imperative and shorten the title.
> 
> E.g. "feat: add option to use EvtGen for heavy hadrons" and "fix: link EGPythia8 to avoid error when using EvtGen". .You can give more info in the commit message body, if needed.

Thank you for the suggestion, I changed the commit messages as suggested

PR #1405, opened 2026-09-11
Keep the split parent's weight in step with the survival ledger

olantwin:

Stacked on #1389.

`PostTrack()` had a branch for a kaon or pion whose transport was not over yet, which pushed a continuation track carrying the remaining weight. Geant4 never hands us such a track:

* `G4TrackingManager::ProcessOneTrack` steps while the status is `fAlive` or `fStopButAlive`, so the loop exits only for `fStopAndKill`, `fKillTrackAndSecondaries`, `fSuspend` or `fPostponeToNextEvent`.
* `TG4TrackingAction::PostUserTrackingAction` calls `PostTrack()` only when the status is not `fSuspend` and the track was not interrupted.
* `TG4StepManager::IsTrackStop()` is true for every status that is left.

So `isParticleDestroyed` was always true, the branch never ran, and `fContinuationTracks` stayed empty. The first commit removes it together with the set and its two lookups, and corrects the buffer cap comment that described the weight going to a continuation track. Behaviour is unchanged.

That exposes the bug the branch could never have fixed. Per-step splitting moves weight `S*P` into clones at every step and reduces the survival factor `S`, but `S` was only a bookkeeping variable: the parent's own weight stayed where it started. Everything the parent went on to do therefore counted at full weight, in particular its hit at the sensitive plane and the secondaries it made when it ended by interacting instead of decaying, overcounting that branch by `1/S`. The second commit scales the parent's weight by the same `1 - P` factor.

The clone weights need no change, since `ShipStack::PushTrack` multiplies a pushed weight by the weight of the track named as its parent, and a clone names the split parent's mother. Geant4 pushes a secondary onto the VMC stack once that secondary starts tracking, which is after the parent is done, so the secondaries pick up the reduced weight on their own. Runs with `--kaon-pion-splits` alone are unaffected: the per-step block is guarded by `!fSplitOnce`, so `S` stays 1 there.

### Checks

15 events with `--kaon-pion-splits 5 --multiple-kpi-splits --intermediate-kaon-pion-splits 2`. For 135 of the 157 split parents that can be reconstructed from the output, the residual `|parent + clones - initial|` is well below the summed clone weight, with a median of 2.4% of it. Before this change that ratio is 1 by construction. The other 22 parents, and the floor at about 1e-7, come from the reconstruction rather than from the physics: `ShipStack` drops and remaps tracks so some clones land under the wrong mother, and `fW` is a `Double32_t` stored as a float.

Left alone: the daughters of a natural decay still carry `W*S` in the track array and are kept out of the sensitive plane only by `fDecayedParentIDs`. #1398 will need a one line merge where `fContinuationTracks.clear()` disappears from `BeginEvent()`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected particle-weight handling for per-step kaon/pion splitting.
  - In runs using `--multiple-kpi-splits`, the parent particle’s weight is now reduced as weight is distributed among clones.
  - Ensured the parent and its clones collectively retain the original weight, with resulting secondary particles receiving the appropriate reduced weight.

### DIFF
--- a/muonShieldOptimization/exitHadronAbsorber.cxx
+++ b/muonShieldOptimization/exitHadronAbsorber.cxx
@@ -114,10 +114,9 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
     }
   }
 
-  // Both counts have to be positive: PostTrack() only stops the parent and
-  // hands on the remaining weight when fNsplits > 0, so buffering per-step
-  // clones with fNsplits == 0 would leave the parent carrying its full weight
-  // alongside the clones.
+  // Both counts have to be positive: PostTrack() only terminates the ledger,
+  // handing what is left of it to the endpoint clones, when fNsplits > 0.
+  // run_fixedTarget.py pairs the two options for the same reason.
   if (fNsplits > 0 && fIntermediateNsplits > 0 && (!fSplitOnce)) {
     Int_t currentTrackId = gMC->GetStack()->GetCurrentTrackNumber();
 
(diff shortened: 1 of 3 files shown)


muonShieldOptimization/exitHadronAbsorber.cxx
```diff
@@ -205,7 +203,20 @@ Bool_t exitHadronAbsorber::ProcessHits(FairVolume* vol) {
...
+          // fCurrentSurvivalFactor, otherwise everything it goes on to do
+          // still counts at full weight: its hit at the sensitive plane, and
+          // the secondaries it makes if it ends by interacting rather than
+          // decaying. Scaling it here keeps parent + clones equal to the
+          // weight the parent started with at every step. Geant4 pushes a
+          // secondary onto the VMC stack when that secondary starts tracking,
+          // which is after the parent is done, so the secondaries pick this
+          // up on their own.
           fCurrentSurvivalFactor *= (1.0 - P_decay);
+          part->SetWeight(part->GetWeight() * (1.0 - P_decay));
```
THanae, line 219: I think `GetWeight` will always return 1, because the weight from the stack is not passed through. That's why there are multiplication factors in the stack

olantwin: Probably related to the coderabbit comment above. I'm so glad we've gotten rid of VMC for aegir.

PR #1411, opened 2026-09-16
Some followups to the FTFT

olantwin:
- Makes it mutually exclusive with options where its' a no-op
- Add tune to tag
- Fix issue with charm tag
- Allow specifying "default" as a tune name

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a `--pythia8-tune` option for fixed-target simulations, supporting the default Pythia8 tune or FTFT.
  - Added validation for incompatible tune selections with charm, beauty, and direct Geant4 modes.
  - Work-directory names now identify the selected production mode or non-default tune.

- **Bug Fixes**
  - Fixed charm simulation work-directory names so they retain the `charm` identifier.
  - Improved handling and error messages for unsupported Pythia8 tune selections.


olantwin:
@coderabbitai review

shipgen/FixedTargetGenerator.h
```diff
@@ -96,9 +96,11 @@ class FixedTargetGenerator : public SHiP::Generator {
     EMax = emax;
   }  // min energy to be copied to Geant4
   inline void SetDebug(Bool_t x) { Debug = x; }
-  // Pythia8 tune for the primary interaction: "" (default, Monash 2013) or
+  // Pythia8 tune for the primary interaction: "" or "default" (Monash 2013) or
   // "FTFT" (fixed-target open charm and beauty tune, arXiv:2608.29076)
-  inline void SetPythiaTune(const TString& tune) { fTune = tune; }
+  inline void SetPythiaTune(const TString& tune) {
```
THanae, line 101: What is the advantage of setting the tune to null every time it is default compared to what was being done before?

olantwin: Basically means that things don't crash if the default arg from run_fixedTarget is ever passed through. Before we relied on it being set to an empty string.

THanae (approved):

PR #1419, opened 2026-09-19
feat(scoringplane)!: replace SplitCal with an ideal scoring plane

matclim:
> Description drafted with AI assistance (claude.ai, Claude Haiku 4.5) and reviewed by me.

## Summary
Replaces the SplitCal calorimeter with `ScoringPlane`, an ideal scoring plane at the former
SplitCal front face, for studies that need the particles arriving at the calorimeter position.

## Changes
- New `ScoringPlane` detector: 4 m × 6 m × 1 mm vacuum box (`ship_geo.ScoringPlane`, node
  `ScoringPlane_1`). Every track *entering* it produces a `ScoringPlanePoint` (MC-truth position,
  momentum, time, PDG code); no energy-deposit requirement, so neutrals are recorded too.
  Tracks with a point are kept in `MCTrack`.
- Removed the `splitcal` library, its classes and branches, `splitcalDetector.py`,
  `ship_geo.SplitCal` and `ship_geo.EcalOption`.
- `kSplitCal` → `kScoringPlane` in the same slot, so other detector IDs are unchanged.
- Muon station positions unchanged.
- `TrackExtrapolateTool` now switches from Runge–Kutta to linear extrapolation at the scoring
  plane (previously at the SplitCal centre, ~71 cm further downstream).

## Breaking change
Files and code relying on SplitCal classes or branches no longer work.

## Testing
- Built on CVMFS 26.07 with aliBuild; ctest passes.
- Particle-gun samples (10 GeV μ, e, π from z = 50 m): points recorded at the plane as expected.

## Checklist
- [x] Changelog updated
- [x] Commit message follows conventional commits

## Summary by CodeRabbit

- **New Features**
  - Added a lightweight Scoring Plane that records entering tracks’ truth-level position, momentum, time, and particle type, including neutral particles.
  - Added Pythia8 tune selection for fixed-target open-charm and beauty production, defaulting to Monash 2013.

- **Changes**
  - Replaced the SplitCal detector with the Scoring Plane in geometry, simulation, reconstruction, and configuration.
  - Track extrapolation now uses linear continuation beyond the scoring plane.

- **Removals**
  - Removed SplitCal calorimeter digitization, clustering, and related configuration options.


matclim:
@coderabbitai full review

CHANGELOG.md
```diff
@@ -22,6 +23,8 @@ it in future.
 
 ### Removed
 
+* Remove the SplitCal calorimeter (`splitcal`, `splitcalPoint`, `splitcalHit`, `splitcalCluster`, `splitcalDetector.py`, `ship_geo.SplitCal`, `ship_geo.EcalOption`). The `kSplitCal` detector ID slot is reused by `kScoringPlane`. Muon station positions are unchanged. `TrackExtrapolateTool` now switches from Runge-Kutta to linear extrapolation at the scoring plane.
```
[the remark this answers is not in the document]
matclim: Done: added a migration note (use ship_geo.ScoringPlane / ScoringPlanePoint; no direct replacement for the calorimeter hits, clustering and digitisation).

python/geometry_config.py
```diff
@@ -370,44 +370,18 @@ def create_config(
...
-    )
 
-    c.MuonStation0 = AttrDict(z=c.SplitCal.ZStart + 10 * u.cm + c.SplitCal.SplitCalThickness)
+    # ScoringPlane: ideal (vacuum) plane that replaces the former SplitCal.
+    # Placed at the old SplitCal front face; every track entering it is scored.
+    c.ScoringPlane = AttrDict()
+    c.ScoringPlane.z = 38.450 * u.m + c.decayVolume.z  # centre, relative to decay vessel centre
+    c.ScoringPlane.DX = 4 * u.m  # full width  (former SplitCal 2*XMax)
+    c.ScoringPlane.DY = 6 * u.m  # full height (former SplitCal 2*YMax)
+    c.ScoringPlane.DZ = 1 * u.mm  # full thickness
```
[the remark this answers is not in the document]
matclim: Keeping this as is: ScoringPlane.z is intentionally the centre of the 1 mm plane, placed at the former SplitCal front face, as documented in geometry_config.py and ScoringPlane/README.md. Shifting it by DZ/2 = 0.5 mm has no effect for an ideal vacuum scoring plane, and no volume sits within that distance.

matclim:
I don't believe the coderabbit summary is correct here: 


`Added Pythia8 tune selection for fixed-target open-charm and beauty production, defaulting to Monash 2013.`

I did no such thing

python/geometry_config.py
```diff
@@ -370,44 +370,18 @@ def create_config(
...
+
+    # Muon stations kept at their previous positions, i.e. behind the space the
+    # SplitCal used to occupy (10 cm + 142 cm SplitCal thickness).
+    c.MuonStation0 = AttrDict(z=38.450 * u.m + c.decayVolume.z + 10 * u.cm + 142 * u.cm)
```
olantwin: I think these don't exist anymore. If they do, they should be removed.

python/shipDet_conf.py
```diff
@@ -397,40 +397,11 @@ def configure(run, ship_geo):
...
-            x.first_precision_layer,
-            x.second_precision_layer,
-            x.third_precision_layer,
-            x.num_precision_layers,
-        )
-        SplitCal.SetNModules(x.NModulesInX, x.NModulesInY)
-        SplitCal.SetNStrips(x.NStripsPerModule)
-        SplitCal.SetStripSize(x.StripHalfWidth, x.StripHalfLength)
-        detectorList.append(SplitCal)
+    if hasattr(ship_geo, "ScoringPlane"):  # absent in geometry files made before SplitCal removal
```
olantwin: Maybe we should add an `else` here with a sensible warning.

ScoringPlane/ScoringPlane.h
```diff
@@ -0,0 +1,54 @@
...
+class ScoringPlanePoint;
+class FairVolume;
+
+/**
+ * @brief Ideal scoring plane downstream of the timing detector.
+ *
+ * Replaces the former SplitCal calorimeter. The plane is a thin vacuum box
+ * placed where the SplitCal front face used to be. Every track *entering*
+ * the box produces one ScoringPlanePoint holding the MC-truth position,
+ * momentum, time and PDG code at the entry face. No energy deposit is
```
olantwin: I think processhits runs after the step, so this should be all at the exit face (difference minimal)

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ it in future.
 
 ### Added
 
+* Add `ScoringPlane` detector: an ideal thin vacuum plane at the former SplitCal front face (`ship_geo.ScoringPlane`). Every track entering it is stored as a `ScoringPlanePoint` (MC-truth entry position, momentum, time, PDG code), without an energy-deposit requirement.
```
olantwin: Again, if I'm not mistaken, this is the exit information

CODEOWNERS
```diff
@@ -15,6 +15,4 @@ python/detectors/MTCDetector.py @eduard322
 
 passive/ShipMuonShield.* @Gfrisella  # Guglielmo Frisella (U Zürich)
 
-splitcal/* @matclim  # Matei Climescu (Ghent University)
```
olantwin, line 18: Unless we decide to make this a more general utility for arbitrary stand-in detectors, please make yourself the codeowner for the PID scoring plane.

CaloScoringPlane/CaloScoringPlane.cxx
```diff
@@ -0,0 +1,76 @@
...
+
+  TParticle* p = gMC->GetStack()->GetCurrentTrack();
+  const Int_t pdgCode = p->GetPdgCode();
+
+  AddHit(fEventID, fTrackID, fVolumeID, TVector3(fPos.X(), fPos.Y(), fPos.Z()),
+         TVector3(fMom.Px(), fMom.Py(), fMom.Pz()), fTime, fLength, 0.,
+         pdgCode);
+
+  // Flag the track as having a detector point so that ShipStack keeps it.
+  ShipStack* stack = dynamic_cast<ShipStack*>(gMC->GetStack());
```
olantwin, line 59: I'm not sure this is always enough to keep the track if there's an energy cut. This should be tested. IIRC, FairStack allows several options, which we inherit in ShipStack, e.g. a threshold number of hits, energy, OR of both, AND of both...

olantwin (changes requested):
Generally, this looks like a great cleanup!

A few comments on first read of the changes:
- Maybe we'd want to give this a less generic name? (or allow several ScoringPlanes, with one with this special role as SplitCal stand-in).
- Shouldn't we remove the muon stations as well in the same go? (if they're even still there? Until your PR description I thought we had already removed them)
- In the different branch lists, e.g. in run_simScript.py, we might need to add the new point type.

matclim:
_Naming_: renamed throughout to CaloScoringPlane, including the directory, classes, enum value, node and branches.
_Muon stations_: removed. Their z values were still used by Genie, the event display and CMBG_conf, so they're replaced by ship_geo.zEndOfDetector, which keeps the old numeric value, leaving the Genie generation volume unchanged.
_Exit face_: confirmed, docs corrected.
_Stack_: confirmed. With the default MCTracksWithEnergyCutOnly, points don't keep a track; a 10 GeV muon sample shows a delta ray whose link is −2 while its hit is still written and identified. Documented in the code and the README.
_Warning, codeowner, branch lists_: added.

python/geometry_config.py
```diff
@@ -370,59 +370,19 @@ def create_config(
...
-    c.Muon = AttrDict()
-    c.Muon.XMax = 250.0 * u.cm
-    c.Muon.YMax = 325.0 * u.cm
 
-    c.Muon.ActiveThickness = 0.5 * u.cm
-    c.Muon.FilterThickness = 30.0 * u.cm
+    # CaloScoringPlane: ideal (vacuum) plane that replaces the former SplitCal.
+    # Placed at the old SplitCal front face; every track entering it is scored.
+    c.CaloScoringPlane = AttrDict()
+    c.CaloScoringPlane.z = 38.450 * u.m + c.decayVolume.z  # centre, relative to decay vessel centre
```
amagnan, line 377: Can we not use something not hardcoded, like decayVolume.z + decay_vessel_length /2 + sst_length + timedet_length or more simply timedet_pos + timedet_length/2 + 1mm ?

python/geometry_config.py
```diff
@@ -370,59 +370,19 @@ def create_config(
...
+    c.CaloScoringPlane.DX = 4 * u.m  # full width  (former SplitCal 2*XMax)
+    c.CaloScoringPlane.DY = 6 * u.m  # full height (former SplitCal 2*YMax)
+    c.CaloScoringPlane.DZ = 1 * u.mm  # full thickness
+
+    # Downstream end of the experiment, used as a reference by the generators and
+    # the event display. Same value as the former MuonStation3, so that the Genie
+    # generation volume is unchanged.
+    c.zEndOfDetector = c.CaloScoringPlane.z + 4.52 * u.m
```
amagnan, line 385: what is this hardcoded 4.52 m, can we not then have a global value, which would be the cavern back wall position, and check instead that the scoring plane end point fits within it ?

amagnan (commented):
Looks good to me except I find very confusing these hardcoding of dimensions, I think it is fair to assume that people will use this scoring plane after the time det and I would use the TD for ref position then. Or, the spectrometer magnet which should also always be at the same z and cannot move as far as I understand. And for the end of the cavern, the cavern wall position is also the most robust I believe.

matclim:
@amagnan Agreed on both, thanks.

The plane can be anchored to the timing detector: c.TimeDet.z + dzBarRow + dzBarCol + zBar/2 is its downstream face, plus a configurable gap and DZ/2 for the plane's centre. Note this moves the plane from z = 96.570 m (the old SplitCal front face) to about 96.010 m, i.e. 56 cm upstream. Nothing but vacuum lies in between, so only the recorded z and the RK-to-linear switch point in TrackExtrapolateTool change. Should I move it, or keep the current position by putting the 56 cm into the gap?

The 4.52 m is the old MuonStation3 offset, which I kept only so that the Genie generation volume would be unchanged. Anchoring to the cavern is better, but the cavern is hardcoded in ShipCave.cxx (z_transition = 20.52 m, ECN3_length = 100 m, so the ECN3 back wall is at z = 120.52 m). I'd move those into c.cave, pass them to ShipCave via setters, define c.cave.zEnd, and assert that the scoring plane fits inside. One consequence to agree on first: Genie's end position would then move from 101.09 m to the wall at 120.52 m, which enlarges the neutrino generation volume and changes the samples. Do you want that, or should Genie keep its current end, defined from the scoring plane plus a margin?

amagnan:
So @matclim , what would it be in reality for the gap between TD and ECAL ? I think we should take that for answering your question. If not clear, maybe smaller gap is prefered and particles can be forward propagated as needed ?? I am not sure what's best....
For the GENIE, who uses the defaults in FairShip at the moment, and why would we generate anything beyond the ECAL ? My understanding was that the configuration is quite tailored per volume where we want events, so in principle I would have expected people to concentrate on just He, just SBT sensitive+frame (or even separate sensitive from frame), just UBT when we have it, and possibly some layer of AIR outside SBT and before/after UBT and before SST, and possibly He balloon also. Nothing beyond ECAL ??

